PR #3, opened 2023-11-30
[sknpm] Fix build on MacOS

skiplabsdaniel:
Add test-client make target

### DIFF
--- a/skbuild/src/harness.sk
+++ b/skbuild/src/harness.sk
@@ -1,5 +1,7 @@
 module Skbuild;
 
+const kBUFFER_LEN: Int = 256 * 1024;
+
 class Env(
   refDir: String,
   outDir: String,
@@ -77,7 +79,7 @@ class Env(
   }
 }
 
-fun build(): Env {
+fun build(): Result<Env, IO.Error> {
   env = Env::create();
   includeDirs = Array[];
   #forEachFunction (@skbuild_include_dirs, #function, #functionName) {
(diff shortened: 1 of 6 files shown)


skbuild/src/harness.sk
```diff
@@ -248,36 +248,31 @@ fun buildLibrary(
...
-    cmd = Array[
-      "objcopy",
-      "-B",
-      b,
-      "-I",
-      "binary",
-      "-O",
-      o,
-      file,
-      object,
```
jberdine, line 274: Question: Is the change from calling objcopy to using the `copyBinary` function only because macos does not ship with gnu objcopy, or is there another motivation? I wonder if since we will need llvm to be installed anyhow, maybe changing this use of `objcopy` to `llvm-objcopy` would be enough?

skiplabsdaniel: That's exactly the motivation, and indeed there is llvm-objcopy, however I have not been able to get it to work correctly on macos. This is the simplest solution I've found to handle all platforms.

skbuild/src/harness.sk
```diff
@@ -322,18 +317,37 @@ fun ensureDirectory(path: String, verbose: Bool): void {
...
+    } else {
+      _ = dstFile.write(",  ");
+    };
+    _ = dstFile.write("0x" + Chars.intToHexDigits(code, 2));
+    !count = count + 1;
+  };
+  _ = dstFile.write(`\n};\nunsigned int ${name}_len = ${count};\n`);
+  srcFile.close();
+  dstFile.close()
 }
```
jberdine: @beauby, this is the function I mentioned would be a good candidate to see if your efficient IO APIs are comfortable to use.

skbuild/src/harness.sk
```diff
@@ -1,5 +1,7 @@
 module Skbuild;
 
+const kBUFFER_LEN: Int = 1024;
```
jberdine: This seems pretty small to me. See e.g. https://git.savannah.gnu.org/cgit/coreutils.git/tree/src/ioblksize.h#n24 which reports on benchmarks done to decide to use 256KiB for buffers in GNU coreutils.

jberdine (approved):
LGTM. The new IO stuff seems to fit well!

PR #4, opened 2023-11-30
[skdate] Add date package and use it in skdb

skiplabsdaniel:


skdate/src/constants.sk
```diff
@@ -0,0 +1,127 @@
+module SKDate;
+
+base class DateConstant() uses Hashable, Orderable {
+  fun getName(): String;
+  fun getShort(): String;
+  fun getCode(): Int;
+
+  fun compare(other: DateConstant): Order {
+    this.getCode().compare(other.getCode())
```
bennostein: This allows to compare across incomparable DateConstants, e.g. `Monday` and `January` are considered equal.  Could we push it down to the child classes?

skdate/src/constants.sk
```diff
@@ -0,0 +1,127 @@
...
+    | 6 -> Saturday()
+    | 7 -> Sunday()
+    | _ -> invariant_violation("Invalid day code")
+    }
+  }
+
+  fun toNext(day: Day): Int {
+    current = this.getCode();
+    next = day.getCode();
+    if (next >= current) next - current else 7 - (current - next)
```
bennostein: This typing/naming was a bit confusing to me -- could we call this `daysUntilNext` or similar?

Also, could just do `(next - current + 7) % 7` instead of the conditional.

skdate/src/parser.sk
```diff
@@ -0,0 +1,267 @@
...
+YYYY-MM-DD HH:MM:SS
+YYYY-MM-DD HH:MM:SS.SSS
+YYYY-MM-DDTHH:MM
+YYYY-MM-DDTHH:MM:SS
+YYYY-MM-DDTHH:MM:SS.SSS
+*/
+
+fun parseDate(
+  iter: mutable Iterator<Char>,
+  yearV: readonly Vector<Char>,
```
bennostein: Would be nicer to just pass the year as an int here, and use the `toInt4` helper at the line 74 callsite.

(and analogously for `parseTime`'s `hoursV` argument -- just pass the int, and use `toInt2` at callsites)

skdate/src/parser.sk
```diff
@@ -0,0 +1,267 @@
...
+/*****************************************************************************/
+
+const codeZero: Int = '0'.code();
+
+fun toInt(input: readonly Vector<Char>): Int {
+  result = 0;
+  power = 1;
+  size = input.size();
+  for (i in Range(0, size)) {
+    !result = result + (input[size - 1 - i].code() - codeZero) * power;
```
bennostein, line 32: There's an implied invariant here that `input` contains only decimal characters '0', '1', '2', ... ,'9', but that's not reflected in the type or checked here.

Might be nicer to have this take a `readonly Vector<Int>` as input, and do the input validation and char -> int conversion separately?  Could also cut down on the repetition of near-identical pattern matches throughout (all of the `Some(x) if (x >= '0' && x <= '9') -> x` stuff) if there were a helper `toDigit : ?Char -> ?Int`

skdate/src/parser.sk
```diff
@@ -0,0 +1,267 @@
...
+}
+
+fun parseTimeZone(str: String): ?TimeZone {
+  iter = str.getIter();
+  sign = iter.next() match {
+  | Some('+') -> Some('+')
+  | Some('-') -> Some('-')
+  | _ -> None()
+  };
+  sign.flatMap(s -> parseTimezone(iter, s))
```
bennostein:
```suggestion
  | Some(sign @ '+') 
  | Some(sign @ '-') -> parseTimeZone(iter, sign)
  | _ -> None()
  };
```

skdate/ts/src/sk_date.ts
```diff
@@ -0,0 +1,94 @@
...
+      let date = new Date(year, month - 1, day);
+      const timeZone = date.toLocaleDateString(undefined, {
+        timeZoneName: "short",
+      });
+      return utils.exportString(timeZone);
+    };
+    this.SKIP_locale = (code: int, value: int) => {
+      let str = String.fromCharCode(code);
+      if (str == "B" || str == "b") {
+        let date = new Date(Date.UTC(2023, 0 + (value - 1), 1)); // January
```
bennostein, line 54: I'm confused -- why this particular constant (and the others below)?

sql/ts/tests/tests.ts
```diff
@@ -1320,6 +1320,56 @@ const watchChangesTests: (asWorker: boolean) => Test[] = (
...
 
+const dateTests = (asWorker: boolean) => {
+  let dn = (name: string, asWorker: boolean) => {
+    return n("Date: " + name, asWorker);
+  };
+  return [
+    {
+      name: dn("Add days", asWorker),
+      fun: async (skdb: SKDB) => {
+        return await skdb.exec(
```
bennostein: Could use `.scalarValue()` on the result of this `exec` and then just expect the value "2023-01-03" (and similarly throughout) to avoid hardcoding the `"col<0>"` and simplify things a bit

skdate/src/date.sk
```diff
@@ -0,0 +1,757 @@
...
+            1
+          } else {
+            0
+          });
+      !month = month + 1;
+    };
+    days + this.day;
+  }
+
+  fun isLeapYear(): Bool {
```
bennostein: unused?

skdate/src/date.sk
```diff
@@ -0,0 +1,757 @@
...
+  }
+
+  private fun weeks(year: Int): Int {
+    52 +
+      if (this.p(year) == 4 || this.p(year - 1) == 3) {
+        1
+      } else {
+        0
+      };
+  }
```
bennostein: Why are these instance and not static methods?

skdate/src/date.sk
```diff
@@ -0,0 +1,757 @@
...
+      };
+      Date(
+        year,
+        Month::monthForCode(month + 1),
+        UInt8::truncate(1 + nMaxDays + nDays),
+      );
+    };
+  }
+
+  fun dayBetween(min: Int, max: Int): Int {
```
bennostein: nit:
```suggestion
  fun daysBetween(minYear: Int, maxYear: Int): Int {
```

skdate/src/date.sk
```diff
@@ -0,0 +1,757 @@
...
+  if (SkargoVersion.kWasm32) {
+    (jsCurrentTimeMillis() * 1000.0).toInt()
+  } else {
+    natCurrentTimeMillis()
+  }
+}
+
+class File(value: Calendar) extends SKStore.File
+
+const date_maxDaysInMonth: Array<Int> = Array[
```
bennostein: Everywhere this is used, it's not as a "maximum" but to get the number of days in a month as in
```
        date_maxDaysInMonth[month] +
        if (month == 1 && this.leapYear(year)) {
          1
        } else {
          0
        }
```
Could I suggest instead to factor that all out as something like the following to minimize repetition and avoiding forgetting that leap year logic in the future?
```
private const daysInNonLeapYearMonth : Array<Int> = [...];
private static fun daysInMonth(month: Int, year: Int): Int {
 days = daysInMonth[month]
 if ( month == 1  && isLeapYear(year) ) { days + 1 } else { days } 
}
```

skdate/src/date.sk
```diff
@@ -0,0 +1,757 @@
...
+  fun getMonthStart(): Date {
+    !this.day = UInt8::truncate(1);
+    this
+  }
+
+  fun getYearStart(): Date {
+    !this.day = UInt8::truncate(1);
+    !this.month = January();
+    this
+  }
```
bennostein, line 73: Seems like a potential foot-gun for these "get" functions to mutate their receiver?

skdate/src/date.sk
```diff
@@ -0,0 +1,757 @@
...
+        if (tsecond >= 0) {
+          tsecond / 60
+        } else {
+          (tsecond / 60) - 1
+        },
+      );
+    };
+  }
+}
+
```
bennostein, line 372: There's some inconsistency here where `add{Minutes|Seconds|Hours}` mutate their receiver, and `add{Days|Months|Years}` above leave their receiver unchanged and construct/return a new `Date`.

For what it's worth, I much prefer the immutable interface

bennostein (commented):
Looked through most of this and have a range of low-to-medium level comments.  But overall looks solid and I think good to land if everything's green on rebase. Really nice test coverage!

You're a hero for working on this
datetime

skdate/src/date.sk
skiplabsdaniel: Not in Skip.

skdate/src/date.sk
skiplabsdaniel: The class is not mutable, so the receiver is not mutated.

skdate/ts/src/sk_date.ts
skiplabsdaniel: Comment added

PR #5, opened 2023-11-30
propagate skfs->skdb renaming to docker files, docs, etc.

bennostein:
I went through and did the skfs to skdb renaming where it refers to the root directory of the repo, including changing the workdir in some docker images and corresponding deployment scripts.

This will likely break some people's $USER-skfs dockerfiles and scripts in the short term, but seems better to do it sooner than later.


jberdine (approved):
I tested this locally and LGTM

PR #7, opened 2023-11-30
Improve wasm exception

skiplabsdaniel:
Manage exception re-throw
Manage exception cause
Prevent external exception message loss.

### DIFF
--- a/compiler/runtime/runtime32_specific.c
+++ b/compiler/runtime/runtime32_specific.c
@@ -2,13 +2,14 @@
 #include <stdbool.h>
 #include "xoroshiro128plus.h"
 
-void __cxa_throw(void*, void*, void*);
+void js_throw(void*, uint32_t);
 
 void* exn = (void*)0;
 
 void SKIP_throw(void* exc) {
+  uint32_t rethow = exn == exc;
   exn = exc;
-  __cxa_throw(exc, 0, 0);
+  js_throw(exc, rethow);
 }
 
 void SKIP_saveExn(void* exc) {
--- a/prelude/ts/src/sk_runtime.ts
+++ b/prelude/ts/src/sk_runtime.ts
@@ -91,9 +91,9 @@ class LinksImpl implements Links {
       utils.log(str, Stream.OUT);
     };
     this.SKIP_delete_external_exception = utils.deleteException;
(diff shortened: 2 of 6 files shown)


compiler/runtime/runtime32_specific.c
```diff
@@ -2,13 +2,14 @@
 #include <stdbool.h>
 #include "xoroshiro128plus.h"
 
-void __cxa_throw(void*, void*, void*);
+void __cxa_throw(void*, uint32_t, uint32_t);
```
jberdine: Is there a spec or some docs for what this does with llvm targeting wasm? I wonder since this doesn't seem to correspond to the usual C++ ABI:
```
    void __cxa_throw (void *thrown_exception, std::type_info *tinfo, void (*dest) (void *) );
```
>The arguments are:
>- The address of the thrown exception object (which points to the throw value, after the header, as specified above).
>- A std::type_info pointer, giving the static type of the throw argument as a std::type_info pointer, used for matching potential catch sites to the thrown exception.
>- A destructor pointer to be used eventually to destroy the object.

prelude/ts/src/sk_runtime.ts
```diff
@@ -75,13 +76,16 @@ class LinksImpl implements Links {
     this.SKIP_print_string = (msg: ptr) => {
       utils.sklog(msg, Stream.OUT, true);
     };
-    this.__cxa_throw = (exc: ptr) => utils.ethrow(exc);
+    this.__cxa_throw = (exc: ptr, rethrow: int) => utils.ethrow(exc, rethrow > 0);
```
jberdine: It isn't obvious to me why this uses `>` rather than `!=`.

jberdine (commented):
LGTM, a couple questions inline.

compiler/runtime/runtime32_specific.c
skiplabsdaniel: Good point, I will manage it cleanly

jberdine (approved):

gregsexton:
Looking forward to this! Let's get it merged. Would have been very helpful yesterday. :D

gregsexton:
I rebased this and reformatted. I'll update the pr in a bit to save you duplicating the work.

gregsexton:
Should be good to merge now.

PR #10, opened 2023-12-05
Simplify global group definitions in init.sql

bennostein:
With the new style of permission checking (i.e. checking admin privileges as a special-case on writes to  skdb_group_permissions) we don't need the indirections previously employed in `init.sql` anymore.

This should preserve the privacy checking semantics and make things much more straightforward to read/understand.

### DIFF
--- a/sql/privacy/init.sql
+++ b/sql/privacy/init.sql
@@ -57,9 +57,9 @@ CREATE UNIQUE INDEX skdb_groups_users_unique ON skdb_groups_users(groupID);
 -------------------------------------------------------------------------------
 
 INSERT INTO skdb_groups VALUES
-  ('read-only', NULL, 'root', 'root'),
-  ('read-write', NULL, 'root', 'root'),
-  ('write-only', NULL, 'root', 'root')
+  ('read-only', NULL, 'root', 'read-only'),
+  ('read-write', NULL, 'root', 'read-only'),
+  ('write-only', NULL, 'root', 'read-only')
 ;
 
 INSERT INTO skdb_group_permissions VALUES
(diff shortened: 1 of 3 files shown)


jberdine (commented):
LGTM. Would be worth having another pair of eyes on this though.

sql/privacy/init.sql
```diff
@@ -53,48 +53,21 @@ CREATE UNIQUE INDEX skdb_groups_users_unique ON skdb_groups_users(groupID);
...
+INSERT INTO skdb_group_permissions VALUES
+       ('read-only', NULL, skdb_permission('r'), 'read-only'),
+       ('write-only', NULL, skdb_permission('w'), 'read-only'),
+       ('read-write', NULL, skdb_permission('rw'), 'read-only')
 ;
 
--- read-write is visible by everyone, but only the root can change the
--- permissions (which should never happen).
+UPDATE skdb_groups SET skdb_access='read-only'
+       WHERE groupID IN ('read-only', 'write-only', 'read-write');
```
gregsexton: I suspect there's a good reason, why not set this when inserting?

gregsexton (commented):
LGTM!

sql/privacy/init.sql
bennostein: hmm, it's necessary when setting up groups as a non-root user (which is what I've been working on recently) because the 'read-only' group doesn't exist yet at the insert.  But this SQL only ever gets interpreted as root, so may be unnecessary here.  will test locally without the update.

ltratt:
This looks good to me. One pain point I've encountered -- which is a bit orthogonal but might be worth mentioning here -- is that it's really easy to misspell (or misremember) the "standard" group names, and then have surprising behaviour. I wonder if it's possible to validate that, at least when a row is inserted (if the group disappears later that's a different matter), the group name is valid?

bennostein:
Unless I'm misunderstanding the pain point you're describing, then this should be happening now!  I.e. if you try to insert with `skdb_access` set to `typo-group-name`, the client should come back with a message like `Error, line 0: cannot insert with skdb_access set to typo-group-name (user ltratt)`

ltratt:
Aha! OK, I might just need to update.

PR #11, opened 2023-12-06
Efficient IO

beauby:


compiler/src/AsmOutput.sk
```diff
@@ -2760,17 +2779,7 @@ private fun llvmGetAddr(
```
jberdine: Aside: I don't understand the need for the whole `if` above. In particular, I don't understand the comment "We need to turn that into a power-of-two-byte load". LLVM IR does not have any such constraint AFAIU. I don't understand why this code isn't duplicating work that LLVM can do itself if we just ask it to do some strange thing like read 2.5 bytes from some inconvenient offset.

compiler/src/AsmOutput.sk
```diff
@@ -2677,6 +2678,25 @@ private fun llvmWriteDefault(
...
+  asm.print("  %%%s = getelementptr inbounds " % id);
+
+  optinfo.getInstr(instr.addr) match {
+  | g @ ConstantGlobalSingleton _ ->
+    globalsType = asm.common.llvmGlobalsType();
+    asm.print("%s, %s* @%n, i64 0" % globalsType % globalsType % g)
+  | base -> asm.print("i8, %N" % base)
+  };
+
+  firstByteOffset = instr.bitOffset.shr(3);
```
jberdine: It isn't clear to me how this gives the same result as before this patch. On line 2751 before this patch, `firstByteOffset` gets decremented in a loop. After this patch, those updates will not take effect. Have I misunderstood?

compiler/src/AsmOutput.sk
```diff
@@ -3723,6 +3732,29 @@ private fun llvmWrite(instr: Stmt, asm: mutable FunAsmDefBuilder): void {
...
+      addr,
+      bitOffset => info.bitOffsets[0] + extraBitOffset,
+      addrByteAlignment,
+      canCSE => optinfo.getInstr(obj).typ.isDeepFrozen(),
+    };
+    llvmWriteGetAddr(
+      fakeLoad,
+      "r" + optinfo.getInstr(id).idValue().toString(),
+      asm,
+    )
```
jberdine: IIUC, this prints a gep instruction where the address calculation is specified by the fakeLoad, and assigns the result to a register named r<something>. But I don't see anywhere this register is used, am I missing something?

compiler/src/IR.sk
```diff
@@ -879,6 +882,10 @@ base class Stmt extends InstrWithPos {
     tupleIndex: Int, // index of field in element
     value: InstrID,
   }
+  | ArrayUnsafePtr{
+    obj: InstrID,
+    offset: InstrID, // index of element in array
```
jberdine: Are `offset`s and `index`es the same? I expected that indices would be expressed in units of the element size, while offsets would be expressed in bytes or maybe bits.

compiler/src/intrinsics.sk
```diff
@@ -302,6 +302,17 @@ mutable class .IntrinsicsInliner{
       };
 
       this.jumpIfInvoke(call)
+    | "Unsafe.array_ptr" ->
+      result = this.emitArrayUnsafePtr{
+        pos,
+        obj => call.args[0],
+        offset => call.args[2],
```
jberdine: Why not `args[1]` since Unsafe.sk has:
```
native fun array_ptr<T>(
  v: readonly Array<T>,
  offset: Int = 0,
): Runtime.NonGCPointer;
```

compiler/src/verify.sk
```diff
@@ -291,7 +291,6 @@ mutable private class Verifier(
 
   private mutable fun verifyBytePointerAdd(add: BytePointerAdd): void {
     this.verifyInstrType(add, tNonGCPointer);
-    this.verifyInstrType(this.getInstr(add.addr), tNonGCPointer);
```
jberdine: Does this no longer hold?

prelude/src/core/primitives/Ptr.sk
```diff
@@ -0,0 +1,21 @@
+module Ptr;
```
jberdine: Since these are inherently unsafe, would it make sense to have a scarier name like `UnsafePtr`?

prelude/src/core/primitives/String.sk
```diff
@@ -10,7 +10,7 @@ module String;
 native class .String uses Hashable, Show, Orderable {
   @cpp_runtime
   @may_alloc
-  native static fun fromChars(chars: readonly Array<Char>): String;
+  native static fun fromChars(chars: readonly Sequence<Char>): String;
```
jberdine: There must be more changes to make this work?

prelude/src/core/primitives/Slice.sk
```diff
@@ -0,0 +1,59 @@
...
+    start: Int = 0,
+    end: Int = Int::max,
+  ): readonly .Slice<T>;
+
+  overridable readonly fun values(): mutable Iterator<T> {
+    mutable SliceIterator<T>(this, 0, this.size())
+  }
+
+  // NOTE: We cannot inherit from IndexedSequence, which would give us
+  // `indexOf()` for free, for obscure reasons.
```
jberdine: I'm curious about the obscure reasons?

prelude/src/core/primitives/Array.sk
```diff
@@ -240,23 +240,23 @@ mutable native class .Array<+T>
...
+    range = Range(0, this.size()).subrange(start, end);
+    if (range.isEmpty()) {
       Array[]
     } else {
-      capacity = end - start;
-      static::fillBy(capacity, i -> this.unsafe_get(start + i))
+      capacity = range.size();
+      static::fillBy(capacity, i -> this.unsafe_get(range.start + i))
     }
   }
```
jberdine: Checking understanding: the changes to this function don't change the behavior, and hopefully compile to the same code, right?

prelude/src/native/String.sk
```diff
@@ -14,6 +14,9 @@ fun string_utf8_size(s: String): Int {
 @cpp_extern("SKIP_Unsafe_string_utf8_get")
 native fun string_utf8_get(s: String, i: Int): UInt8;
 
+@cpp_extern("SKIP_Unsafe_string_ptr")
+native fun string_ptr(s: String, i: Int): Runtime.NonGCPointer;
```
jberdine: Curious why not `Ptr`?

jberdine (commented):
I haven't looked at everything as closely as I want to, but here is a first batch of comments.

A high-level design question I'm thinking about is that the Slice type looks, from its interface, to be more general than it really is. The key way it is used seems to be using the ptr and mptr functions to get the arguments for calls to Posix.read and Posix.write. Those uses mean that the Slice type instances have an obligation to ensure that their memory representation is a contiguous array of bytes starting at the value returned by `ptr`, and that the other operations like `get` must be implemented in a way that is compatible with the underlying memory being assigned via that pointer.

I wonder if it might be just as feasible to introduce a type, say `Bytes`, that corresponds to `Slice<UInt8>`. This could be used for the args of the Posix read and write functions, and might be less likely to be mistakenly misused later. Modules like `Array`, `String`, and `Vector` would need to provide unsafe functions returning `Bytes` values referring to their internal representations, but those modules need to do similar things for `Slice` as well. Do you see what I mean? The Bytes approach seems like it would be a more limited and explicit introduction of unsafety.

compiler/runtime/posix.c
```diff
@@ -64,29 +64,20 @@ void SKIP_posix_close(int64_t fd) {
   }
 }
 
-int64_t SKIP_posix_write(int64_t fd, char *buf) {
-  int rv = write((int)fd, buf, SKIP_String_byteSize(buf));
+int64_t SKIP_posix_write(int64_t fd, char *buf, int64_t len) {
```
ltratt, line 67: Possibly silly question: is there an expectation that this code should work on non-x86/AArch64? For example, the casts in this function will have odd effects on 32 bit platforms (including, I guess, WASM?). But that might not be an issue if they're not expected to be supported.

compiler/src/IR.sk
beauby: I agree, the name `offset` is confusing here, and `index` should be used instead.

compiler/src/intrinsics.sk
beauby: That one got me debugging for a while. The reason has to do with how the compiler translates calls to native functions with default parameters: they get expanded to two parameters, the first one being a flag signalling whether a value was explicitly specified, and the second one being the actual value (whether default or explicit). So `array_ptr(foo)` gets translated (in C equivalent) to `array_ptr(addr_of_foo, 0, 0)` and `array_ptr(foo, 42)` gets translated to `array_ptr(addr_of_foo, 1, 42)`.

compiler/src/verify.sk
beauby: It does not because we want to take the address of arrays and strings, which do not have the `NonGCPointer` type.

prelude/src/core/primitives/Array.sk
beauby: Correct, I factored that bit into `Range.subrange()` because I needed the logic for each `Slice` implementation. It is expected to be semantically equivalent, though checking the generated code is a good idea.

prelude/src/core/primitives/Ptr.sk
beauby: We could put it inside the `Unsafe` module.

prelude/src/core/primitives/String.sk
beauby: Yes, this commit is not finished yet, sorry I should have marked it `[WIP]`.

prelude/src/native/String.sk
beauby: Fair point, I initially wanted to implement it as an intrinsic (may still be worth it, instead of calling into the runtime -- the current implementation was mainly for testing), in which case I think the back end would complain about types (though I'd have to check to be sure).

prelude/src/core/primitives/Slice.sk
beauby: IIRC the issue was mainly `BytesSlice` inheriting from `Slice<UInt8>`, itself inheriting from `IndexedSequence<UInt8>`, causing the typer to choke on the signature of `concat()` (defined in `IndexedSequence` as `readonly fun concat<U, S: readonly Sequence<U>>[T: U](other: S): IndexedSequence<U>`), because it seems to dislike the lower bounds on value classes (or at elast that's my understanding). It errors with a message saying that this case is not implemented in the compiler.

Also, the current definition of `MutableIndexedSequence` is a bit too broad for `MutableSlice`, because it assumes the sequence can be resized, and requires implementations for `push()`/`pop()`/`mcreate()`. It might make sense to split out the extra stuff into a `ResizableMutableIndexedSequence`, though `MutableIndexedSequence` is currently only implemented by `Vector` and `RadixTreeVector`.

compiler/src/AsmOutput.sk
beauby: This is a bit of a hack. The `"r" + optinfo.getInstr(id).idValue().toString()` part effectively generates the name of the register where the code in `intrinsics.sk` expects the result of the computation.

There may be a cleaner way to generate the register name, instead of manually generating it here, but the code above expects a `String`.

compiler/src/AsmOutput.sk
beauby: To be frank I have no idea, but I'd like to know as well.

compiler/src/AsmOutput.sk
beauby: I was assuming that for any array we would be in the "non-bitfield case" below, hence the `if()` would not be executed. I may be mistaken though. What's your take?

compiler/runtime/posix.c
beauby: At this point, there really isn't any guarantee anything works beyond amd64/aarch64/wasm32. The runtime could use quite a bit of clean up though, I concede!

beauby:
> A high-level design question I'm thinking about is that the Slice type looks, from its interface, to be more general than it really is. The key way it is used seems to be using the ptr and mptr functions to get the arguments for calls to Posix.read and Posix.write. Those uses mean that the Slice type instances have an obligation to ensure that their memory representation is a contiguous array of bytes starting at the value returned by ptr, and that the other operations like get must be implemented in a way that is compatible with the underlying memory being assigned via that pointer.

While you are right that the current main use is for IO, it can also be used to handle sub-sequences without copying objects (for instance `some_array.as_slice().subslice(i, j)` should be substituable to `some_array.slice(i, j)` saving an allocation + copy). I am not sure it makes sense to generalize the concept of "sub-sequence without copy" to containers with non-contiguous memory representations.


> I wonder if it might be just as feasible to introduce a type, say Bytes, that corresponds to Slice<UInt8>. This could be used for the args of the Posix read and write functions, and might be less likely to be mistakenly misused later. Modules like Array, String, and Vector would need to provide unsafe functions returning Bytes values referring to their internal representations, but those modules need to do similar things for Slice as well. Do you see what I mean? The Bytes approach seems like it would be a more limited and explicit introduction of unsafety.

As far as IO is concerned, this would indeed be sufficient. One caveat to keep in mind is that any `Array<T>` would expose a `bytes(): Bytes`, whereas currently only `Array<UInt8>` exposes `slice(): Slice<UInt8>` (which prevents calling `Posix.read` on an array of non-value classes for instance).

compiler/runtime/posix.c
ltratt: I think even on wasm32, we have to be careful about the `(int64_t) -> (size_t)` cast(s) as that will do something surprising if someone asks for more than 4GiB, silently truncating. [I've been bitten by this sort of thing in the past so I'm a bit paranoid!]

prelude/src/core/primitives/Slice.sk
beauby: The actual error message is:
```
File "/home/beauby/Code/skfs/prelude/src/core/primitives/String.sk", line 574, characters 18-22:
ICE TODO This constraint adds a lower bound to a type parameter which is not yet supported
572 |   private str: String,
573 |   private range: Range,
574 | ) extends .Slice<UInt8> {
    |                  ^^^^^
575 |   fun ptr(): Ptr<UInt8> {
```

and it seems to originate from [this place](https://github.com/SkipLabs/skdb/blob/main/compiler/src/skipTypingUtils.sk#L2307).

compiler/runtime/posix.c
beauby: That's a fair point, though `Array`'s size is stored as a 32bit integer, so it shouldn't be an issue (for now).

jberdine: In that case, wouldn't it be better in terms of ease of future debugging to declare the len arg as an int32_t?

compiler/src/AsmOutput.sk
jberdine: That may be true for the array case, but isn't this code now used in the general case as well due to the call on line 2782 below?

compiler/src/intrinsics.sk
jberdine: Ugh. Ok, I see. I think a comment briefly explaining this would be good.

compiler/src/AsmOutput.sk
jberdine: I see. I wonder if it would be easy enough even to just have a function from instr to string that did this string concatenation, to document the connection between this use and those in intrinsics.sk?

compiler/src/verify.sk
jberdine: I wonder if there is a way to relax the assertion without completely removing it. Not a huge deal.

prelude/src/core/primitives/Slice.sk
jberdine: Ah, I see. Ok, that is something to dodge for now then. Maybe it would help to change the comment to say something about when constraints that add lower bounds are supported, then this could change.

compiler/runtime/posix.c
ltratt: What I've done in the past is put the casts into a function/macro: it can be infallible for now, and not too verbose, and if this ever becomes an issue in the future it can become a fallible check. [Warning: not only am I paranoid about this but I am also a bore. https://tratt.net/laurie/blog/2021/static_integer_types.html goes into too much detail about these challenges!]

jberdine (commented):
> I am not sure it makes sense to generalize the concept of "sub-sequence without copy" to containers with non-contiguous memory representations.

I don't know. Many balanced tree data structures support an efficient operation to split a tree into two such that one contains the keys <= a given one and the other contains the rest. Using two calls to such an operation gives an efficient way to obtain a slice of a Map. Or thinking of the sequences of a Map's keys, values, or bindings, it would also enable efficient implementations. Another example would be large strings, or string buffers, implemented using a representation based on ropes, or other unbalanced trees. Those generally support identifying subranges/slices, but are not represented as contiguous byte sequences.

So I maintain that conceptually the notion of a subsequence of a sequence that is defined by a range of indices/keys is a distinct concept from the notion of a byte sequence view of the underlying representation of some higher-level sequence. For posix IO, it is the second sort that is needed. In the more general first case, it is not clear that the `ptr` operation is needed/useful. So perhaps the most expressive path is to have Slice that is like what is in this PR, but without `ptr` and so would be safe, with a specialization for those cases that the slice is a subsequence of the byte array memory representation, with additional `ptr` operation, and only this specialized subtype would need to be unsafe. WDYT?

gregsexton:
> So I maintain that conceptually the notion of a subsequence of a sequence that is defined by a range of indices/keys is a distinct concept from the notion of a byte sequence view of the underlying representation of some higher-level sequence.

Makes sense. I would design this as a bunch of types that implement each original data structure's interface. e.g. a 'subset map' type that implements map and works in terms of a key range making use of the original underlying persistent datastructure. The original map can hand out a 'subset map' that references the node structure.

And I think this is what slice is to array?

beauby:
> So I maintain that conceptually the notion of a subsequence of a sequence that is defined by a range of indices/keys is a distinct concept from the notion of a byte sequence view of the underlying representation of some higher-level sequence. [...] So perhaps the most expressive path is to have Slice that is like what is in this PR, but without ptr and so would be safe, with a specialization for those cases that the slice is a subsequence of the byte array memory representation, with additional ptr operation, and only this specialized subtype would need to be unsafe. WDYT?

Yup, that makes perfect sense. Will update the PR.

compiler/runtime/posix.c
beauby: > In that case, wouldn't it be better in terms of ease of future debugging to declare the len arg as an int32_t?

My understanding is that while the array length itself is stored in memory as an `int32_t`, it is exposed as an `Int` in (native) skip binaries, which always use `int64_t`.

In any case, the POSIX interface in general, and this file (`posix.c` in particular) are never compiled for wasm.

> What I've done in the past is put the casts into a function/macro

I agree that it would be best (and there is actually a `SkipInt` typedef that is being used sporadically). I'll try to do a clean up pass on the runtime soon™.

compiler/src/AsmOutput.sk
beauby: You are completely right, I got carried away. I think I'll implement it as a runtime function first, and we can make a proper intrinsic implementation later, as there are too many shady things in the current implementation.

compiler/runtime/posix.c
jberdine: Ok yeah, makes sense since it's exposed as a skip int.

Sync with me before making a broad pass, I'm accumulating some cleanup diffs for the runtime too.

compiler/runtime/runtime.c
```diff
@@ -181,3 +181,26 @@ void* SKIP_context_sync(uint64_t txTime, char* old_root, char* delta,
   SKIP_call_after_unlock(synchronizer, delta);
   return new_root;
 }
+
+int64_t SKIP_Unsafe_Ptr__toInt(char *ptr) {
```
jberdine:
```suggestion
int64_t SKIP_Unsafe_Ptr__toInt(char* ptr) {
```

prelude/src/core/primitives/Bytes.sk
```diff
@@ -0,0 +1,57 @@
...
+  readonly fun size(): Int;
+
+  readonly fun values(): mutable Iterator<UInt8> {
+    mutable BytesIterator(this)
+  }
+
+  // TODO: extends IndexedSequence<UInt8>
+  readonly fun get(index: Int): UInt8;
+  readonly fun slice(start: Int, end: Int = Int::max): readonly Bytes;
+  readonly fun indexOf(predicate: UInt8): ?Int {
```
jberdine, line 38: I find `predicate` to be a curious name, would expect some function of one arg returning a bool. Maybe `needle` or `mark`?

prelude/src/core/Unsafe.sk
```diff
@@ -47,6 +47,26 @@ native value class RawStorage<+T> {
...
+  readonly fun toString(): String {
+    digits = mutable Vector[];
+    n = this.toInt();
+    while (n.ugt(0)) {
+      digits.push(Chars.intToHexDigit(n.and(0xF)));
+      !n = n.shr(4)
+    };
+    if (digits.isEmpty()) {
+      "(null)"
+    } else {
```
jberdine: I don't know if we'll be happy or sad about this special case down the road.

prelude/src/stdlib/other/Range.sk
```diff
@@ -36,6 +36,20 @@ class .Range(start: Int, end: Int) extends Sequence<Int> uses Equality, Show {
     Range(start, end)
   }
 
+  fun subrange(start: Int, end: Int = Int::max): this {
```
jberdine, line 44: Can we document the behavior of negative values for start and end? It is especially needed since the Range type itself does not permit them, and it seems reasonable to expect that subrange(s,e) needs to satisfy the same conditions as Range(s,e), but it doesn't.

prelude/src/stdlib/other/Range.sk
```diff
@@ -36,6 +36,20 @@ class .Range(start: Int, end: Int) extends Sequence<Int> uses Equality, Show {
...
+    sz = this.size();
+    if (start < 0) {
+      !start = max(0, sz + start);
+    };
+    if (end < 0) {
+      !end = max(0, sz + end);
+    };
+    !start = this.start + min(sz, start);
+    !end = this.start + min(sz, end);
```
jberdine, line 53: Array.slice normalizes all empty slices to Array[], so perhaps best to do similar here.

prelude/src/core/primitives/String.sk
```diff
@@ -420,6 +420,10 @@ native class .String uses Hashable, Show, Orderable {
     StringIterator::make(this)
   }
 
+  fun bytes(): readonly Bytes {
+    StringBytes(this, Range(0, Unsafe.string_utf8_size(this)))
```
jberdine, line 424: Not about this diff, but it messed with my head that `string_utf8_size` returns the size in bytes of the string argument, and has nothing to do with utf8 at all. WTF.

prelude/src/core/primitives/String.sk
```diff
@@ -552,3 +556,24 @@ base class Encoding {
...
+private mutable class StringBytes(
+  private str: String,
+  private range: Range,
+) extends Bytes {
+  readonly fun ptr(): Unsafe.Ptr<UInt8> {
+    Unsafe.Ptr<UInt8>(Unsafe.string_ptr(this.str, this.range.start))
+  }
+
+  readonly fun get(index: Int): UInt8 {
+    Unsafe.string_utf8_get(this.str, this.range.start + index)
```
jberdine, line 569: Here too, string_utf8_get loads a byte, it has nothing to do with utf8.

compiler/src/skfmt.sk
```diff
@@ -42,7 +42,13 @@ untracked fun main(): void {
   for (file in filenames) {
     contents = if (file == "-") {
       !file = "<stdin>";
-      IO.stdin().read_to_end()
+      buf = mutable Vector[];
+      IO.stdin().read_to_end(buf) match {
+      | Success _ -> String::fromUtf8(buf.collect(Array))
+      | Failure(err) ->
+        print_error(err.getMessage());
+        skipExit(1)
```
jberdine, line 50: I wonder if the IO.Error type should have a wrapper that takes a Result<Int, Error> and does this pattern match and error report. Maybe that would enable deduping a lot of error handling boilerplate code, and report errors consistently.

prelude/src/stdlib/collections/mutable/Vector.sk
```diff
@@ -246,6 +246,15 @@ mutable class .Vector<+T> private (
     this.invalidateIterators();
   }
 
+  mutable fun unsafeSetSize(newSize: Int): void {
```
jberdine: I wonder if a more focused operation such as unsafeConsumeSpareCapacity would be better. It could take a number of bytes, check that there are at least that many that are currently spare, and advance the size by that many.

prelude/src/stdlib/other/IO.sk
```diff
@@ -34,68 +34,108 @@ class OpenOptions{
...
+      }
+    };
+
+    Success(nbytes)
+  }
+
+  mutable fun read_to_string(): Result<String, Error> {
+    buf = mutable Vector<UInt8>[];
+    this.read_to_end(buf) match {
+    // TODO: Get rid of `.collect(Array)`.
```
jberdine, line 68: Just to check my understanding, doing this would involve adding Vector operations to produce Bytes and adding String operations to construct from Bytes? Or did you have something else in mind?

prelude/src/stdlib/other/IO.sk
```diff
@@ -34,68 +34,108 @@ class OpenOptions{
...
+}
+
+trait Read {
+  mutable fun read(buf: mutable MutableBytes): Result<Int, Error>;
+
+  mutable fun read_to_end(buf: mutable Vector<UInt8>): Result<Int, Error> {
+    nbytes = 0;
+    loop {
+      if (buf.size() == buf.capacity()) {
+        buf.ensureCapacity(buf.capacity() + 32)
```
jberdine, line 51: 32 seems like a magic constant, and seems very small too. I would expect something like exponential increase in the capacity added each time until hitting something like at least 4KB, maybe as high as 16MB, and then flat from there.

prelude/src/stdlib/other/IO.sk
```diff
@@ -34,68 +34,108 @@ class OpenOptions{
...
-      !res = res + s
-    };
-    res
-  }
-
-  mutable fun write(s: String): Int {
-    Posix.write(this.fileno, s)
+  mutable fun read(buf: mutable MutableBytes): Result<Int, Error> {
+    rv = Posix.read(this.fileno, buf.mptr(), buf.size());
+    if (rv >= 0) {
```
jberdine, line 162: Just to double-check, 0 is a success case here since if the underlying read e.g. hit EOF, then it would be detected as a failure via a -1 size read, correct?

prelude/src/stdlib/other/IO.sk
```diff
@@ -34,68 +34,108 @@ class OpenOptions{
...
-    loop {
-      nbytes = Posix.write(this.fileno, s);
-      // TODO: This is broken for wide utf8 chars.
-      if (nbytes == s.length()) {
-        break void
-      };
-      !s = s.substring(s.getIter().forward(nbytes))
+  mutable fun write(buf: readonly Bytes): Result<Int, Error> {
+    rv = Posix.write(this.fileno, buf.ptr(), buf.size());
+    if (rv >= 0) {
```
jberdine, line 171: Same here about 0 as for read above.

prelude/src/stdlib/other/IO.sk
```diff
@@ -105,90 +145,113 @@ mutable class File(
...
     false
   }
 
-  mutable fun read(sz: Int): String {
-    start = this.buffer.getIter().forward(this.pos);
-    if (sz == 0) {
-      return ""
+  mutable fun read(buf: mutable MutableBytes): Result<Int, Error> {
+    readSz = min(buf.size(), this.buffer.size() - this.pos);
+    // TODO: Proper copy.
```
jberdine, line 207: This needs Vector::asBytes and a memcpy operation on Bytes, right?

prelude/src/stdlib/other/IO.sk
```diff
@@ -105,90 +145,113 @@ mutable class File(
 }
 
 mutable class StringIO(
```
jberdine, line 197: This class is dead isn't it?

prelude/src/stdlib/other/IO.sk
```diff
@@ -105,90 +145,113 @@ mutable class File(
...
-    this.!buffer = this.buffer + s;
+trait BufRead extends Read {
+  mutable fun fill_buf(): Result<readonly Bytes, Error>;
+  mutable fun consume(nbytes: Int): void;
+
+  mutable fun has_data_left(): Result<Bool, Error> {
+    this.fill_buf() match {
+    | Success(buf) -> Success(buf.size() > 0)
+    | Failure(err) -> Failure(err)
+    }
```
jberdine: If you like:
```suggestion
    this.fill_buf().map(buf -> buf.size() > 0)
```

prelude/src/stdlib/other/IO.sk
```diff
@@ -105,90 +145,113 @@ mutable class File(
...
       }
-    } catch {
-    | EndOfFile _ if (this.buffer.length() > 0) -> void
-    | exn -> throw exn
     };
-    res = this.buffer.sub(this.buffer.getIter(), pos);
-    this.!buffer = this.buffer.substring(this.buffer.getIter().forward(pos));
-    res
+
+    Success(this.buffer.bytes().slice(this.available.start, this.available.end))
```
jberdine, line 297: Maybe nice to have a slice operation that takes a range directly.

jberdine (approved):
Sweet! A few comments inline, but nothing major.

prelude/src/core/primitives/Bytes.sk
beauby: The signature is taken from [IndexedSequence.sk](https://github.com/SkipLabs/skdb/blob/main/prelude/src/stdlib/collections/interfaces/IndexedSequence.sk#L62). I find the name `predicate` confusing as well, but figured I'd favour consistency for now.

prelude/src/core/Unsafe.sk
beauby: A special case is needed (by default we would just get `0x`), though I don't mind making it `0x0` if you prefer.

prelude/src/stdlib/other/IO.sk
beauby: No actually `EOF` is reported as a size 0 read, and is not an error.

prelude/src/stdlib/other/IO.sk
beauby: It is currently not being used, yes. I don't mind getting rid of it.

prelude/src/stdlib/other/IO.sk
beauby: Yes that would be the most straightforward.

prelude/src/stdlib/other/IO.sk
beauby: I think that won't work because the type of `this.fill_buf()` is `Result<readonly Bytes, Error>` and we want to return a `Result<Bool, Error`, hence the somewhat cumbersome `Failure(err) -> Failure(err)`.

prelude/src/stdlib/other/IO.sk
beauby: I thought about it, but the current APIs supporting `slice()` take two `Int`s so I favoured consistency.

prelude/src/stdlib/other/Range.sk
beauby: What normalization did you have in mind? `Range(0, 0)` or `Range(start, start)`?

prelude/src/stdlib/other/IO.sk
beauby: That's correct.

prelude/src/stdlib/other/IO.sk
beauby: The `ensureCapacity()` method will figure it out and do the right thing:
```
 192 │ // Grow the Vector's internal storage as necessary to accommodate the                                                                                                                             
 193 │ // given (absolute) capacity. This method *may* allocate more storage                                                                                                                             
 194 │ // than requested: if the method succeeds the Vector's capacity will                                                                                                                              
 195 │ // be equal or greater than the given capacity.                                                                                                                                                   
 196 │ mutable fun ensureCapacity(capacity: Int): void {                                                                                                                                                 
 197 │ │ if (capacity > this.capacity()) {                                                                                                                                                               
 198 │ │ │ // Allocate new memory. If the new capacity is close enough to exponential                                                                                                                    
 199 │ │ │ // growth, grow to it. Otherwise use the normal growth heuristic to                                                                                                                           
 200 │ │ │ // avoid O(n^2) behavior (e.g. from calling ensureCapacity(n+1) in a loop).                                                                                                                   
 201 │ │ │ if (capacity < (this.sz * 5).ushr(2)) {                                                                                                                                                       
 202 │ │ │ │ !capacity = getCapacityForSize(capacity)                                                                                                                                                    
 203 │ │ │ };                                                                                                                                                                                            
 204 │ │ │ this.unsafeGrowCapacity(capacity)                                                                                                                                                             
 205 │ │ } else {
```

prelude/src/core/primitives/String.sk
beauby: Agreed, this is poorly named.

prelude/src/stdlib/collections/mutable/Vector.sk
beauby: That's a good idea.

prelude/src/core/Unsafe.sk
jberdine: Oh yeah, I see. I would go with `0x0` personally, but whatever you like.

prelude/src/stdlib/other/IO.sk
jberdine: That is a change to the external spec of the read function then, right? Maybe that should be advertised more loudly, at least called out in the PR summary?

prelude/src/stdlib/other/IO.sk
jberdine: Either way is ok, I was just looking around for its uses since it seemed odd to have the operations returning Results that were never Failures, and it made me wonder what the uses in context looked like.

prelude/src/stdlib/other/IO.sk
jberdine: That's unfortunate given that `fun map<T2>(f: T -> T2): Result<T2, E>` is meant to be polymorphic enough. Do you know if it is the readonly qualifier that causes trouble?

prelude/src/stdlib/other/Range.sk
jberdine: I would go for `Range(0,0)` to avoid having many different representations of the same semantic value, the empty interval.

prelude/src/core/primitives/Bytes.sk
jberdine: Yes makes sense. I think it really just means that IndexedSequence is also wrong though. :-)

prelude/src/stdlib/other/IO.sk
jberdine: Ok, that is pretty good. Note though that buffers meant for reading files are ideally a different size than those for writing, and general case vectors are another case entirely. For instance, ideally we would want to get to a steady state where we issue read calls with fixed size of something like 4MB to get good IO speed.

compiler/src/skfmt.sk
beauby: Ideally I'd rather have syntactic sugar for 

```
x = foo match {
| Success(y) -> y
| Failure(err) -> return Failure(err)
}
```

jberdine:
I guess that to make CI happy you need to do some bootstrap dance, or are there other known issues?

compiler/runtime/runtime.c
```diff
@@ -129,3 +129,26 @@ void* SKIP_context_sync(uint64_t txTime, char* old_root, char* delta,
...
+  return (int64_t)ptr;
+}
+
+void* SKIP_Unsafe_array_ptr(char* arr, int64_t offset) {
+  return arr + offset;
+}
+
+int64_t SKIP_Unsafe_array_byte_size(char* arr) {
+  SKIP_gc_type_t* ty = *(*(((SKIP_gc_type_t***)arr) - 1) + 1);
+  size_t len = *(uint32_t*)(arr - sizeof(char*) - sizeof(uint32_t));
```
mbouaziz:
```suggestion
  SKIP_gc_type_t* ty = get_gc_type(arr);
  size_t len = skip_array_len(arr);
```

compiler/runtime/runtime.c
```diff
@@ -129,3 +129,26 @@ void* SKIP_context_sync(uint64_t txTime, char* old_root, char* delta,
   SKIP_call_after_unlock(synchronizer, delta);
   return new_root;
 }
+
+int64_t SKIP_Unsafe_Ptr__toInt(char* ptr) {
+  return (int64_t)ptr;
+}
+
+void* SKIP_Unsafe_array_ptr(char* arr, int64_t offset) {
```
mbouaziz:
```suggestion
void* SKIP_Unsafe_array_ptr(char* arr, SkipInt offset) {
```
I guess this would document the fact that the value comes from skip.

(similarly at other places where it makes sense)

prelude/src/core/Unsafe.sk
```diff
@@ -62,4 +82,19 @@ native fun array_get<T>(v: readonly Array<T>, i: Int): T;
 @intrinsic
 native fun array_set<T>(v: mutable Array<T>, i: Int, x: T): void;
 
+@cpp_extern
+native fun array_ptr<T>(v: readonly Array<T>, i: Int): Runtime.NonGCPointer;
```
mbouaziz:
```suggestion
native fun array_ptr<T>(v: readonly Array<T>, byte_offset: Int): Runtime.NonGCPointer;
```
otherwise, I may think it's an offset of `T`s.
Or change the function name if you can think of a better one.

prelude/src/core/primitives/String.sk
mbouaziz: I agree, it should be renamed `string_byte_size`

beauby: I agree as well.

PR #14, opened 2023-12-07
update circlci "test determinator" logic

bennostein:


.circleci/generate_config.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/bash
 
-# TODO: This is flaky as it relies on the fork's master branch state.
-git diff --quiet HEAD master -- compiler/ prelude/ skfs/
+# TODO: This is flaky as it relies on the fork's main branch state and
```
beauby: Actually I believe this is not the case (CircleCI checks out `SkipLabs/skdb.git`). My bad for initially writing that comment ;)

beauby (approved):

PR #17, opened 2023-12-11
Minor runtime cleanups related to failure exits

jberdine:


compiler/runtime/free.c
```diff
@@ -41,7 +42,7 @@ void sk_free_class(sk_stack_t* st, char* obj) {
 #ifdef SKIP64
     if (!sk_is_nofile_mode()) {
       fprintf(stderr, "You cannot use external pointers in persistent mode.\n");
-      _exit(23);
```
beauby, line 44: Do we care about the `exit()`/`_exit()` distinction? And should this be a `SKIP_exit()`?

compiler/runtime/free.c
```diff
@@ -50,7 +51,7 @@ void sk_free_class(sk_stack_t* st, char* obj) {
     if (magic != 234566 && magic != 234567) {
 #ifdef SKIP64
       fprintf(stderr, "Invalid external pointer found\n");
-      _exit(24);
```
beauby, line 53: Same as above.

compiler/runtime/memory.c
```diff
@@ -161,14 +161,10 @@ int memcmp(const void* ptr1, const void* ptr2, size_t num) {
...
-    perror("malloc");
-    exit(1);
-  }
   sk_lower_static(result);
 
   if (result == NULL) {
-    fprintf(stderr, "Out of memory\n");
+    perror("malloc");
     SKIP_throw_cruntime(ERROR_OUT_OF_MEMORY);
   }
```
beauby: Should this be moved above the `sk_lower_static()` call?

compiler/runtime/runtime64_specific.cpp
```diff
@@ -671,40 +674,6 @@ void SKIP_exit(uint64_t code) {
...
-    ;
-  exit(0);
-  return NULL;
-}
-
-void SKIP_unix_die_on_EOF() {
-  pthread_t thread_id;
-  pthread_create(&thread_id, NULL, &wait_for_EOF, NULL);
-}
-
```
beauby, line 707: Sweet.

beauby (commented):
Nice, this has been bugging me for a while. Only questions are:
- does the `_exit()`/`exit()` distinction matter in the current runtime? 
- should we use `SKIP_exit()` everywhere rather than `exit()`?

compiler/runtime/free.c
jberdine: Did you see the commit message: https://github.com/SkipLabs/skdb/pull/17/commits/73a53599d28f60f2a22f74adef0ec13a58c4e485

compiler/runtime/memory.c
jberdine: It could go either way. `sk_lower_static` won't change `result`, so the whole process will crash either way. I went with the current code since this error check is older than the one above. Happy to change the order though.

compiler/runtime/free.c
beauby: My bad, I hadn't read it.

compiler/runtime/memory.c
beauby: I think both are fine.

compiler/runtime/free.c
jberdine: No worries, just wondered if that wasn't clear or just not seen.

compiler/runtime/memory.c
jberdine: I think doing the error check directly after malloc is clearer, so change it.

jberdine:
Regarding whether to call `SKIP_exit` instead of `exit`, I'm not sure. My understanding is that `SKIP_exit` is mainly to mediate between the signature needed to be called from skip code and whatever is used in the C stdlib. For now it is just a cast, but maybe it will need to do more if we change the skip exit function. Also, we'd need to include the runtime.h header everywhere. Possible, but I don't know.

gregsexton:
Sheep it!

PR #18, opened 2023-12-11
Reduce write-csv memory usage by introducing 'GC'

gregsexton:
I wrote a test script that inserts 50k rows at once. RSS used is 587Mb
under node. After these changes it is 285Mb. 51% reduction in peak
memory.

More importantly I tried running larger inputs, e.g. 200k rows and it
survives and processes this. Previously it would crash where it runs
out of address space.

In summary: by managing the GC carefully in the row iterations we
ensure the processing overhead is O(1). We still build all the rows
though.

This also extends the soak running time from 18hrs to 21. We're now
hitting the db size limit which after 21hrs is > 1gb. Will work on
making us more memory efficient here next.

### DIFF
--- a/sql/src/SqlCsv.sk
+++ b/sql/src/SqlCsv.sk
@@ -350,7 +350,7 @@ fun applyDiffStrategy(
 
     dir = newRoot.unsafeGetEagerDir(table.dirName);
 
-    !dir = SKStore.withRegionFold(
+    !dir = SKStore.withRegionFoldRec(
       Some(newRoot),
       rows.iterator(),
       dir,
--- a/sql/src/SqlCsv.sk
+++ b/sql/src/SqlCsv.sk
@@ -355,11 +355,12 @@ fun applyDiffStrategy(
       rows.iterator(),
       dir,
       (optContext, row, newDir) ~> {
+        ctx = optContext.fromSome();
         key: SKStore.Key = SKDB.RowKey(row.setRepeat(1), table.kinds);
         for ((tick, source, _) in newDir.unsafeGetDataIterWithoutTombs(key)) {
           if (tick.value < snapshot.default(0) || source.path() == writer) {
             !newDir = newDir.writeEntry(
-              optContext.fromSome(),
+              ctx,
               source.path(),
               writer,
               key,
@@ -370,13 +371,19 @@ fun applyDiffStrategy(
 
         if (row.repeat > 0) {
           !newDir = newDir.writeEntry(
-            optContext.fromSome(),
+            ctx,
             writer,
             writer,
             SKDB.RowKey(row, table.kinds),
             Array[row],
           );
         };
+
+        // update incrementally to keep memory usage down. this has a
+        // significant effect for large write sets and negligible
+        // impact on performance
+        ctx.setDir(newDir.dirName, newDir);
+        ctx.update();
 
         newDir
       },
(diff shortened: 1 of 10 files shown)


gregsexton:
I haven't yet measured the effect on perf this has. We're allocating and destroying obstacks a lot now. I suspect it won't be too bad, but if it is I'll introduce some batching.

sql/src/SqlCsv.sk
```diff
@@ -555,6 +625,8 @@ fun parseTableChange(
   table: SKDB.DirDescr,
   line: String,
 ): ?Array<SKDB.RowValues> {
+  saved = SKStore.newObstack();
```
jberdine: I couldn't convince github to let me suggest changes, so I pushed a commit to your branch with a suggestion to use withRegion here instead of manually saving and restoring the obstack.

gregsexton: Oh, could probably use that in the other places too. I might have misunderstood: I thought it only copied out the context.

jberdine: It does something quite similar to what you had, but also protects against exceptions, etc.:
```
fun withRegion<T>(
  context: mutable Context,
  f: (Obstack, mutable Context) ~> T,
): T {
  optSaved: ?Obstack = None();
  try {
    !optSaved = Some(newObstack());
    saved = optSaved.fromSome();
    fresult = f(saved, context);
    copied = destroyObstackWithValue(saved, List[(context.clone(), fresult)]);
    (icontext, copiedResult) = copied.getHead();
    context.replaceFromSaved(icontext);
    copiedResult
  } catch {
  | exn ->
    optSaved match {
    | Some(saved) ->
      cexn = destroyObstackWithValue(saved, List[exn]).getHead();
      throw cexn
    | _ -> throw exn
    }
  }
}
```

jberdine: `withRegion` does something different in the `destroyObstackWithValue` call, passing in the context and then later calling `replaceFromSaved`. I haven't fully understood the consequences of that difference yet.

gregsexton:
> I haven't yet measured the effect on perf this has. We're allocating and destroying obstacks a lot now. I suspect it won't be too bad, but if it is I'll introduce some batching.

Life is full of surprising wins. Turns out that GCing frequently improves perf. I get a stat sig 16% win. (95%CI: 13.9 -- 18.5%).

CE9F08D3-30D7-4CF2-8B3B-F58FB3DCFC7D

gregsexton:
I've rebased this on #19 and started using withRegionFold. Results are great:

RSS drops 12% in native 23% wasm. Wasm also runs faster.

@jberdine, I rebased your commit and tested this too. Unfortuantely this regresses wasm memory by 13%. I think it's because we now copy the context for every row which I was avoiding before. I see that Daniel introduced withRegionValue. Might move to this and add some exn handling there which is a nice win from your change.

jberdine:
Yes, the regression with my commit is what I meant when I mentioned last week that I thought my suggestion would be a bad idea. It makes sense, but independent of this diff, I'm overall not fully satisfied that, at least to me, it isn't entirely obvious when to copy the context and when it can be safely avoided. Feels like a missing abstraction or automation or something. Just thinking out loud.

gregsexton:
I've been grinding away at this for too long. I'm going to draw a line under it now and move on, but I think there are some more wins to be found in EagerDir.flattenData.

I just measured the latest results of this stack against my test setup:

origin/main (37518c19):
- Peak RSS processing 100k rows on native: 1324 Mib
- Max input rows before crashing (wasm): 90k

branch tip (c817eb43):
- Peak RSS processing 100k rows on native: 187 Mib 
- Max input rows before crashing (wasm): 675k

gregsexton:
I haven't measured wall clock very scientifically. FWIW the runs for 100k were 1590ms for origin/main and 1050ms for the branch. These were dev builds though, so probably not too significant. The important thing is the memory wins aren't at the cost of perf.

jberdine:
This is awesome!

PR #19, opened 2023-12-14
[skstore] Add garbage collection on update

skiplabsdaniel:


skfs/src/Context.sk
```diff
@@ -1626,4 +1626,31 @@ fun withRegionValue<T>(f: () ~> T): T {
...
+        saved,
+        List[value],
+      ).getHead();
+      !optSaved = Some(SKStore.newObstack());
+    }
+  };
+  SKStore.destroyObstackWithValue(
+    optSaved.fromSome(),
+    List[valueInRegion],
+  ).getHead();
```
jberdine: withRegion protects against exceptions, I guess this ought to in a similar way right?

compiler/runtime/obstack.c
```diff
@@ -272,8 +272,18 @@ void SKIP_enable_GC() {
...
+  size_t size = 0;
+  void* cursor = page;
+  while (cursor != NULL && cursor != saved->page) {
+    cursor = *(char**)cursor;
+    nbr_page++;
+    if (nbr_page > 3 || (nbr_page > 1 && (head - page > (2 * PAGE_SIZE / 3)))) {
+      return 1;
+    }
+  }
+  return 0;
```
jberdine, line 274: This makes the decision almost not depend on `saved`. Is there any description of the intuition for the new heuristic?

skfs/src/Context.sk
```diff
@@ -1626,4 +1626,31 @@ fun withRegionValue<T>(f: () ~> T): T {
   copied.getHead();
 }
 
+fun withRegionIterate<K, T>(
```
jberdine: Since `fn` takes and returns an accumulator, this more of a `fold` than an `iterate`.

skfs/src/EagerDir.sk
```diff
@@ -1576,105 +1577,130 @@ class EagerDir{
...
+    it = dirty.iterator();
+    dKeys = Array<Key>::createFromIterator(it);
+    index = 0;
+    size = dKeys.size();
+    while (index < size) {
+      indexCopy = index;
+      childCopy = childRef;
+      (!index, !childRef) = withRegion(context, (obstack, context) ~> {
+        idx = indexCopy;
+        child = childCopy;
```
jberdine: I haven't thought through in detail, but would it be possible to simplify this dance between e.g. `index`, `indexCopy`, and `idx` using your new `withRegionIterate`?

jberdine (approved):
LGTM, a few minor comments inline.

gregsexton:
I will move the changes in #18 to use withRegionIterate once this lands - this abstraction works well there too! :)

skfs/src/Context.sk
gregsexton: Ah, I was immediately comfortable here, follows the same naming pattern as e.g. https://clojuredocs.org/clojure.core/iterate

skfs/src/Context.sk
gregsexton: Yes please. This is my main reason for wanting to replace my manual obstack code in write-csv with something like this.

skfs/src/Context.sk
gregsexton: I take this back. I misread the fun sig. It's a fold. :)

compiler/runtime/obstack.c
skiplabsdaniel: The previous on return always true. It's a try, if you have some intuition on it, you're welcome.

skfs/src/Context.sk
```diff
@@ -1626,4 +1626,52 @@ fun withRegionValue<T>(f: () ~> T): T {
...
+  contextOpt: ?mutable Context,
+  it: mutable Iterator<K>,
+  value: T,
+  fn: (?mutable Context, K, T) ~> T,
+): T {
+  optSaved: ?Obstack = None();
+  valueInRegion = value;
+  loop {
+    if (
+      try {
```
jberdine: Am I confused, or could the whole `loop` be in the body of the `try`? I'm not entirely sure of the compilation scheme, but doing that might save some exception setup and cleanup on each loop iteration.

jberdine (commented):
Nit: Note misspelling in the commit message.

PR #21, opened 2023-12-14
[sql] Bring back support for `foo IN (SELECT ...)`.

beauby:


jberdine (approved):
Makes me wonder why it wasn't done before :-)

PR #22, opened 2023-12-15
SKDBGroup API for JS client

bennostein:
And some related fixes/changes.

The new SKDBGroup abstraction should be self-explanatory, but I would appreciate eyes in particular on the fifth commit here -- it's a small change, but that fixes the bug I was chasing for the last week or two.

### DIFF
--- a/sql/src/Main.sk
+++ b/sql/src/Main.sk
@@ -706,33 +706,14 @@ fun execWriteCsv(args: Cli.ParseResults, options: SKDB.Options): void {
   | None() -> SKStore.genSym(0)
   };
 
-  // block until we receive the first line of input, so that the subsequent
-  // `runSql` call executes under an up-to-date context.
-  firstLine = {
-    buffer = mutable Vector[];
-    loop {
-      c = getChar();
-      if (c == '\n') break String::fromChars(buffer.toArray());
-      buffer.push(c)
-    }
-  };
-
   SKDB.runSql(options, context ~> {
     user = args.maybeGetString("user");
-    isFirst = true;
     myReadLine = () -> {
-      if (isFirst) {
-        !isFirst = false;
-        return firstLine;
-      };
       buffer = mutable Vector[];
       loop {
         c = getChar();
-        if (c == '\n') {
-          if (buffer.size() > 0) return String::fromChars(buffer.toArray());
-        } else {
-          buffer.push(c)
-        };
+        if (c == '\n') return String::fromChars(buffer.toArray());
+        buffer.push(c);
       }
     };
     SKCSV.replayDiff(context, myReadLine, user, source)
--- a/sql/src/Main.sk
+++ b/sql/src/Main.sk
@@ -705,14 +705,34 @@ fun execWriteCsv(args: Cli.ParseResults, options: SKDB.Options): void {
   | Some(src) -> src.toInt()
   | None() -> SKStore.genSym(0)
   };
+
+  // block until we receive the first line of input, so that the subsequent
+  // `runSql` call executes under an up-to-date context.
+  firstLine = {
+    buffer = mutable Vector[];
+    loop {
+      c = getChar();
+      if (c == '\n') break String::fromChars(buffer.toArray());
+      buffer.push(c)
+    }
+  };
+
   SKDB.runSql(options, context ~> {
     user = args.maybeGetString("user");
+    isFirst = true;
     myReadLine = () -> {
+      if (isFirst) {
+        !isFirst = false;
+        return firstLine;
+      };
       buffer = mutable Vector[];
       loop {
         c = getChar();
-        if (c == '\n') return String::fromChars(buffer.toArray());
-        buffer.push(c);
+        if (c == '\n') {
+          if (buffer.size() > 0) return String::fromChars(buffer.toArray());
+        } else {
+          buffer.push(c)
+        };
       }
     };
     SKCSV.replayDiff(context, myReadLine, user, source)
--- a/sql/src/SqlCsv.sk
+++ b/sql/src/SqlCsv.sk
@@ -670,10 +670,6 @@ mutable class DPBuildingTableDiff(
   currentDiff: mutable TableDiff,
 ) extends DiffParsingState
 
-mutable class DPRejectingTxn(
-  writer: mutable Debug.BufferedWriter,
-) extends DiffParsingState
-
 fun nack(writer: mutable Debug.BufferedWriter, diff: readonly TableDiff): void {
   writer.write("^" + SKDB.feedbackTable(diff.table.name.origName) + "\n");
   for (line in diff.lines) {
@@ -725,6 +721,7 @@ fun replayDiff(
         tableDiffs,
         currentDiff,
       ) if (line.startsWith(":")) ->
+        currentDiff.lines.push(line);
         tableDiffs.push(currentDiff);
         checkpoint = parseCheckpoint(line);
 
@@ -766,7 +763,8 @@ fun replayDiff(
           for (tdiff in tableDiffs) {
             nack(writer, tdiff)
           };
-          mutable DPRejectingTxn(writer)
+          writer.flush();
+          break SKStore.CContinue(None())
         } else {
           break SKStore.CContinue(
             buildDiffChain(tableDiffs, identity, userFileOpt, checkpoint),
@@ -781,22 +779,6 @@ fun replayDiff(
         currentDiff.lines.push(line); // capture the raw input for if we end up rejecting the txn
         newRows = parseTableChange(context, currentDiff.table, line);
         currentDiff.rows.extend(newRows);
-        current
-
-      | current @ DPRejectingTxn(writer) if (line.startsWith("^")) ->
-        (table, snapshot) = parseHeader(context, line);
-        nack(writer, TableDiff::create(table, snapshot));
-        current
-
-      | DPRejectingTxn(writer) if (line.startsWith(":")) ->
-        writer.write(line);
-        writer.write("\n");
-        writer.flush();
-        break SKStore.CContinue(None())
-
-      | current @ DPRejectingTxn(writer) ->
-        writer.write(line);
-        writer.write("\n");
         current
       };
     }
--- a/sql/src/SqlCsv.sk
+++ b/sql/src/SqlCsv.sk
@@ -631,23 +631,25 @@ fun parseTableChange(
   accessSolverOpt match {
   | None() -> Some(newRows)
   | Some((accessSolver, tableName, authorCol, accessCol)) ->
-    for (row in newRows) {
-      SKDB.checkUserCanWriteRow(
-        context,
-        accessSolver,
-        tableName,
-        authorCol,
-        accessCol,
-        row,
-      ) match {
-      | SKDB.AROK() -> void
-      | SKDB.ARError(err) ->
-        print_error("Error: " + err.toString());
-        break None()
+    SKStore.withContext(ctx ->
+      for (row in newRows) {
+        SKDB.checkUserCanWriteRow(
+          ctx,
+          accessSolver,
+          tableName,
+          authorCol,
+          accessCol,
+          row,
+        ) match {
+        | SKDB.AROK() -> void
+        | SKDB.ARError(err) ->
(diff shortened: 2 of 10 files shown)


gregsexton:
I'll take a close look at the context usage.

gregsexton:
Ok, I took a look at the code and I think I understand what's happening.

write-csv is called from here: https://github.com/SkipLabs/skdb/blob/main/sql/src/Main.sk#L708

We run in a loop. Essentially: grab the context, read all the input, build the rows, and then return a function that is applied under lock that actually updates stuff with the rows we built.

The problem is the 'read all the input'. Once we enter the loop, we have the latest context, and then immediately block waiting on input. Which may arrive any time or never. But we don't update the context once we receive input.

This make sense?

So, I think we need to fix this rather than go with the solution you propose. For a few reasons:

1. Your solution will grab the latest context on each row check. This might be expensive. Not sure. But it will definitely allow privacy to change mid way through applying a transaction, so we can end up applying a partial transaction.
2. We hack a fix here, but it's an edge just waiting to trip someone up again.

I think a fix should be fairly straightforward but will probably require not using the looping abstraction `runSql`. We'd need to buffer a line, then begin the loop which uses this and then consumes the rest of the stream.

Or -- less ideal but should cover things -- we could rename the context passed in to replayDiff to make it clear it's potentially old. Then we could do a privacy check filter pass on all the rows after they've been accumulated but before we `buildDiffChain`. This would work with the latest context which only needs to be grabbed once.

WDYT? These were just the first couple of ideas that came to mind, open to others.

gregsexton:
(I'll look over the rest of this btw, but first concentrated on write-csv.)

bennostein:
> WDYT? These were just the first couple of ideas that came to mind, open to others.

Ahh, I see what you mean!  Was wondering what the runtime cost of grabbing the global context would be -- I don't think it's quite as bad as you say here (per-table, not per-row) but I agree on the reasoning that updating the context properly in execWriteCsv is a better solution going forward.

gregsexton:
> I don't think it's quite as bad as you say here (per-table, not per-row)

parseTableChange is called for every row being updated, one line at a time. Within that we call compute insert which can return multiple rows for the change (but I don't think it ever does?) so there's a loop here too. Between the function name and the loop within, not surprised there's confusion here. :D

bennostein:
Okay, this should be ready for review now.

I played around with a few options and ended up _both_ blocking on the first line before grabbing a context _and_ taking the freshest privacy context possible to check an entire transaction at once.

Just blocking on the first line does help but is insufficient to ensure that you have the latest privacy context.  The `withContext` in replayDiff does that, and allows to simplify some logic in the diff-parsing state machine there.

gregsexton:
(Looking at this today, haven't forgotten.)

sql/ts/src/skdb_types.ts
```diff
@@ -163,3 +165,21 @@ export interface SKDBShared extends Shared {
...
+
+  addAdmin: (skdb: SKDB, userID: string) => Promise<void>;
+  removeAdmin: (skdb: SKDB, userID: string) => Promise<void>;
+
+  addOwner: (skdb: SKDB, userID: string) => Promise<void>;
+  removeOwner: (skdb: SKDB, userID: string) => Promise<void>;
+  transferOwnership: (skdb: SKDB, userID: string) => Promise<void>;
+
+  removeMember: (skdb: SKDB, userID: string) => Promise<void>;
+}
```
gregsexton: Given you create one of these from an SKDB object, can we keep that link and remove the skdb param from every method? I think the simpler interface would be worth it.

sql/ts/src/skdb_types.ts
```diff
@@ -163,3 +165,21 @@ export interface SKDBShared extends Shared {
   create: (dbName?: string, asWorker?: boolean) => Promise<SKDB>;
   createSync: (dbName?: string, asWorker?: boolean) => Promise<SKDBSync>;
 }
+
+export interface SKDBGroup {
+  setDefaultPermission: (skdb: SKDB, perm: string) => Promise<void>;
```
gregsexton: Can we make perm something nicer than a string? I haven't worked with permissions for a few weeks and I already am not sure what to put here.

[edit]: looked at the tests and the permissions strings make sense. Given there are few of them, what about an enum? Would make parsing them trivial too.

sql/ts/src/skdb_types.ts
```diff
@@ -163,3 +165,21 @@ export interface SKDBShared extends Shared {
   create: (dbName?: string, asWorker?: boolean) => Promise<SKDB>;
   createSync: (dbName?: string, asWorker?: boolean) => Promise<SKDBSync>;
 }
+
+export interface SKDBGroup {
```
gregsexton, line 171: I'm not a fan of the word 'group'. It makes me think of a membership list/set. But we couple that with permissions per entry. I've seen this cause confusion a few times. I think I prefer 'PrivacyContext' - open to other ideas - as this conveys that you need one of these each time you have a situation that will have different privacy rules. Would you be open to working on the naming of this stuff? Am I making any sense?

sql/ts/src/skdb_types.ts
```diff
@@ -163,3 +165,21 @@ export interface SKDBShared extends Shared {
...
+  setMemberPermission: (
+    skdb: SKDB,
+    userID: string,
+    perm: string,
+  ) => Promise<void>;
+
+  addAdmin: (skdb: SKDB, userID: string) => Promise<void>;
+  removeAdmin: (skdb: SKDB, userID: string) => Promise<void>;
+
+  addOwner: (skdb: SKDB, userID: string) => Promise<void>;
```
gregsexton: Can you have multiple owners?

sql/ts/src/skdb_types.ts
```diff
@@ -87,6 +87,8 @@ export interface SKDB {
   ) => Promise<void>;
   connectedRemote: () => Promise<RemoteSKDB | undefined>;
   closeConnection: () => Promise<void>;
+
+  createGroup: () => Promise<SKDBGroup>;
```
gregsexton, line 91: How do you get an existing group to make modification to? I see in the tests that you create a group and then work with this from two skdb contexts, but it'll be more common for the non-creator-skdb to need to get a handle on the created group. The tests should probably do this?

sql/ts/src/skdb_group.ts
```diff
@@ -0,0 +1,129 @@
...
+    );
+    await skdb.exec(
+      "UPDATE skdb_groups SET skdb_access=@adminGroupID WHERE groupID=@adminGroupID;" +
+        "UPDATE skdb_group_permissions SET skdb_access=@adminGroupID WHERE groupID=@adminGroupID;",
+      { adminGroupID },
+    );
+    await skdb.exec(
+      "UPDATE skdb_groups SET skdb_access=@ownerGroupID, adminID=@ownerGroupID WHERE groupID=@ownerGroupID;",
+      { ownerGroupID },
+    );
```
gregsexton, line 100: Can this all be one txn in case of partial failure?

sql/src/Main.sk
```diff
@@ -705,14 +705,34 @@ fun execWriteCsv(args: Cli.ParseResults, options: SKDB.Options): void {
...
-        if (c == '\n') return String::fromChars(buffer.toArray());
-        buffer.push(c);
+        if (c == '\n') {
+          if (buffer.size() > 0) return String::fromChars(buffer.toArray());
+        } else {
+          buffer.push(c)
+        };
       }
     };
     SKCSV.replayDiff(context, myReadLine, user, source)
```
gregsexton: This runs in a loop so I don't think it makes sense to block waiting on first line before it. That will only handle the first case.

gregsexton: (It also makes this complex and I got a big memory win out of moving the readline stuff to a native impl.)

gregsexton: Now that I see the change made to the write-csv state machine loop, let's just kill this? I don't think it's needed?

sql/src/SqlCsv.sk
```diff
@@ -728,75 +688,95 @@ fun replayDiff(
...
         current
 
-      | DPRejectingTxn(writer) if (line.startsWith(":")) ->
-        writer.write(line);
-        writer.write("\n");
-        writer.flush();
-        break SKStore.CContinue(None())
-
-      | current @ DPRejectingTxn(writer) ->
-        writer.write(line);
```
gregsexton, line 798: [no action] It's a shame we lost the rejecting txn state as this saved buffering the entire input. But I can live with it. We should probably protect ourselves from buffering too much here regardless.

sql/src/SqlCsv.sk
```diff
@@ -728,75 +688,95 @@ fun replayDiff(
...
 
-      | current @ DPRejectingTxn(writer) if (line.startsWith("^")) ->
-        (table, snapshot) = parseHeader(context, line);
-        nack(writer, TableDiff::create(table, snapshot));
+      | current @ Some(DiffParsingState(_, currentDiff)) if (line == "\t\t") ->
+        currentDiff.!reset = true;
         current
 
-      | DPRejectingTxn(writer) if (line.startsWith(":")) ->
-        writer.write(line);
```
gregsexton, line 792: I think we've lost this? The checkpoint is required for any nack'd lines to be applied and stored. So I think - might be missing it - that we're rejecting and sending lines back to the client, but they'll never be stored in the rejected changes table because the checkpoint is missing to apply them. Can you test this scenario?

gregsexton (commented):
This is great! Will be much easier to use than crafting the sql.

Because this makes changes to the core loop of replication, which is always hairy, can you ensure that the simulator and soak run well?

make test-replication # should take about an hour
make test-soak #leave it over night

sql/ts/src/skdb_types.ts
bennostein: Yep! This was the design that Julien had specced out, modelling after e.g. GitHub where you have two tiers of "admin-ness".  We could simplify things a bit by maintaining exactly one owner, but this seems more a more general/expressive setup.

sql/src/SqlCsv.sk
bennostein: It's just moved! To line 722.

bennostein: I have tested this scenario -- rejection/feedback works as before, and it's covered by `test_replication_privacy_rejection_and_feedback`

sql/ts/src/skdb_group.ts
bennostein: I can group the inserts into one transaction, but the updates need to be separate for the writes to be accepted -- otherwise, the `skdb_access` on some of these rows is a group that, from the server's perspective, doesn't exist yet.

sql/ts/src/skdb_types.ts
bennostein: You are making sense and I think I agree, but for a different reason.

To me, this is more of a "group" ( ~ a named collection of privileges) than a "privacy context" ( ~ a given user's view of the world).  But, I could see some confusion cropping up between the rows of `skdb_groups`, and this abstraction of a "group" that comprises multiple such rows.

I'll think on better names...

sql/src/SqlCsv.sk
gregsexton: 722 is `currentDiff.lines.push(line);`? Where is it actually written to stdout?

bennostein: It's nack'ed at line 762 if the transaction is rejected, and written to stdout during the call to `buildDiffChain` otherwise

sql/ts/src/skdb_types.ts
gregsexton: Sounds good. And obviously can be a follow up PR. Just starting the conversation.

sql/src/SqlCsv.sk
gregsexton: The nack just takes care of the table. We need to call that for each table and then finally write the checkpoint line. It's this last bit that's missing. It will catch up and do the right thing when something eventually succeeds which might be why a test is passing?

bennostein: (closing the loop here for future reference: the current setup is correct/fine; the `nack` takes care of the checkpoint line)

gregsexton:
Looks great!

PR #23, opened 2023-12-17
Add arparser package. 

beauby:
This package provides a simple parser for static library files. It
supports the 3 flavors in use on UNIX-like systems: Common, GNU, and
BSD.

This PR is based on #11.

### DIFF
--- a/arparser/src/parser.sk
+++ b/arparser/src/parser.sk
@@ -1,0 +1,621 @@
+module ARParser;
+
+const kGLOBAL_HEADER: String = "!<arch>\n";
+const kGLOBAL_HEADER_LEN: Int = kGLOBAL_HEADER.length();
+const kENTRY_HEADER_LEN: Int = 60;
+const kENTRY_IDENTIFIER_OFFSET: Int = 0;
+const kENTRY_IDENTIFIER_LEN: Int = 16;
+const kENTRY_MTIME_OFFSET: Int = 16;
+const kENTRY_MTIME_LEN: Int = 12;
+const kENTRY_UID_OFFSET: Int = 28;
+const kENTRY_UID_LEN: Int = 6;
+const kENTRY_GID_OFFSET: Int = 34;
+const kENTRY_GID_LEN: Int = 6;
+const kENTRY_MODE_OFFSET: Int = 40;
+const kENTRY_MODE_LEN: Int = 8;
+const kENTRY_SIZE_OFFSET: Int = 48;
+const kENTRY_SIZE_LEN: Int = 10;
+const kENTRY_HEADER_TERMINATOR_OFFSET: Int = 58;
+const kENTRY_HEADER_TERMINATOR_LEN: Int = 2;
+const kENTRY_HEADER_TERMINATOR: String = "`\n";
+
+const kBSD_SYMBOL_LOOKUP_TABLE_ID: String = "__.SYMDEF";
+const kBSD_SORTED_SYMBOL_LOOKUP_TABLE_ID: String = "__.SYMDEF SORTED";
+const kBSD_EXTENDED_IDENTIFIER_PREFIX: String = "#1/";
+
+const kGNU_SYMBOL_LOOKUP_TABLE_ID: String = "/";
+const kGNU_NAME_TABLE_ID: String = "//";
+
+// An ar archive starts with the magic string "!<arch>\n", followed by a
+// sequence of header/data pairs.
+//
+// Each header is 60 bytes, divided as follows:
+// +------------+----------+---------+---------+---------+----------+------------+
+// | identifier |  mtime   |   uid   |   gid   |  mode   |   size   | end marker |
+// |  16 bytes  | 12 bytes | 6 bytes | 6 bytes | 8 bytes | 10 bytes |  2 bytes   |
+// |   ASCII    | decimal  | decimal | decimal |  octal  | decimal  |   ASCII    |
+// +------------+----------+---------+---------+---------+----------+------------+
+//
+// The decimal/octal numbers are written out in ASCII, padded (on the right)
+// with spaces (0x20).
+//
+// Each header starts at an even offset, and records with an odd size are padded
+// with a single newline (0x0a) byte.
+//
+// The `size` field represents the size of the data itself (excluding the header
+// and any possible padding).
+//
+// Since identifiers are at most 16 bytes, longer file names need a different
+// representation. The format is not standardized, but there are mainly two
+// flavors around: GNU/SysV and BSD.
+//
+// In the BSD flavor, file names longer than 16 bytes, or containing a space,
+// are stored directly after the header. The identifier is set to the string
+// "#1/" followed by the length of the file name written in decimal, and the
+// size is set to the decimal representation of the combined size of the data
+// and the file name.
+// File names up to 16 bytes and containing no space are stored directly in the
+// identifier, padded with spaces (0x20).
+//
+// In the GNU/SysV flavor, a special file named "//" is added to the archive
+// before any regular file. File names longer than 15 bytes are stored within
+// the "//" file. The identifier is set to the string "/" followed by the offset
+// in "//" at which the file name is stored, in decimal, followed by the string
+// "/\n".
+// File names up to 15 bytes (possibly containing spaces) are stored directly in
+// the identifier, followed by a "/".
+//
+// An optional symbol table can be present as a special file (named "/" in the
+// GNU/SysV flavor and "__.SYMDEF" in the BSD flavor). If present, it must be
+// the first member in the archive.
+
+private class Header{
+  identifier: String,
+  size: Int,
+  mtime: Int = 0,
+  uid: Int = 0,
+  gid: Int = 0,
+  mode: Int = 0o600,
+} {
+  static fun read<R: mutable IO.Read>(reader: R): Result<?Header, Error> {
+    buf = Array::mfill(kENTRY_HEADER_LEN, UInt8::truncate(0));
+
+    reader.read(buf.mbytes()) match {
+    | Failure(err) -> return Failure(IOError(err))
+    | Success(nbytes) if (nbytes == 0) -> return Success(None())
+    | Success(nbytes) if (nbytes < kENTRY_HEADER_LEN) ->
+      reader.read_exact(buf.mbytes().mslice(nbytes)) match {
+      | Failure(err) -> return Failure(IOError(err))
+      | Success _ -> void
+      }
+    | Success _ -> void
+    };
+
+    identifier_raw = Vector::mcreateFromItems(
+      buf.slice(
+        kENTRY_IDENTIFIER_OFFSET,
+        kENTRY_IDENTIFIER_OFFSET + kENTRY_IDENTIFIER_LEN,
+      ),
+    );
+
+    while (identifier_raw.maybeLast() == Some(UInt8::truncate(0x20))) {
+      _ = identifier_raw.pop()
+    };
+
+    // TODO: Explicit ASCII check.
+    identifier = String::fromUtf8(identifier_raw.collect(Array));
+
+    size = parse_number(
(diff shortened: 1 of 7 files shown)


arparser/src/parser.sk
```diff
@@ -0,0 +1,372 @@
+module ARParser;
+
```
jberdine, line 2: Are there any references to the format specifications that could be added in comments here?

beauby: Sadly there is no such thing as a standardized spec for this format. The FreeBSD man page is the best resource I found: https://man.freebsd.org/cgi/man.cgi?query=ar&sektion=5.

beauby: I added a description of the format.

prelude/src/core/primitives/Bytes.sk
```diff
@@ -72,4 +72,55 @@ private mutable class BytesIterator(
...
+
+    Success(readSz)
+  }
+}
+
+mutable class MutableCursor(
+  buffer: mutable MutableBytes,
+  mutable pos: Int = 0,
+) uses IO.Write, IO.Read {
+  static fun create(buffer: mutable MutableBytes): mutable MutableCursor {
```
mbouaziz, line 114: I haven't played enough with Skip yet, I'm wondering if you could write `Cursor` as a generic `Cursor<T : Bytes>` and have `write` defined optionally: `write[T : mutable MutableBytes]` in order to reduce code duplication.

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+//
+// In the BSD flavor, file names longer than 16 bytes, or containing a space,
+// are stored directly after the header. The identifier is set to the string
+// "#1/" followed by the length of the file name written in decimal, and the
+// size is set to the decimal representation of the combined size of the data
+// and the file name.
+// File names up to 16 bytes and containing no space are stored directly in the
+// identifier, padded with spaces (0x20).
+//
+// In the GNU/SysV flavor, a special file name "//" is added to the archive
```
mbouaziz: named

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+
+const kGNU_SYMBOL_LOOKUP_TABLE_ID: String = "/";
+const kGNU_NAME_TABLE_ID: String = "//";
+
+// An ar archive starts with the magic string "!<arch>\n", followed by a
+// sequence of header/data pairs.
+//
+// Each header is 60 bytes, divided as follows:
+// +------------+----------+---------+---------+---------+----------+------------+
+// | identifier |  mtime   |   uid   |   gid   |  mode   |   size   | end marker |
```
mbouaziz, line 34: You describe `identifier` as being either a filename or something that allows recovering the filename.
`mtime`, `uid`, `gid`, `mode` seems to be what I expect.
`size` too. Let's clarify that it's the size of the data itself, not counting the header (but possibly the filename in the BSD flavor), and whether it counts the possible extra 0x0a.
What is the `end marker`? Edit: after reading the code, seems to be the constant `kENTRY_HEADER_TERMINATOR`

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+
+    while (identifier_raw.maybeLast() == Some(UInt8::truncate(0x20))) {
+      _ = identifier_raw.pop()
+    };
+
+    // TODO: Explicit ASCII check.
+    identifier = String::fromUtf8(identifier_raw.collect(Array));
+
+    size = parse_number(
+      buf.slice(kENTRY_SIZE_OFFSET, kENTRY_SIZE_OFFSET + kENTRY_SIZE_LEN),
```
mbouaziz, line 109: `slice` is begin + end, what if you added another method for begin + length?

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+      data = this.get(header.identifier) match {
+      | Success(Some(data)) -> data
+      | _ -> break void
+      };
+      yield (header, data)
+    }
+  }
+
+  mutable fun scan(): Result<void, Error> {
+    if (this.scanned) {
```
mbouaziz: `scan` is called by the constructor, is it meant to be called again? If no -> `private`.
Then is `scanned` still useful?

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+    | VariantGNU() -> identifier == kGNU_SYMBOL_LOOKUP_TABLE_ID
+    }
+  }
+
+  private mutable fun parse_global_header(): Result<void, Error> {
+    this.parse_fixed_tag(kGLOBAL_HEADER) match {
+    | Failure(err) -> return Failure(err)
+    | Success _ -> void
+    };
+    this.!started = true;
```
mbouaziz: Similarly, I think `started` is useless because you start as you construct an `Archive`. Probably a remain from a past lazy parsing?
Also, you never use its value

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+  mutable fun scan(): Result<void, Error> {
+    if (this.scanned) {
+      return Success(void)
+    };
+    this.parse_global_header() match {
+    | Failure(err) -> return Failure(err)
+    | Success _ -> void
+    };
+    loop {
+      header_start = this.next_entry_start;
```
mbouaziz: `next_entry_start` could just be a local, isn't it?

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+        header.identifier.startsWith("/")
+      ) {
+        this.!variant = VariantGNU();
+        if (header.identifier == kGNU_NAME_TABLE_ID) {
+          this.name_table.resize(header.size, UInt8::truncate(0));
+          this.reader.read_exact(this.name_table.mbytes()) match {
+          | Failure(err) -> return Failure(IOError(err))
+          | Success _ -> void
+          }
+        } else if (header.identifier != kGNU_SYMBOL_LOOKUP_TABLE_ID) {
```
mbouaziz: You didn't describe what just `/` means

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+    Success(void)
+  }
+
+  private mutable fun parse_gnu_extended_identifier(
+    header: Header,
+  ): Result<String, Error> {
+    start = parse_number(
+      header.identifier
+        .bytes()
+        .slice(kENTRY_IDENTIFIER_OFFSET + 1, kENTRY_IDENTIFIER_LEN)
```
mbouaziz: `kENTRY_IDENTIFIER_OFFSET + 1` -> `1` because you're in the bytes of `identifier`, not `header`

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+      };
+
+      header_len = kENTRY_HEADER_LEN;
+      if (
+        !(this.variant is VariantBSD _) &&
+        header.identifier.startsWith("/")
+      ) {
+        this.!variant = VariantGNU();
+        if (header.identifier == kGNU_NAME_TABLE_ID) {
+          this.name_table.resize(header.size, UInt8::truncate(0));
```
mbouaziz, line 284: Shall we check this is the first entry?

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+          header_extra_len,
+        ) = this.parse_bsd_extended_identifier(header) match {
+        | Failure(err) -> return Failure(err)
+        | Success(v) -> v
+        };
+        !header.identifier = extended_identifier;
+        !header.size = header.size - header_extra_len;
+        !header_len = header_len + header_extra_len
+      };
+
```
mbouaziz, line 327: IIUC, we know from the very first file whether we're in the GNU (`//` or name ending with `/`) or BSD variant (anything else), right?

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+      this.!next_entry_start =
+        this.next_entry_start +
+        header_len +
+        header.size +
+        (header.size % 2);
```
mbouaziz: I think that you should compute that before the big `if`.

This version is probably wrong in the very special case of BSD-extended identifiers when the `header_extra_len` is odd. Isn't it?

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+
+      if (this.is_name_table_id(header.identifier)) {
+        continue
+      };
```
mbouaziz: Then you can move this `continue` directly in the correct branch above and get rid of `is_name_table_id`

arparser/src/parser.sk
```diff
@@ -0,0 +1,614 @@
...
+    this.variant is VariantGNU _ && identifier == kGNU_NAME_TABLE_ID
+  }
+
+  private readonly fun is_symbol_lookup_table_id(identifier: String): Bool {
+    this.variant match {
+    | VariantCommon() -> false
+    | VariantBSD() ->
+      identifier == kBSD_SYMBOL_LOOKUP_TABLE_ID ||
+        identifier == kBSD_SORTED_SYMBOL_LOOKUP_TABLE_ID
+    | VariantGNU() -> identifier == kGNU_SYMBOL_LOOKUP_TABLE_ID
```
mbouaziz, line 407: Could you describe them in the header comment?

arparser/src/parser.sk
beauby: This would not work as `next_entry_start` still needs to be properly updated.

arparser/src/parser.sk
beauby: Not really because BSD allows file names to contain `/`.

arparser/src/parser.sk
beauby: Hmm that seems right.

arparser/src/parser.sk
beauby: My bad, I read your comments out of order!

mbouaziz: _Then_ referred to the previous comment being taken into account

arparser/src/parser.sk
beauby: Which means we're too eager setting the variant to GNU.

mbouaziz: Can a BSD filename **end** with `/`?

beauby: According to the [FreeBSD man page for `ar (5)`](https://man.freebsd.org/cgi/man.cgi?query=ar&sektion=5), it does not seem to be forbidden. I'll re-check FreeBSD's `ar` implementation to be sure.

arparser/src/parser.sk
beauby: Correct. It happened to work because `kENTRY_IDENTIFIER_OFFSET` is actually `0`.

arparser/src/parser.sk
beauby: Do you mean adding a helper function within `ARParser` or adding a method to the `Array` API?

mbouaziz: `Array`

prelude/src/core/primitives/Bytes.sk
beauby: I remember trying to avoid the `Foo/MutableFoo` thing (for both `Cursor` and `Bytes`), but without success. If you have a cleaner API that works, feel free to nuke `MutableCursor`/`MutableBytes`!

arparser/src/parser.sk
beauby: I agree on the principle, let's discuss it in an issue.

arparser/src/parser.sk
beauby: Looking at [binutils](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/archive.c;h=9f3fbce9bc963fd72292451f85113cf33a8d002d;hb=HEAD#l591), they use the presence of a terminating `/` as an indication of `GNU/SysV` format. Whether BSD archives can _technically_ have a `/` at the end is an other question, but it seems a fringe enough use case that we should not worry about it.

mbouaziz (approved):
Go!

PR #24, opened 2023-12-17
[skip] Require `module end;` statement at end of file.

beauby:
Until this commit, the closing `module end;` statement could be
omitted at the end of a file.
The benefits were tenuous (and can be obtained using a linter with
autofix), and came at the cost of not being able to concatenate skip
files.


jberdine (approved):
LGTM.

PR #25, opened 2023-12-17
[skc] Merge `Main.Config` and `Config.Config`.

beauby:
There is no clear criterion for deciding what goes where, so let's just merge them.


jberdine (approved):
LGTM

PR #27, opened 2023-12-18
Remove redundant error checking

jberdine:
sk_malloc already checks its success and throws on failure

### DIFF
--- a/compiler/runtime/obstack.c
+++ b/compiler/runtime/obstack.c
@@ -82,12 +82,6 @@ char* sk_large_page(size_t size) {
   size_t block_size = size + sizeof(char*) + sizeof(size_t) + sizeof(sk_saved_obstack_t);
   block_size += 64;
   char* lpage = (char*)sk_malloc(block_size);
-  if (lpage == NULL) {
-#ifdef SKIP64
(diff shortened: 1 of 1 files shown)


compiler/runtime/obstack.c
```diff
@@ -82,12 +82,6 @@ char* sk_large_page(size_t size) {
   size_t block_size = size + sizeof(char*) + sizeof(size_t) + sizeof(sk_saved_obstack_t);
   block_size += 64;
   char* lpage = (char*)sk_malloc(block_size);
-  if (lpage == NULL) {
-#ifdef SKIP64
-    fprintf(stderr, "Out of memory\n");
-#endif
-    SKIP_throw_cruntime(ERROR_OUT_OF_MEMORY);
```
gregsexton, line 89: sk_malloc under 32 bit doesn't do this. But then it doesn't really handle this well at all and doesn't return null.

jberdine: Yes true, in 32 bit it never returns NULL, just keeps bumping, I guess just assuming that the JS VM will kill it if it goes too far.

beauby (approved):
LGTM

PR #34, opened 2024-01-04
Clean up test output

bennostein:


gregsexton:
Do the replication tests print anything now on success? I'm not sure how robust the orchestration is to exit codes. Can we distinguish all passed from none ran? Other than this, lgtm.

bennostein:
Yes, there's still output on success -- it's just more nicely formatted/aligned and easier to read now :)

PR #35, opened 2024-01-04
Make SkipLabs logo SVG compatible with React JSX

bennostein:


gregsexton:
Out of curiosity, what changed?

bennostein:
I stripped out all of the XML namespaces (i.e.  keys of the form `foo:bar` became `fooBar`)

PR #37, opened 2024-01-06
Share more data on INSERT

pikatchu:
Up until now, we were recomputing the order of the rows for every single row in a table. It is correct, but it was incredibly wasteful. The order is always going to be the same for every single entry of the table, we should make them all share the same pointer instead of recomputing it every time.

This is a 20% win in memory for a table 100000 integer entries.


gregsexton:
LGTM

PR #38, opened 2024-01-09
Use 'TEXT' instead of 'STRING' type in SQL

bennostein:


rfc/005-foreign-key-based-privacy-model.org
```diff
@@ -303,7 +303,7 @@ rows in ~skdb_access~:
 | from_table | to_table | from_col | id         | permission | user_group                   |
 |------------+----------+----------+------------+------------+------------------------------|
 | auctions   | users    | creator  | $A_USER_ID | read       | GROUP_ID_FOR_EVERYONE        |
-| auctions   | users    | creator  | $A_USER_ID | insert     | GROUP_ID_FOR_JUST_$A_USER_ID |
+| auctions   | users    | creator  | $A_USER_ID | insert     | GROUP_ID_FOR_JUST_$A_USER_ID  |
```
jberdine: ?

rfc/005-foreign-key-based-privacy-model.org
```diff
@@ -53,11 +53,11 @@ We have a table ~skdb_access~:
...
-    to_table STRING NOT NULL,     --e.g. users
-    from_col STRING NOT NULL,     --e.g. authorID
-    id STRING NOT NULL,           --e.g. 1234
-    permission STRING NOT NULL,   --read, insert, or delete
+    from_table TEXT NOT NULL,   --e.g. posts
+    to_table TEXT NOT NULL,     --e.g. users
+    from_col TEXT NOT NULL,     --e.g. authorID
+    id TEXT NOT NULL,           --e.g. 1234
+    permission TEXT NOT NULL,   --read, insert, or delete
     user_group INTEGER NOT NULL   --e.g. 4567 - encoded elsewhere, represents a list of users
```
jberdine: nit: this doesn't preserve vertical alignment of the comments

jberdine (approved):

sqlparser/src/Token.sk
```diff
@@ -242,7 +242,7 @@ base class Token uses Equality, Show, Hashable {
   | TInteger _ -> "INTEGER"
   | THexInteger _ -> "INTEGER"
   | TFloat _ -> "FLOAT"
-  | TSingleQuotedString _ -> "STRING"
```
beauby: This is not required and probably not what we want, as "STRING" here refers to the token type of a single quoted string in a SQL statement (such as `SELECT 'hello'`).

beauby: However we probably want to get rid of [sqlparser/src/Parser.sk#L100](https://github.com/SkipLabs/skdb/blob/dd193e8e34e5cf59fb4eafd36333c450a22bb6ee/sqlparser/src/Parser.sk#L100):
```
    if (names.any(s -> s.lower.contains("int"))) {
      INTEGER()
    } else if (
      names.any(s ->
        s.lower.contains("char") ||
          s.lower.contains("clob") ||
          s.lower.contains("text") ||
-         s.lower.contains("string") // NOTE: This differs from SQLite.
      )
    ) {
      TEXT()
```

(which also makes me realize there is a typo two lines above, where `clob` should be `blob`).

bennostein: Ahh, I was wondering why I couldn't find lexing code for a keyword "STRING"!

Will fix both things.

beauby: Actually, the `"clob"` thing was correct, my bad (cf. [SQLite's algorithm](https://www.sqlite.org/datatype3.html)).

PR #40, opened 2024-01-11
Change s/STRING/TEXT/ in newly added test from write-csv GC PR

bennostein:
Extend #38 to #18, which both test green against master but fail when combined.


gregsexton:
LGTM. Sorry for the churn.

PR #41, opened 2024-01-11
[skdb] Add support for CURRENT_{DATE,TIME,TIMESTAMP}.

beauby:


sql/src/SqlEval.sk
```diff
@@ -332,6 +334,21 @@ fun valueToCValue(
       }
     | P.VId(None()) -> Some(CString(Ksuid::create().toString()))
     | P.VString(str) -> Some(CString(str))
+    | P.VCurrentTime() ->
+      if (inTransaction) {
+        error(pos, "CURRENT_TIME cannot be used in a transaction")
+      };
+      Some(CString(strftime(context, "%H:%M:%S", "now", Array[])))
```
gregsexton, line 341: Will this be UTC or local?

gregsexton (commented):
LGTM

sql/src/SqlEval.sk
beauby: According to [SQLite](https://www.sqlite.org/lang_datefunc.html#time_values), `'now'` is supposed to be UTC, though I'm not sure what's currently happening [here](https://github.com/SkipLabs/skdb/blob/main/sql/src/SqlDate.sk#L211). @skiplabsdaniel Can you confirm?

skiplabsdaniel: Indeed, here the default timezone is the local one. In most databases, the default timezone is based on the database server's time zone setting. We can make a special case to do like SQLite, but SQLite don't have real timezone management, i'm not sure it's a good choice. I prefer to be consistent and have same timezone for everything, it could be UTC, even if I prefer local one.

gregsexton: UTC by default will prevent some nasty surprises in prod, particularly if/when we run servers in different AWS regions.

beauby: I would personally lean towards UTC as well for consistency, though it might be ok to have `now` be in the local tz, as long as we change the format to include a tz indicator. Thoughts?

ltique: I don't think we should overthink this: all UTC all the time is never a bad idea, and certainly good enough for now IMHO.

ltique:
@beauby I have done some limited testing and this works well! I think we should merge this soon (perhaps after the minor timezone issue above has been fixed)?

PR #42, opened 2024-01-11
Add client-specified schemas to mirroring/tailling protocol

bennostein:


sql/server/core/src/main/kotlin/core/Orchestration.kt
```diff
@@ -21,6 +21,7 @@ data class ProtoQuery(val query: String, val format: QueryResponseFormat) : Prot
 
 data class ProtoRequestTail(
     val table: String,
+    val schema: String,
```
gregsexton: Can we call this expectedSchema? Makes it clearer that this is a client driven thing.

sql/server/core/src/main/kotlin/core/Orchestration.kt
```diff
@@ -92,12 +93,16 @@ fun decodeProtoMsg(data: ByteBuffer): ProtoMessage {
...
+      val schemaLength = data.getShort()
+      val schemaBytes = ByteArray(schemaLength.toInt())
+      data.get(schemaBytes)
       val filterExprLength = data.getShort()
       val filterBytes = ByteArray(filterExprLength.toInt())
       data.get(filterBytes)
       val params = decodeParams(data)
       ProtoRequestTail(
           String(tableNameBytes, StandardCharsets.UTF_8),
+	  String(schemaBytes, StandardCharsets.UTF_8),
```
gregsexton: [nit] tab indent

rfc/003-protocol.org
```diff
@@ -311,6 +311,8 @@ Where not specified, strings are UTF-8.
   - since (64)
   - table_name_length (16) - in bytes
   - table_name (table_name_length * 8)
+  - schema_length (16) - in bytes
+  - schema (schema_length * 8) - JSON-encoded expectation of table schema
```
gregsexton, line 315: This is a backwards breaking change. I think this is actually what we want to do at this stage, so we should bump the minimum supported client version: https://github.com/SkipLabs/cloud/blob/main/skgw/src/main/kotlin/App.kt#L476

For future though, if you wanted to maintain backwards compat, we could add this at the end and make finding and decoding it optional. That only works once though, so probably better would be something like adding a flag that the client sets in the reserved space or a version number perhaps.

sql/server/core/src/main/kotlin/core/Orchestration.kt
```diff
@@ -223,14 +228,16 @@ fun encodeProtoMsg(msg: ProtoMessage): ByteBuffer {
       buf.flip()
     }
     is ProtoRequestTail -> {
-      val buf = ByteBuffer.allocate(16 + msg.table.length * 4 + (msg.filterExpr?.length ?: 0) * 4)
+      val variableWidth = msg.table.length + msg.schema.length + (msg.filterExpr?.length ?: 0)
```
gregsexton: Ah, you can remove the conditional on filterExpr - this is a hang over and isn't optional any more.

sql/ts/src/skdb_types.ts
```diff
@@ -17,12 +17,11 @@ export interface SKDBHandle {
...
-  | {
-      table: string;
-      filterExpr?: string;
-    }
-  | string;
+export type MirrorDefn = {
+  table: string;
+  schema: string;
+  filterExpr?: string;
+};
```
gregsexton: This is great. Exactly what I hoped you would do. Can we call schema 'expectedSchema' though? Just to make the semantics clearer to users. Actually, even better would be something like 'expectedColumns' or 'expectedColDefns', something like this. Until I saw your usage in skdb_database.ts, I wasn't sure what a user should be providing.

sql/ts/src/skdb_database.ts
```diff
@@ -183,9 +183,21 @@ export class SKDBSyncImpl implements SKDBSync {
 
     // Setting up local privacy checks
     await this.mirror(
-      "skdb_user_permissions",
-      "skdb_groups",
-      "skdb_group_permissions",
+      {
+        table: "skdb_user_permissions",
+        schema:
+          "(userID TEXT PRIMARY KEY, permissions INTEGER NOT NULL, skdb_access TEXT NOT NULL)",
```
gregsexton, line 189: Are the parens required?

sql/server/skgw/src/main/kotlin/App.kt
```diff
@@ -169,6 +169,10 @@ class RequestHandler(
         stream.close()
       }
       is ProtoRequestTail -> {
+        if (!skdb.canMirror(request.table, request.schema)) {
+          stream.error(2003u, "Schema mismatch")
+          return this
+        }
```
gregsexton: Can you add this and any other changes under skgw/ to the cloud repo? I'm about to kill this dir.

sql/ts/src/skdb-cli.ts
```diff
@@ -490,7 +490,9 @@ const localRepl = async function () {
       console.log(".schema -- Output the schema.");
       console.log(".table-schema <table> -- Output the schema <table>.");
       console.log(".view-schema <view> -- Output the schema for <view>.");
-      console.log(".mirror <table1> <table2> ... -- Mirror the remote tables.");
+      console.log(
+        ".mirror <table1> <schema1> <table2> <schema2> ... -- Mirror the remote tables.",
+      );
```
gregsexton: This is going to be a pain in the ass when developing, testing, debugging. Can we do something where you can say "just give me the latest schema, it's cool, this isn't for application code". Maybe if you send a special string like "(*)"? I'm open to whatever but I think it'll be worth it for user driven tooling that doesn't have app logic written against a static schema.

gregsexton (commented):
Looks great!

Could do with a test perhaps in apitests.ts.

It would be really cool if we hook in to this info with the dev server so we don't have to duplicate the schema in the code and also removes dev-specific code. Currently we write something like:

```
  await remoteDb.schema(
    "CREATE TABLE tasks (id STRING PRIMARY KEY, name STRING, complete INTEGER, skdb_access STRING);",
  );
  ```

And on change the auto-migration kicks in. Maybe the dev server can, instead of checking and validating the schema, just use this to drive a migration. I guess this only works on mirroring though and you might want the schema setup before then.

File: sql/ts/src/skdb_database.ts (hunk 6 above)
Comment (bennostein) [reply sql/ts/src/skdb_database.ts:189]: Yes, though we could drop that requirement if we wanted -- it just makes parsing easy on the Skip side of things.

File: sql/ts/src/skdb_orchestration.ts [hunk 9]
```diff
@@ -168,7 +178,11 @@ function encodeProtoMsg(msg: ProtoMsg): ArrayBuffer {
       return buf.slice(0, paramsOffset + 2 + (encodeResult.written || 0));
     }
     case "tailBatch": {
-      const buffers = msg.requests.map((req) => encodeProtoMsg(req));
+      const buffers = msg.requests.map((req) => {
+        const td = new TextDecoder();
```
gregsexton: td isn't used?

gregsexton (commented):
This looks great! Thanks for addressing my comments. :)

sql/ts/src/skdb_orchestration.ts
bennostein: Ahh, good catch!  Was using that to debug.

gregsexton:
Ship it!

PR #44, opened 2024-01-12
Remove vestiges of support for 'memoized' functions

jberdine:
Skip no longer has builtin support for `memoized` functions, but the compiler has a bunch of left-over dead code for them. This PR removes this old support.


jberdine:
Based on https://github.com/SkipLabs/skdb/pull/43

compiler/runtime/runtime.h
```diff
@@ -167,10 +168,12 @@ typedef struct {
   void* unused1;
   void* unused2;
 #endif
-  void (*m_onStateChange)(void*, long);
+  void* unused;
```
jberdine, line 171: Changing this dead field to unused padding is not the change I wanted to make. This PR started from investigations in the runtime system memory layout code, the copy from obstacks, etc. where there was unclear to me alignment, and this odd field. So far my attempts to remove this field, and to update `createGCType` in vtable.sk correspondingly, have failed. Even though removing this field preserves the needed 64-bit alignment, removing this field (or adding another one of padding), results in a skc binary that segfaults early. I suspect that somewhere else in the compiler or runtime system knows about the size of these vtable entries, but I have not found it. Any pointers welcome!

jberdine: @pikatchu tagging you in case you have any suggestions of where to look for other parts of the code that might rely on the size of vtable entries.

jberdine:
@pikatchu Is there any thought to re-introduce `memoized` into the language? If not, even without changing the vtables this PR is a simplification and cleanup of the compiler. Any objection to merging this?

beauby (approved):

pikatchu:
> @pikatchu Is there any thought to re-introduce `memoized` into the language? If not, even without changing the vtables this PR is a simplification and cleanup of the compiler. Any objection to merging this?

No objection. I think the lazyDirectory approach in SKStore is cleaner, because it defines a scope for the things that are being cached. One can define different policies for eviction etc ...

So it's unlikely that we would want to bring memoization in its old form back.

PR #46, opened 2024-01-12
Factor SKGW out to separate repo

gregsexton:


gregsexton:
WIP diff. Tests won't pass. Just wanted to flag that this is happening so if you're modifying skgw we're briefly in danger of dropping changes. I'll try to make sure anything changed concurrently to this is backported.

PR #48, opened 2024-01-12
Remove dead support for disabling GC

jberdine:
The sk_gc_enabled flag is dead as it is never changed.


jberdine:
I don't know if we might want to keep this, but I noticed that it's dead so figured I'd send this PR to flag it.

jberdine:
I'll incorporate this into https://github.com/SkipLabs/skdb/pull/114.

PR #50, opened 2024-01-13
[skdb] Get rid of stale `fixUpDirName()`.

beauby:


jberdine:
The "Database already exists" error failing the CI doesn't make any sense, right?

gregsexton:
Try rebasing on what I just landed. I suspect this will go away forever now.

beauby:
I don't see how it could be related (it should simply fail to build if there was an issue with this PR). I triggered a re-run for good measure, but we should fix the flaky test.

Also, we should probably disable retries for wasm tests involving a server, as it invariably leads to stale state (hence the `Database already exists`).

PR #51, opened 2024-01-15
Fix: query params were not serialising null in the server

gregsexton:
This meant that logging inserts were failing. I think it also affected
config lookups.


jberdine (approved):
Ah, so easy to overlook...

gregsexton:
Yeah, I'm really not a fan of these clever serialisation libraries. You wouldn't believe how hard it is to find a json library in java that just lets you build and serialise a value.

PR #53, opened 2024-01-16
[skdb] Fix `LIKE` pattern matching.

### LINKED ISSUE #52: Pattern starting with % don't work
This doesn't work:

echo "select * from (select '' as path) where path like '%foo';" | skdb

But the following does:

echo "select * from (select '' as path) where path like '%foo%';" | skdb

beauby:
Closes #52

### DIFF
--- a/sql/src/SqlCAst.sk
+++ b/sql/src/SqlCAst.sk
@@ -6,7 +6,6 @@ module alias P = SQLParser;
 module SKDB;
 
 value class Pattern{
-  hasUnderscore: Bool,
   firstIsPercent: Bool,
   lastIsPercent: Bool,
   subs: Array<Array<Char>>,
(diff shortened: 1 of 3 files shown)


jberdine (approved):
LGTM

jberdine:
Note that while these changes look good, I don't understand how they fix the linked issue. Any explanation appreciated. :-)

beauby:
The actual fix is [here](https://github.com/SkipLabs/skdb/pull/53/commits/f7227e8aa10de2f3a64696a3247bc8faaba141f5#diff-cba1c5f6e6db930296ffb089687b5978ca9ed12605f32ebc9d393c7825dcd7ceR963). The previous code worked under the assumption that if there was no trailing `%` in the pattern, the LHS would be at least as long as the last non wildcard component of the pattern.

PR #55, opened 2024-01-16
[skdb] Make expression aliases available in `WHERE` clauses.

### LINKED ISSUE #54: Aliases unavailable in `WHERE` clauses.
Example: `SELECT 1 AS foo WHERE foo = 1`.

beauby:
Closes #54


jberdine (approved):
LGTM

beauby:
This approach won't work as is. I'll work on a proper solution.

PR #56, opened 2024-01-16
Catch exceptions on sending data/close

gregsexton:
On IO error we pessimistically assume that we're in a bad state and
fail the socket. This is the only practical option to ensure invariants.

We must catch here otherwise the exception bubbles up leaving the
socket in a bad state but the application doesn't get notified. This
was the root cause of running out of connections when we should not.

Why do we get IO exceptions? Because undertow is badly designed. If
the client exits quickly it closes the TCP socket before we've had a
chance to gracefully close the muxed socket, websocket, etc. This
_should_ be fine, it's half-closed and so we should be allowed to send
our closing messages even though the client is not reading them. But
undertow throws exceptions in this state even though there's nothing
exceptional about this.

### DIFF
--- a/sql/server/core/src/main/kotlin/core/Mux.kt
+++ b/sql/server/core/src/main/kotlin/core/Mux.kt
@@ -1035,11 +1035,19 @@ class MuxedSocket(
   }
 
   private fun sendClose(code: Int, reason: String) {
-    channel.sendClose(code, reason)
+    try {
+      channel.sendClose(code, reason)
+    } catch (ex: Exception) {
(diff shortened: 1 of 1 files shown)


gregsexton:
This takes care of the infamous:

```
Exception in thread "pool-3-thread-1" java.nio.channels.ClosedChannelException
```
or
```
java.io.IOException: UT002002: Channel is closed
```

This was a harmless race but turns out we weren't handling the exceptions well and this was messing up the socket accounting, wrongly running us out of available connections, not logging, etc.

bennostein:
Nice! LGTM

PR #58, opened 2024-01-16
Speed up skdate tests by parallelizing execution and storing generated inputs

bennostein:
Checking how this affects CircleCI times


bennostein:
Pending CircleCI numbers, but these changes speed up the skdate test suite from ~6 minutes to ~1 minute on my machine.

Parallelized where possible, and stored much more pre-generated input than the initial commit in this stack.

There's roughly ~7MB (compressed) of inputs checked in here.  I don't think that's big enough to cause issues being checked in, but can roll back that part of this diff and still get a decent win from the parallelization if so.

jberdine:
Nice!

gregsexton:
2.4 million lines changed. This is going to take a while to review, bear with me.

jberdine:
Clearly we need PSC to capitalize on LOC changed. :-)

PR #59, opened 2024-01-17
Include the runtime system in the bootstrapped files

jberdine:
Currently the compiler is bootstrapped by keeping the llvm bitcode of the compiler in source control, building the current version of the C/C++ code constituting the runtime system, and compiling and linking the bitcode with the runtime. This makes it possible for the bootstrap compiler and runtime system to get out of sync, and either fail to build or produce a buggy executable.

This PR changes the compilation of the 64-bit runtime system to produce llvm bitcode instead of a native archive, adds this bitcode to the files maintained in the compiler/bootstrap directory, and adjusts the build accordingly.

One side benefit is that the prelude no longer needs to build the runtime from source, it can simply use the bitcode version.


beauby (approved):
LGTM

beauby (commented):
There's a suspicious segfault in CI though.

jberdine:
CI is happy up until the last commit, so I suspect that there is something that is not independent enough between aarch64 and x86_64 bitcode.

I've put up https://github.com/SkipLabs/skdb/pull/65 as an alternative work-around for the main problem this PR was trying to solve.

gregsexton:
Is the memory issue you're seeing with the build happening only on this PR or it's orthogonal?

jberdine:
Orthogonal. I have tried building the compiler at each point in the git history where the bootstrap bitcode was updated back for about a year to the point where Lucas moved the runtime directory. None of them build reliably for me. Makes me very suspicious that it is my environment/hardware.

PR #62, opened 2024-01-18
Tweak JS tests to avoid synchronization timing flakiness

bennostein:
I _think_ that adding retries to the JS privacy test expectations should get rid of the flakiness that we're seeing in CircleCI.

This works locally for me -- passing 20 times in a row -- so hopefully CircleCI agrees.


gregsexton:
lgtm

PR #64, opened 2024-01-18
[skc] Exit on internal errors

jberdine:
If exceptions escape out of functions run by `runWithGc`, they are
just printed and ignored. In the case of skc, this results in skargo
not noticing when compilation fails.

### DIFF
--- a/compiler/src/main.sk
+++ b/compiler/src/main.sk
@@ -69,7 +69,13 @@ fun main(): void {
     _context ~> {
       SKStore.CStop(
         Some((context, _, _) ~> {
-          _result = compile(config, context, config.unknown);
+          try {
+            compile(config, context, config.unknown)
+          } catch {
+          | exn ->
+            print_error(exn.getMessage());
+            skipExit(2)
+          }
         }),
       )
     },


gregsexton:
> they are just printed and ignored

By skc?

jberdine:
By `Context.runWithGcIntern` really, see [here](https://github.com/SkipLabs/skdb/blob/main/skfs/src/Context.sk#L1455-L1464) . Note that if the function passed in throws, its exception message is printed and the context is returned, so the skc-specific code has no idea it has failed once `runWithGc` returns. Maybe I'm confused, but that is my understanding.

jberdine:
I ran into this when trying to make sense of confusing behavior when changing the runtime, and noticed that there was no skc binary in the appropriate stage directory. Then I found when passing skargo -v, that skc had failed but skargo hadn't noticed and so the skc that ended up being used was the next one in the PATH, which didn't work.

compiler/src/main.sk
```diff
@@ -69,7 +69,13 @@ fun main(): void {
     _context ~> {
       SKStore.CStop(
         Some((context, _, _) ~> {
-          _result = compile(config, context, config.unknown);
+          try {
+            compile(config, context, config.unknown)
+          } catch {
+          | exn ->
+            print_error(exn.getMessage());
+            skipExit(1)
```
beauby: What do you think about a special exit code to distinguish between "your code is faulty" and "the compiler is broken"?

gregsexton:
I've also run in to skargo not realising that skc failed - and it's super confusing! A different exit code makes sense to me.

jberdine:
Ok, I'll change the exit code. No preferences I guess? I don't have much experience with distinguishing exit codes, have never gone beyond success, internal failure / uncaught exception, crash, oom.

jberdine:
Changed the exit code on uncaught exception to 2. I surveyed existing uses of skipExit and 0 is success, obviously; 1 is wrong arguments and other things that could be categorized as "user error", as well as failures to run clang or find the expected version of llvm; 2 is often used for skc internal errors, and 4 is used for no input files.

gregsexton:
Ship it!

PR #65, opened 2024-01-18
Add compiler build stage that uses the host compiler

jberdine:
The current bootstrap procedure does not keep the compiler and runtime system in sync. The stage0 build compiles the current runtime sources together with the checked-in compiler bitcode, and then the stage1 build compiles the current compiler and runtime sources. This means that the current runtime sources have to work for both the bootstrap and current compiler sources. This makes it difficult to make changes that require modifications in both the compiler and runtime.

This PR adds an alternate build process that starts with the host skc, which is already linked with its contemporary runtime system, instead of the bootstrap bitcode. In particular, stageA builds the current runtime sources, and copies the host skc. Then stageB compiles the current compiler sources using the stageA compiler and links them with the current runtime compiled during stageA. Then stageC compiles the current compiler sources using the stageB executable. It seems necessary to copy the host compiler so that it can be in the same directory as the new runtime system library, as that is where skargo looks for it. Note that due to the way builds are time-stamped in magic.c, builds are guaranteed to be non-reproducible and no fixed-point will ever be reached.

### DIFF
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -74,6 +74,26 @@ stage3/bin/skc stage3/bin/skfmt stage3/bin/skargo: stage2
 	(cd ../skargo && PATH=../compiler/stage2/bin:$(PATH) skargo build -r --target-dir ../compiler/stage3/bin)
 # 	TODO: Ensure stage2 == stage3.
 
+stageA/bin/skc:
+	mkdir -p stageA/bin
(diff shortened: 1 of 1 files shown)


beauby (approved):
LGTM

I am also working on a cleaner stage0 mechanism (based on pre-built skip packages) that should fix this issue, but it's not quite ready yet, so let's merge this first.

PR #66, opened 2024-01-19
Support for JSON in skdb

pikatchu:
This PR introduces a new type: JSON
    
A JSON type is an object that is represented internally with an array
of values, coupled with an array of field names. The field names are
interned, so if two objects share the same field names, the table will
be shared between the two objects.
    
JSON objects cannot be decomposed directly in SQL, instead, we
introduce a new primitive that operates at the table level: json_split.
    
This primitive can only be used in a virtual view (for now?). It takes
as parameter an sql table name, that table is expected to have at
least 2 columns, where the first one is expected to be an TEXT ID and
the second one either TEXT or JSON (containing a JSON object).
    
CREATE VIRTUAL VIEW paths AS json_split(my_table);
    
This virtual view creates something of the form:
    ID|PATH|TYPE|INTEGER|FLOAT|TEXT
    
For example, given the object {"a": 3}, the split will produce:
    
    'f7xB8m2VDA5JqzOW4tkSlX_7Lw7'|.a|'int'|3|NULL|NULL
    
The idea is then to let people do their queries on this form.


gregsexton:
CI is unhappy. I think this is an issue that Benno fixed, so rebasing may help.

I'll review this but it's going to take me a while so will set aside time.

Does this support being able to index and query for documents/objects matching an email address if they look like this?

```{json)
{
  "name": "Greg",
  "emails": [
    {
      "type": "work",
      "address": "greg@skiplabs.io"
    },
    {
      "type": "personal",
      "address": "greg@gmail.com"
    }
  ],
  "other_stuff": {}
}
```

pikatchu:
> {
>   "name": "Greg",
>   "emails": [
>     {
>       "type": "work",
>       "address": "greg@skiplabs.io"
>     },
>     {
>       "type": "personal",
>       "address": "greg@gmail.com"
>     }
>   ],
>   "other_stuff": {}
> }

It does!

```SQL
CREATE VIRTUAL VIEW user_paths AS json_split(users);

CREATE VIRTUAL VIEW emails AS
  SELECT id, svalue AS email
   FROM user_paths
   WHERE path like '.emails.[%].address';

CREATE INDEX email_idx ON emails(email);
```

And then you can search on emails directly

gregsexton:
That's an interesting solution! `%` is greedy right? Maybe we need a non-greedy wildcard to make this safer. I could imagine in complex docs this accidentally matching something you didn't intend. Perf should be better too as a nice side effect.

gregsexton:
Should it be `'.emails.[%].address'` or `'.emails[%].address'`? The latter is practically standard AFAIK for json paths.

To avoid the wildcard we could probably have a few functions that normalise paths. Maybe one that strips indexes from `[]` operators. Maybe one that strips `[]` entirely and 'flattens' the list. So you could do:

```
CREATE VIRTUAL VIEW user_paths AS json_split(users);

CREATE VIRTUAL VIEW emails AS
  SELECT id, svalue AS email
   FROM user_paths
   WHERE json_path_flatten(path) = '.emails.address';

CREATE INDEX email_idx ON emails(email);
```

skjson/src/FlattenType.sk
```diff
@@ -0,0 +1,211 @@
...
+    yield (id, mkPath(), Some(PTBool()), Some(0), None(), None())
+  | CJInt(i) -> yield (id, mkPath(), Some(PTInt()), Some(i), None(), None())
+  | CJFloat(f) -> yield (id, mkPath(), Some(PTFloat()), None(), Some(f), None())
+  | CJString(s) ->
+    yield (id, mkPath(), Some(PTString()), None(), None(), Some(s))
+  | CJArray(arr) if (arr.isEmpty()) ->
+    yield (id, mkPath(), Some(PTEmptyArray()), None(), None(), None())
+  | CJArray(arr) ->
+    for (i => v in arr) {
+      for (elt in flattenHelper(id, List.Cons(".[" + i + "]", path), v)) {
```
gregsexton: Can we change this to `List.Cons("[" + i + "]", path)`. i.e. remove the dot. This makes the path compatible with all the json tooling I've ever used and with JS too.

gregsexton: Just to sanity check myself :D
```
> a = {b: [{c: 0}]}
{ b: [ { c: 0 } ] }
> a.b[0].c
0
> a.b.[0].c
a.b.[0].c
    ^

Uncaught SyntaxError: Unexpected token '['
```

gregsexton:
@beauby there are changes to the parser in here which might be worth looking at. I think you now know SQL grammar better than anyone!

sqlparser/src/AstPp.sk
```diff
@@ -4,14 +4,18 @@ extension base class Stmt {
...
       (if (virtual) "VIRTUAL " else "") +
       "VIEW " +
       (if (ifNotExists) "IF NOT EXISTS " else "") +
       name.toString() +
       " AS " +
-      query.toString()
+      subQuery match {
+      | SQuery(query) -> query.toString()
+      | SCall(funName, args) ->
+        funName.toString() + "(" + args.map(x -> x.toString()).join(",")
```
gregsexton: Missing the close paren?

sqlparser/src/AstPp.sk
```diff
@@ -380,6 +384,7 @@ extension base class Value {
   | VTime() -> "TIME()"
   | VParam(handle) -> "@" + handle
   | VPermission(s) -> "SKDB_PERMISSION(" + s + ")"
+  | VJSON(s) -> "json(" + s + ")"
```
gregsexton: Does s need any escaping?

sqlparser/src/Create.sk
```diff
@@ -41,9 +41,20 @@ extension class Parser {
 
     this.expect_keyword(TAs());
 
-    query = this.parseSelect();
+    subQuery = this.peek_keyword() match {
+    | TSelect() -> SQuery(this.parseSelect())
```
beauby, line 45: Technically a SELECT statement can also start with the VALUES keyword.

sqlparser/src/Ast.sk
```diff
@@ -57,6 +57,16 @@ base class Stmt{pos: Int = 0} uses Equality, Show {
   | Pragma{name: Name, value: ?Value}
 }
 
+base class SubQuery uses Equality, Show {
```
beauby: Maybe call this ViewDefinition?

gregsexton (commented):
This looks great. Makes a lot of sense. I don't know the design of the parser and compile/eval as well as you and @beauby, so only really skimmed these parts. But looks solid to me. Great that we're interning fields.

My only real request is - from the earlier comments - can we add some functions to work with paths to make indexing really convenient?

skjson/src/FlattenType.sk
pikatchu: Ok will do!

pikatchu: Done!

sqlparser/src/AstPp.sk
pikatchu: Yup! Good catch!

sqlparser/src/AstPp.sk
pikatchu: It does. Fixed.

sqlparser/src/Create.sk
pikatchu: Ok added a not implemented branch.

sqlparser/src/Ast.sk
pikatchu: Ok done!

prelude/ts/src/sk_types.ts
```diff
@@ -196,6 +196,35 @@ export interface WasmSupplier {
...
+      charcode = 0x10000 + (((charcode & 0x3ff) << 10) | (string.charCodeAt(i) & 0x3ff));
+      utf8.push((charcode >> 18) | 0xf0);
+      utf8.push(((charcode >> 12) & 0x3f) | 0x80);
+      utf8.push(((charcode >> 6) & 0x3f) | 0x80);
+      utf8.push((charcode & 0x3f) | 0x80);
+    }
+  }
+
+  return utf8;
+}
```
gregsexton: new TextEncoder().encode(string)

pikatchu: TextEncoder is a large dependency, that's why I did it this way.
I don't know ...

gregsexton: We use it for the protocol e.g.[1]. It's built in to browsers since ~2014. https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder

I just figure that it's more likely to handle any weird edges and it'll probably be in C so faster. Oh and every little helps with minimizing JS code size.

[1]: https://github.com/SkipLabs/skdb/blob/main/sql/ts/src/skdb_orchestration.ts#L86

pikatchu: Let me try to revisit this. might take a bit of time because I am having issues with my setup.

pikatchu: Ok so looking closer, encode is not what we need. It returns an Uint8Array which is not what we need here. We need the unicode int value for each character. I am going to check if there is something built-in that does that.

pikatchu: Ok, so I found a way to do this natively. Diff coming up!

gregsexton:
CI failed before merge. Was this something flaky?

bennostein:
Looks like flakiness to me, and I just ran `make test-wasm` without failure 20 times consecutively (locally, at trunk)

I'll push a PR raising the timeouts on synchronization in the JS tests; hopefully that improves the flakiness situation on those!

PR #67, opened 2024-01-19
Wire up query params with mirroring

gregsexton:
I already did most of the work to support this, it just wasn't hooked
up. Finally circling back round.

While here I ran in to all sorts of grief parsing json in java. Every
library wants to handle the most sophisticated cases but can't handle
a simple case of deserialising a value of unknown shape so that it can
be spliced in to another json object and then re-serialised. Ended up
migrating off of moshi and on to good ol' trusty jackson, which just
works. Oh and can work directly with bytes rather than going through
strings. Oh and it doesn't need experimental kotlin features. Win.

### DIFF
--- a/sql/ts/tests/apitests.ts
+++ b/sql/ts/tests/apitests.ts
@@ -1,5 +1,5 @@
 import { expect } from "@playwright/test";
-import { createSkdb, SKDB } from "skdb";
+import { createSkdb, SKDB, Params } from "skdb";
 
 type dbs = {
   root: SKDB;
@@ -244,7 +244,7 @@ function waitSynch(
   skdb: SKDB,
   query: string,
   check: (v: any) => boolean,
-  query_params: Map<string, any> = new Map(),
+  query_params: Params = new Map(),
   server: boolean = false,
   max: number = 6,
 ) {
--- a/sql/ts/tests/apitests.ts
+++ b/sql/ts/tests/apitests.ts
@@ -164,7 +164,7 @@ async function testQueriesAgainstTheServer(skdb: SKDB) {
     x: 42,
   });
   expect(rows).toEqual([{ x: 42, y: 21 }]);
-  await remote.exec("delete from test_pk where x in (43,44);", {});
+  await remote.exec("delete from test_pk where x in (44);", {});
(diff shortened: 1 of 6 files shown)


gregsexton:
cc @bennostein, you ran in to this recently.

sql/ts/tests/apitests.ts
```diff
@@ -213,26 +213,23 @@ async function testMirroring(skdb: SKDB) {
...
-    skdb,
-    "SELECT x,y FROM view_pk",
-    (tail) => tail[0] && tail[0].x == 42,
-  );
-  expect(viewPkRows).toEqual([{ x: 42, y: 63 }]);
+  const viewPkRows = await skdb.exec("SELECT x,y FROM view_pk");
+  expect(viewPkRows).toEqual([
+    { x: 42, y: 63 },
+    { x: 43, y: 66 },
+  ]);
```
jberdine, line 232: Not saying I see a problem, but I don't understand if/how these test changes are related.

jberdine (approved):
LGTM. Curious how much simpler the jackson-using code looks than the moshi version.

gregsexton:
> Curious how much simpler the jackson-using code looks than the moshi version.

This is the jackson deserialisation:
```
  val typeRef = object : TypeReference<HashMap<String, Any>>() {}
  val ret: Map<String, Any?> = jsonMapper.readValue(paramsBytes, typeRef)
```

I gave up on Moshi after about 70 lines. I had to write a custom parser.

sql/ts/tests/apitests.ts
gregsexton: FWIW I tried to document in the commit message: https://github.com/SkipLabs/skdb/pull/67/commits/426a37fc5b6699fe4d2c39a2970eaeabe1e106ef

The idea is that these calls should be local (we want that) and shouldn't need any resiliency around waiting for a result to come in. If they get flaky then something is very wrong.

jberdine: Ah, my bad. I failed to look at the individual commits as the whole thing wasn't overwhelming. :-)

PR #68, opened 2024-01-19
Configure port without needing a file

gregsexton:
Prod wants to use 8080 - it's a real pain and unnecessary work - to
make everything point to 3586.


bennostein (approved):
lgtm

PR #69, opened 2024-01-22
Minor runtime build simplifications

jberdine:
- Handle posix.c the same as other native-only runtime source files

- Use generic build rules for magic.bc and magic.o

- Increase debug info to -g3


gregsexton:
Can we make the %.o build -g3 while we're here?

gregsexton:
LGTM

jberdine:
> Can we make the %.o build -g3 while we're here?

Done.

PR #70, opened 2024-01-22
Fix spelling of sk_call_external_pointer_descructor

jberdine:


beauby (approved):
LGTM

jberdine:
This PR is one of the sort where `make -C compiler stage1` can't work, while `make -C compiler stageB` does. As a result, CI is unhappy. I'm somewhat motivated to change the CI job to use the latter (or maybe `stageC`) instead of the former, since the base docker image has a skc and it is not so useful to rebuilt the bootstrap bitcode every time. Thoughts?

beauby:
You probably need to update the CI docker image, as non-compiler test do not rebuild the tool chain.

jberdine:
I think that what I'm hitting is that even though the compiler ci job is triggered and builds the compiler, the skfs job does not wait for or use that just-built compiler, and so the skfs job fails.

Is what you have in mind for updating the docker image something like:
1. build the base docker image from the tip of this PR
2. push it to docker with some other version than ":latest"
3. change the base.yml file to point to refer to that new version
4. send the PR again, including a commit with the base.yml change
5. if CI is happy, merge the PR without the base.yml change, and promote the temp docker image to "latest"

That seems like enough of a pain that I'm tempted to find a good local testing procedure and then just override CI for this sort of runtime change, and just let everyone be ready with the revert hammer if a PR turns out to be unhappy in CI after merging. What do you think @beauby @gregsexton ?

beauby:
Correct, the `skfs` and `compiler` jobs are unrelated, and the former won't use artifacts from the latter. We could design a system such that if the compiler tests are triggered, then all other jobs should wait and used the produced compiler, but it would add a fair bit of complexity to CI and the initial assumption was that runtime changes would be pretty rare (which also means it's fine to override CI in this case IMO).

jberdine:
(Also, note that this PR itself is not worth much, but I'm using it as an example benign change to work through the testing process for other runtime changes.)

jberdine:
I'll incorporate this into https://github.com/SkipLabs/skdb/pull/114.

PR #72, opened 2024-01-24
Fail fast improvements

gregsexton:
A couple of improvements to get things to fail faster rather than allowing you to continue with bad assumptions.

- We check table, view, and schema creation to make sure when idempotent you really are getting the expected result. This prevents you moving forward with a misaligned schema.

- Made sure that the dev server checks the result of process calls and does not drop valuable error info.

### DIFF
--- a/sql/server/core/src/main/kotlin/core/Skdb.kt
+++ b/sql/server/core/src/main/kotlin/core/Skdb.kt
@@ -23,14 +23,14 @@ data class ProcessOutput(val output: ByteArray, val exitCode: Int) {
   }
 
   fun logOnError(): ProcessOutput {
-    if (exitCode != 0) {
+    if (!exitSuccessfully()) {
       System.err.println(decode())
     }
     return this
   }
 
   fun getOrThrow(): ByteArray {
-    if (exitCode != 0) {
+    if (!exitSuccessfully()) {
       logOnError()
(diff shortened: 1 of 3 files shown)


sql/server/core/src/main/kotlin/core/Skdb.kt
```diff
@@ -22,9 +22,16 @@ data class ProcessOutput(val output: ByteArray, val exitCode: Int) {
...
-  fun getOrThrow(): ByteArray {
+  fun logOnError(): ProcessOutput {
     if (exitCode != 0) {
       System.err.println(decode())
+    }
+    return this
+  }
+
+  fun getOrThrow(): ByteArray {
+    if (exitCode != 0) {
```
jberdine: nit: some code tests `exitCode` and some code calls `exitSuccessfully`, is there a pattern?

jberdine (commented):
The code change LGTM, but some of the existing tests hit the new checks, and some others fail but I haven't investigated.

gregsexton:
I'll investigate the test failures.

sql/server/core/src/main/kotlin/core/Skdb.kt
gregsexton: exitSuccessfully is the interface. The implementation was just using the field (which admittedly isn't private so could be used externally). I change the implementation to call exitSuccessfully.

PR #73, opened 2024-01-25
Change of the representation of indexes keys

pikatchu:
We use to represent keys of an index in a very inefficient way: this was mostly due to accommodate "INSERT OR REPLACE". That construction requires meta-data to understand the nature of the conflict. In this diff, we only reconstruct that meta-data when necessary, saving a lot of space.

This is a 20% win in space for a table that defines an index.

### DIFF
--- a/sql/src/SqlIndex.sk
+++ b/sql/src/SqlIndex.sk
@@ -6,18 +6,37 @@ module alias P = SQLParser;
 
 module SKDB;
 
-class IndexProjKey(values: Array<(Int, ?CValue)>) extends SKStore.Key
-
-fun makeIndexProjKey(
-  v: RowValues,
+class IndexProjKey private (
+  row: RowValues,
   kinds: Array<(Int, P.IKind, P.Type)>,
-  size: Int,
-): IndexProjKey {
-  indexedValues = mutable Vector[];
-  for (i in Range(0, size)) {
-    indexedValues.push((kinds[i].i0, v.values[kinds[i].i0]));
-  };
-  IndexProjKey(indexedValues.toArray());
+) extends SKStore.Key {
+  static fun create(values: Array<(Int, P.Type, ?CValue)>): this {
+    row = RowValues::create(values.map(x -> x.i2));
+    kinds = values.map(x -> (x.i0, P.IASC(), x.i1));
+    static(row, kinds)
+  }
+
+  static fun createFromRowValues(
+    row: RowValues,
+    kinds: Array<(Int, P.IKind, P.Type)>,
+  ): IndexProjKey {
+    static(row, kinds)
+  }
+
+  fun toArray(): Array<(Int, ?CValue)> {
+    indexedValues = mutable Vector[];
+    for (i in Range(0, this.kinds.size())) {
+      indexedValues.push((this.kinds[i].i0, this.row.values[this.kinds[i].i0]));
+    };
+    indexedValues.toArray()
+  }
+
+  fun compare(other: SKStore.Key): Order {
+    other match {
+    | IndexProjKey(row2, _) -> compareRows(this.kinds, this.row, row2)
+    | _ -> this.getClassName().compare(other.getClassName())
+    }
+  }
 }
 
 fun createIndexName(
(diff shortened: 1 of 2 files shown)


sql/src/SqlIndex.sk
```diff
@@ -6,18 +6,37 @@ module alias P = SQLParser;
...
+    indexedValues = mutable Vector[];
+    for (i in Range(0, this.kinds.size())) {
+      indexedValues.push((this.kinds[i].i0, this.row.values[this.kinds[i].i0]));
+    };
+    indexedValues.toArray()
+  }
+
+  fun compare(other: SKStore.Key): Order {
+    other match {
+    | IndexProjKey(row2, _) -> compareRows(this.kinds, this.row, row2)
```
jberdine, line 36: It is not obvious to me why it is ok to ignore the `kinds` of `other`. Is the situation where `this.row` and `row2` but the associated kinds are different impossible?

sql/src/SqlIndex.sk
```diff
@@ -6,18 +6,37 @@ module alias P = SQLParser;
 
 module SKDB;
 
-class IndexProjKey(values: Array<(Int, ?CValue)>) extends SKStore.Key
-
-fun makeIndexProjKey(
-  v: RowValues,
+class IndexProjKey private (
+  row: RowValues,
   kinds: Array<(Int, P.IKind, P.Type)>,
```
jberdine, line 11: Other than the call to `compareRows`, only the `Int` component is used, so maybe a little more memory can be saved:
```suggestion
  kinds: Array<Int>,
```
Since these `kinds` values are always constructed with `P.IASC()`, only the first `Int` component of `kinds` is non-constant and used by `compareRows`, so a specialized version could be written.

sql/src/SqlIndex.sk
```diff
@@ -6,18 +6,37 @@ module alias P = SQLParser;
...
+    row = RowValues::create(values.map(x -> x.i2));
+    kinds = values.map(x -> (x.i0, P.IASC(), x.i1));
+    static(row, kinds)
+  }
+
+  static fun createFromRowValues(
+    row: RowValues,
+    kinds: Array<(Int, P.IKind, P.Type)>,
+  ): IndexProjKey {
+    static(row, kinds)
```
jberdine, line 23:
```suggestion
    static(row, kinds.map(x -> x.i0))
```

sql/src/SqlIndex.sk
```diff
@@ -6,18 +6,37 @@ module alias P = SQLParser;
...
+  }
+
+  fun toArray(): Array<(Int, ?CValue)> {
+    indexedValues = mutable Vector[];
+    for (i in Range(0, this.kinds.size())) {
+      indexedValues.push((this.kinds[i].i0, this.row.values[this.kinds[i].i0]));
```
jberdine, line 29:
```suggestion
      indexedValues.push((this.kinds[i], this.row.values[this.kinds[i]]));
```

sql/src/SqlIndex.sk
```diff
@@ -6,18 +6,37 @@ module alias P = SQLParser;
...
-  for (i in Range(0, size)) {
-    indexedValues.push((kinds[i].i0, v.values[kinds[i].i0]));
-  };
-  IndexProjKey(indexedValues.toArray());
+) extends SKStore.Key {
+  static fun create(values: Array<(Int, P.Type, ?CValue)>): this {
+    row = RowValues::create(values.map(x -> x.i2));
+    kinds = values.map(x -> (x.i0, P.IASC(), x.i1));
+    static(row, kinds)
+  }
```
jberdine, line 17: The changes adding `P.TEXT()` seem not necessary:
```suggestion
  static fun create(values: Array<(Int, ?CValue)>): this {
    row = RowValues::create(values.map(x -> x.i1));
    kinds = values.map(x -> x.i0);
    static(row, kinds)
  }
```

jberdine (commented):
Mainly, one question on ignoring kinds when comparing rows.

I also made some suggestions to potentially reduce memory further at the cost of a specialized version of the `compareRows` function.

jberdine:
@pikatchu would you like one of us to commandeer this?

jberdine:
This was included in https://github.com/SkipLabs/skdb/pull/115

PR #75, opened 2024-01-25
Move definition of SKIP_String_eq from preamble to runtime

jberdine:
It seems that SKIP_String_eq is defined in the preamble only to be
able to have an i1 return type instead of i8. This is an optimization
that LLVM should be able to do itself, so this PR simplifies the code
by moving the definition from LLVM assembly in the prelude to C in the
runtime.


jberdine:
This change has further ripple effects in the compiler, closing for now.

PR #77, opened 2024-01-25
Add build instructions.

ltratt:


README.md
```diff
@@ -8,7 +8,42 @@ components.
...
- * LLVM 15.
- * wasm-ld 15 (this is part of LLVM, but your distribution might split it into
-   a separate package: your wasm-ld version must match your LLVM version).
- * typescript 5.2 or later.
+ * LLVM 15. Depending on your distribution you may need to install `llvm-devel`
+   or `llvm-extra` to get necessary extra parts such as `wasm-ld`.
+ * typescript (`tsc`) 5.2 or later. If your distribution doesn't package
+   typescript, you can install your own with `npm install
+   --prefix=/path/to/install typescript`.
+ * `yarn`
```
jberdine, line 16: When I set up a new dev environment, I look at the Dockerfile and replay those commands, modifying as necessary for systems that don't use apt. It might be helpful to point people to that approach. Also, the Dockerfile does significantly more, with gradle, java, playwright, sdkman, etc. I think all of those are needed to get a working development environment for making and testing code changes across the system, but maybe these instructions are aiming at something else?

Makefile
```diff
@@ -89,6 +89,7 @@ build/init.sql: sql/privacy/init.sql
 clean:
 	rm -Rf build
 	find . -name 'Skargo.toml' -print0 | sed 's|Skargo.toml|target|g' | xargs -0 rm -rf
+	cd compiler && ${MAKE} clean
```
jberdine, line 92: I am not sure if we always want to clean the compiler, there are many times where the compiler doesn't change for long periods during development of the other parts of the system. The steps you have in the README don't need this change, right?

Also, elsewhere we usually use `$(MAKE)`, is there a reason to prefer `${MAKE}`?

README.md
```diff
@@ -8,7 +8,42 @@ components.
...
+
+
+## Build instructions
+
+First build and install the bootstrap compiler:
+
+```
+cd /path/to/repo/compiler && make clean && prefix=/path/to/install make install STAGE=0
```
gregsexton, line 24: Why make clean?

README.md
```diff
@@ -8,7 +8,42 @@ components.
...
+```
+
+Then build the rest of the system:
+
+```
+cd /path/to/repo && make clean && make
```
gregsexton, line 30: And here. Shouldn't need to clean. I'm not sure I've ever run it.

README.md
```diff
@@ -8,7 +8,42 @@ components.
...
+```
+
+If you also want to build the NPM packages locally:
+
+```
+cd /path/to/repo/packages/dev
+rm -rf node_modules && yarn install && yarn build && yarn link
+cd ../react
+rm -rf node_modules && yarn install && yarn build && yarn link
```
gregsexton, line 39: Shouldn't need to wipe out node_modules. And yarn link is definitely not needed for build. This sets this repo up to be linked in to another on the filesystem if you want to develop with a local package.

README.md
```diff
@@ -8,7 +8,42 @@ components.
...
+```
+
+To build local copies of the Docker images:
+
+```
+cd /path/to/repo
+bin/build_docker_images.sh
+```
+
+This will build and add to Docker an image named ``
```
gregsexton, line 49: EOF error

README.md
jberdine: Probably the fault of my suggestion. For a fresh checkout, clean is not needed. But if you make changes to the compiler or runtime after building, the Makefile will not notice and `make -C compiler` will be a no-op until it is cleaned.

gregsexton: *face-palm*

PR #78, opened 2024-01-26
Change our js/ts/npm build config to get types working in downstream projects

gregsexton:
This was a journey. And I've had some journeys with JS. I hope it's
over. But I'm definitely playing with fire here.

The goal: when you use skdb as a dependency in a downstream project,
you don't get any typing info other than for the one exported function
defined in skdb.ts. I wanted to fix this.

I tried a lot of things, but eventually found compiler and npm config
that works and is recommended by the typescript documentation.

https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library

I had to make a lot of adjustments to how we define imports/exports to
make typescript happy under this new config. Which then broke what
sknpm produces, so I fixed that.

Something to note -- which sknpm was already taking care of, so I left
well alone -- is that using paths in libraries is bad and we should
make sure they never leak out in to the deliverable. They are not
re-written by the compiler and require all downstream libraries to
define them. Complete insanity.

https://stackoverflow.com/a/52501384
https://github.com/Microsoft/TypeScript/issues/15479

I'm somewhat confident this works and is correct, but it's JS and
there are many many permutations of env, so who knows?

- The tests all pass
- My node service that I'm working on gets type info, compiles, and runs
- I checked out the todo app as a vite example and this appears to load up, run, and work

### DIFF
--- a/prelude/ts/src/package.json
+++ b/prelude/ts/src/package.json
@@ -4,5 +4,6 @@
   },
   "dependencies": {
     "ws": "^8.12.0"
-  }
+  },
+  "type": "module"
 }
--- a/prelude/ts/src/sk_browser.ts
+++ b/prelude/ts/src/sk_browser.ts
@@ -1,5 +1,5 @@
-import { float, int, Environment, Wrk, Shared } from "#std/sk_types";
-import { MemFS, MemSys } from "#std/sk_mem_utils";
+import type { float, int, Environment, Wrk, Shared } from "./sk_types.js";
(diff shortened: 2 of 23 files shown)


gregsexton:
@bennostein, when you're back, I'd really appreciate you testing the todo app with a package built from these changes. I think I tested it and it worked but I think there's multiple branches etc right so I'm not confident. Note you'll need to rebuild sknpm first.

skiplabsdaniel (approved):
LGTM

PR #80, opened 2024-01-26
Add build instructions.

ltique:


README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
 
-You will need at least the following installed:
+You can run the latest release of SKDB's development server with Docker:
 
- * LLVM 15.
- * wasm-ld 15 (this is part of LLVM, but your distribution might split it into
-   a separate package: your wasm-ld version must match your LLVM version).
- * typescript 5.2 or later.
+```sh
+docker run -it -p 3586:8080 skiplabs/skdb-dev-server
```
jberdine, line 12: This will need to be updated once https://github.com/SkipLabs/skdb_www/pull/66 lands, right?

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+
+## Building From Source
+
+It is recommended that you use the prepackaged version of SKBD in most
+situations.
+
+For those who need to build from source, you will need a number of
+dependencies installed: SKDB's
+[`Dockerfile`](https://github.com/SkipLabs/skdb/blob/main/Dockerfile) contains
+the definitive list for Ubuntu, and will give users of other operating systems
```
jberdine: If we want to switch to e.g. debian this reference to ubuntu will become stale and we won't notice. Maybe:
```suggestion
the definitive list used for SKDB's CI, and will give users of other operating systems
```

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+a good idea of what is needed. Note that you will need LLVM 15, and that other
+tools relevant to LLVM need to be compatible with LLVM 15.
+
+If your distribution/OS doesn't include typescript, you can install your own
+with `npm install --prefix=/path/to/install typescript`.
+
+
+### Build steps
+
+These build steps can be simplified for your first build, but for simplicities
```
jberdine:
```suggestion
These build steps can be simplified for your first build, but for simplicity's
```

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+sake we assume that you are doing multiple builds, and need to clean artefacts
```
jberdine:
```suggestion
sake we assume that you are doing multiple builds, and need to clean artifacts
```

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+from previous builds: this is also safe, if unnecessary, for your first build.
+
+First build and install the bootstrap compiler:
+
+```sh
+cd /path/to/skdbrepo/compiler && make clean && prefix=/path/to/install make install STAGE=0
+```
+
+Then build the rest of the system ensuring that `/path/to/install` from the
+previous step is in your `$PATH`:
```
jberdine: If the advice is to use a custom install prefix, it is just as easy to tell people to ensure that `/path/to/skdbrepo/compiler/stage0/bin` is in `PATH`, and change `make install STAGE=0` to `make default STAGE=0`.

README.md
ltique: Indeed.

README.md
ltique: You're right, but also I don't see how we can change this: the Dockerfile only tells you what's right for Ubuntu. It's not even necessarily right for Debian (though it will probably work) and for other distributions, let alone other OSs, with different packages you've potentially got some guesswork to do, even with the Dockerfile as a partial guide.

README.md
ltique: Excellent idea! Fixed in https://github.com/SkipLabs/skdb/pull/80/commits/fb935fea05d0be47bf41d0b0462b1547e00972c8.

README.md
jberdine: Yes I agree that it is ubuntu specific, but the dockerfile itself states that, so to me it seems more robust to leave that detail out of the sentence that essentially just points people to the dockerfile. Once they see it, it will be clear that it is for ubuntu, and some adjustment will be needed for other OSes.

README.md
jberdine: In case it is helpful, my current workflow involves running the following command:
```
(set -a; SKARGO_PROFILE=dev; STAGE=C; PATH=$PWD/compiler/stage$STAGE/bin:$PATH; set +a; make clean && cd compiler && make clean test && cd .. && make test)
```

ltratt: What does `STAGE=C` do?

README.md
ltique: Fixed in e801d35e.

README.md
jberdine: Stage{A,B,C} are like stage{0,1,2} but where stage0 builds the bitcode checked in to the compiler/bootstrap dir, stageA uses the skc on the host found in PATH. This avoids things going out of sync when the runtime has changed since the bootstrap bitcode has been updated. See https://github.com/SkipLabs/skdb/pull/65 for more. It is useful for testing changes to the compiler and runtime system, but probably irrelevant for people just wanting to build from source.

ltique: Right, got it!

ltique:
@jberdine I think this one might be ready for merging?

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+
+```sh
+cd /path/to/skdbrepo/compiler && make clean && default STAGE=0
+```
+
+Then build the rest of the system ensuring that `$PATH` points to the compiler
+built in the previous stage:
+
+```sh
+cd /path/to/skdbrepo && make clean && PATH=/path/to/skdbrepo/compiler/stage0/bin:$PATH make
```
gregsexton, line 51: Let's not encourage cleaning? I think it looks amateur and the PR aspect of our readme is important for the product success. Cleaning is not necessary for skdb. I think it might be for the compiler, but for somewhat good reasons?

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+
+```sh
+cd /path/to/skdbrepo/packages/dev
+rm -rf node_modules && yarn install && yarn build
+cd ../react
+rm -rf node_modules && yarn install && yarn build
+```
+
+If you wish to use the NPM packages locally add `&& yarn link` after both `yarn
+build` commands.
```
gregsexton, line 64: This will be known to any JS dev.

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+cd /path/to/skdbrepo && make clean && PATH=/path/to/skdbrepo/compiler/stage0/bin:$PATH make
+```
+
+To build the NPM packages locally:
+
+```sh
+cd /path/to/skdbrepo/packages/dev
+rm -rf node_modules && yarn install && yarn build
+cd ../react
+rm -rf node_modules && yarn install && yarn build
```
gregsexton, line 60: No need to wipe out the modules.

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
+
+```sh
+cd /path/to/skdbrepo
+bin/build_docker_images.sh
+```
+
+This will build and add three Docker images locally. Note that if you have
+previously built/installed SKDB Docker images, you should have stopped both any
+running containers (`docker container prune`) and their associated images
+(`docker image prune`) before running `build_docker_images.sh`.
```
gregsexton, line 76: You shouldn't need to stop containers nor prune anything. I regularly keep several versions of images around.

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
...
- * LLVM 15.
- * wasm-ld 15 (this is part of LLVM, but your distribution might split it into
-   a separate package: your wasm-ld version must match your LLVM version).
- * typescript 5.2 or later.
+```sh
+docker run -it -p 3586:8080 skiplabs/skdb-dev-server
+```
+
+The three SKDB packages (`skdb`, `skdb-dev`, and `skdb-react`) are all
+available as normal NPM packages.
```
gregsexton, line 16: I suggest:

Our Javascript packages are all available on NPM. Install with: `npm i skdb skdb-dev skdb-react`.

gregsexton: JS devs will know what this means and it's a pretty standard instruction, even the abbreviated 'install'.

README.md
```diff
@@ -4,11 +4,73 @@ This repository contains the source code for SKDB and its associated
 components.
```
gregsexton, line 4: Let's add a call out to check out the docs:

https://www.skdb.io/quickstart/

PR #82, opened 2024-01-26
Add test-skfs target, unbreak tests

jberdine:
The skfs tests, which includes the tests for the stdlib, are not run
by any of the existing test targets, and seem to be only rarely
triggered in the CI. As a result they have been broken for a while.

This PR adds a `test-skfs` target, makes `test` depend on it, and
unbreaks the tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,11 @@ fmt:
 
 .PHONY: test
 test:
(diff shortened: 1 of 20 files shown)


gregsexton:
Great stuff!

PR #84, opened 2024-01-29
Make db creation idempotent

gregsexton:
This makes partial failure and retries much easier. It also gives us a
mechanism of fetching the root credentials for a database if we need
them. Something that couldn't be done remotely before.

The dev server is not impacted, it only creates a db after checking
that it needs to. It creates databases just in time and doesn't
support explicit creation.


bennostein (approved):
LGTM

PR #85, opened 2024-01-29
Update devserver dockerfile to match changed skdb-base

bennostein:
Use renamed base image, and drop no-op copy which was causing errors.

### DIFF
--- a/sql/server/dev/Dockerfile
+++ b/sql/server/dev/Dockerfile
@@ -1,4 +1,4 @@
-FROM skiplabs/skdb as skdb
+FROM skiplabs/skdb-base as skdb
 FROM eclipse-temurin:20 as skgw
 
 COPY . /skdb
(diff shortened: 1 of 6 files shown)


bennostein:
I've pushed latest+quickstart versions of skdb-base and skdb-dev-server using these changes, and NPM packages to match.  Everything seems good on my end, but shout if that's broken you.

sql/server/dev/Dockerfile
```diff
@@ -14,7 +14,6 @@ FROM eclipse-temurin:20 as dev
 
 WORKDIR /skdb
 
-COPY --from=skdb /skdb/build ./build
```
gregsexton, line 17: We need skdb/build for the skdb binary that the dev server uses. Without this we don't have the binary.

sql/server/dev/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM skiplabs/skdb as skdb
+FROM skiplabs/skdb-base as skdb
```
gregsexton, line 1: This should be skiplabs/skdb which builds on skiplabs/skdb-base. The dockerfile for this is in sql/Dockerfile. I refactored to make building for prod possible. I think I might have missed updating a script that needs to build this for the dev server?

gregsexton (changes requested):

gregsexton:
I imagine the dev server you built and pushed doesn't work if you try to connect to it?

bennostein:
Yes, you're right.  I think I jumped the gun here with these changes after running into some unrelated issues building a new devserver image yesterday.  Spent most of yesterday afternoon trying to figure that out; will keep investigating today; closing this one now.

PR #91, opened 2024-01-31
Fix: replication was broken when skdb_author column is defined

gregsexton:
Replication was working correctly and delivering the row but the local
write checker was preventing the write to the local dir. It's
safeguarding it in the same way it doesn't let you insert a row with
the wrong skdb_author.

The design of write-csv assumes that, on the client, it is operating
effectively as root and can do whatever it likes. writeChecker breaks
this assumption.

I fixed this by allowing write-csv to disable the write checker. See
the commit message for why I disable using a flag rather than some
other approach - I did put some thought in to this. :D

I disable the check rather than moving it higher in to the SQL layer
to ensure that with future changes we don't forget to check writes -
which although wouldn't be terrible - we don't trust the client and
bad writes are still rejected at the server - would create a very
annoying dev experience and be very confusing. In sum, I do agree that
this concept belongs in the skfs layer.

### DIFF
--- a/skfs/src/EagerDir.sk
+++ b/skfs/src/EagerDir.sk
@@ -1722,6 +1722,7 @@ class EagerDir{
     writerPath: Path,
     k: Key,
     rvalues: Array<File>,
+    force: Bool = false,
   ): this {
     origSource = Source::create(origSourcePath);
     oldValues = this.getArraySourceKey(origSource, k);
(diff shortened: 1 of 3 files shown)


bennostein:
Looks good to me!  Thanks for adding the test, too.

And, the reasoning about a bool parameter vs. nulling out the checker sound solid to me.

PR #93, opened 2024-01-31
Make group ids visible in SKDBGroup

gregsexton:
These are needed, particularly groupID, for writing to skdb_access.
The implementation defines the fields - and they're used in our tests
- but type checking fails in downstream packages.


bennostein (approved):
lgtm

Issue #92, opened 2024-01-31
Adding a member to a group leaks this information to all users.


gregsexton:
```
      const grp = await skdb.createGroup();
      await grp.setMemberPermission(someUser, "r");
```

Results in 
```
select * from skdb_group_permissions;
┌─────────┬───────────────────────────────┬───────────────────────────────┬─────────────┬──────────────┐
│ (index) │            groupID            │            userID             │ permissions │ skdb_access  │
├─────────┼───────────────────────────────┼───────────────────────────────┼─────────────┼──────────────┤
│    0    │ '3ZVVNSpKUkylYNO0ZIGm7Ey5oI1' │ 'fA96m2waGvUIAw-dyA9Z3I75IJC' │      4      │ 'read-write' │
```

The read-write permission means that all users can see that this user is in the group with permission 4. We should only be sharing this information with the user and group admins.

This info could be used to figure out e.g. a user's friend list or worse depending on the app.

PR #94, opened 2024-01-31
Limit permission visibility to admins and relevant users

bennostein:
@gregsexton, this is what I had in mind as a followup to the SKDBGroup stuff and alluded to earlier today -- I think it should fix the issues you surfaced with #92, and also adds a test to check/prevent regression.

It's documented a bit inline and in commit messages, but the idea is that users should only see `skdb_group_permissions` rows if (1) they have sufficient permissions to see the group in `skdb_groups` and (2) either they are an admin of the group or the permission could affect them.


gregsexton:
This is great! Thanks so much for working on this and fixing quickly.

PR #95, opened 2024-02-01
Check length expectations after replication converges in JS tests

bennostein:
Use waitSynch more liberally in JS tests to force convergence, and check expectations on the results.

This reinstates some tests I had inadvertently skipped a couple weeks ago assuming that waitSynch asserts its predicate on convergence.  Just ran `make test-wasm` 50x in a row on my machine without failure, but will keep an eye out for any flakiness in CI going forward

### DIFF
--- a/sql/ts/tests/apitests.ts
+++ b/sql/ts/tests/apitests.ts
@@ -487,14 +487,21 @@ async function testJSPrivacy(skdb: SKDB, skdb2: SKDB) {
 
   await skdb.exec("INSERT INTO test_pk VALUES (37, 42, 'my_group');");
 
-  let user1_view = await skdb.exec(
-    "SELECT * FROM test_pk WHERE skdb_access='my_group';",
-  );
-  let user2_view = await skdb2.exec(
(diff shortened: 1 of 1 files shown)


sql/ts/tests/apitests.ts
```diff
@@ -487,14 +487,21 @@ async function testJSPrivacy(skdb: SKDB, skdb2: SKDB) {
...
-  );
-  expect(user1_view).toHaveLength(1);
-  expect(user2_view).toHaveLength(0);
+  expect(
+    await waitSynch(
+      skdb,
+      "SELECT * FROM test_pk WHERE skdb_access='my_group';",
+      (rows) => rows.length == 1,
+    ),
+  ).toHaveLength(1);
```
gregsexton, line 496: waitSynch runs the query on the remote side, but it looks like here - and all your other cases - were testing that things are correct on the client-side. Is this what you intended?

gregsexton: And I've just noticed the server flag defaulting to false. Ignore this comment.

gregsexton:
LGTM

PR #97, opened 2024-02-01
Fix: return any rather than Object

gregsexton:
Object is a pain to work with. Typescript doesn't allow it to be cast
easily, so you can't add typing info, and it doesn't like you using
properties as accessors, which makes code messy.

### DIFF
--- a/sql/ts/src/skdb_util.ts
+++ b/sql/ts/src/skdb_util.ts
@@ -38,7 +38,7 @@ export class ExternalFuns {
 /* Class for query results, extending Array<Object> with some common selectors
    and utility functions for ease of use. */
 /* ***************************************************************************/
-export class SKDBTable extends Array<Object> {
+export class SKDBTable extends Array<Record<string, any>> {
   scalarValue(): any {
     const row = this.onlyRow();
     const cols = Object.keys(row);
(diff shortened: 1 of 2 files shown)


sql/ts/src/skdb_util.ts
```diff
@@ -51,7 +51,7 @@ export class SKDBTable extends Array<Object> {
     return row[cols[0]];
   }
 
-  onlyRow(): Object {
+  onlyRow(): any {
```
jberdine: I don't claim to know what I'm talking about, but when I was looking into enabling eslint I ran into various advice to use `Record<string, unknown>` or just `unknown` instead of `any`. Just mentioning.

jberdine (approved):

sql/ts/src/skdb_util.ts
gregsexton: Oh very nice, I didn't know about Record. I'll look in to changing to this. And this is a good enough reason to finally setup eslint too. Watch this space.

jberdine: https://github.com/SkipLabs/skfs/pull/731 might be helpful

gregsexton:
I've tried to track down any usage of Object and use Record where possible. We can probably do better but this seems like an improvement for now.

Issue #99, opened 2024-02-01
We're not wrapping tables in SKDBTable when using a worker??


gregsexton:
Not sure the title is correct, but I don't know what else is going on.

I run an app using skdb-react and the dev console and it doesn't load the users when run as a worker, but it does when not run as a worker.

I get this error in the console.
```
dev-console.js:105 Uncaught (in promise) TypeError: userRows.onlyColumn is not a function
    at dev-console.js:105:35
```

bennostein:
Is this with `exec`? That should be wrapping [here](https://github.com/SkipLabs/skdb/blob/5b9c3b5117e45e635a85687f472e295f9f733463/sql/ts/src/skdb_wdatabase.ts#L104), not sure what else is going on.

gregsexton:
Yeah. It's this line that's the problem:

https://github.com/SkipLabs/skdb/blob/main/packages/react/src/dev-console.tsx#L220

With a worker it fails complaining about the existence of onlyColumn. Without it's fine.

There could be some resolution confusion as the dev console does also define a function called SKDBTable. I tried renaming that and it didn't seem to make a difference. But my npm link might not have worked or vite may have been caching stuff now I think about it.

bennostein:
I've now repro'ed this in the demo todo-app with new versions of everything, should have it fixed shortly.

PR #100, opened 2024-02-01
Use SKDBTable in WrappedRemote worker SKDB

bennostein:
Don't ask me how TypeScript failed to catch this obvious type error...


bennostein:
Fixes #99

gregsexton:
Amazing! Thanks so much for fixing this. I didn't have time to look yesterday, but thought it would be something silly.

And, Typescript...

PR #101, opened 2024-02-02
Fix MInfo creation

skiplabsdaniel:


gregsexton:
Can you explain the change a little?

skiplabsdaniel:
MInfo is a structure used to saves some file dependencies data (generated keys, created subdirectories and reads that have taken place during a mapped file creation). Until now, when only one key was created, the created directories and reads were systematically ignored, with the result that when the source file was deleted, the generated subdirectories were not deleted.
Fixing this problem was relatively straightforward: save the list of directories and their reads, when exists.
In doing so, it revealed issues occurring when deleting subdirectories, such as in the case of watch, when a watch was deleted, the new Minfo was empty and therefore removed from the structure conserving this information. However, the set of dependencies meant that the file in question was updated again, and as the information that the MInfo had not been explicitly saved, the update retrieved the information relating to the fixed data, which caused the entire dependency chain to be re-looped, generating an infinite loop.
Once identified, fixing this problem was straightforward: always save the Minfo.

The second consequence was that it was no longer possible to share existing SELECT at VIRTUAL VIEW level, the principle being that is when a view entry is created, it checks whether a table relative to the SELECT already exists, otherwise it creates it.
Up until now, there had been no problem, as in the case where sub-directory creation data was not saved, SELECT related directories were never deleted. However, when these are present, all entries in the virtual view are considered as owner of the SELECT related directories, so when the first entry is deleted, all the directories are deleted, which has no consequence in most cases, since they are regenerated when required. The only issues arises when the SELECT contains an aggregate function.
In that case, the directory performing the aggregate computation is also deleted, with the result that it is not updated.
Several solutions were considered to fix that issue, but they all had a heavy impact on the code.
It turns out that adding a reference counter for sub-directory owners solves the issue with very little impact on the code.
And to avoid saving useless data, the ability to identify directories authorizing multiple owners was added.

gregsexton:
LGTM. CI compiler tests failed, but I'm guessing that's not related given the changes?

skiplabsdaniel:
Indeed, it seems that the CI compiler tests fail on all PRs based on the current main.

PR #104, opened 2024-02-06
[ci] Hardcode number of jobs and revise action trigger

jberdine:


jberdine:
The compiler CI job failures are due to circleci misreporting the number of processors, this seems to fix it.

beauby:
Interesting. It shouldn't crash/hang though.

jberdine:
It is a memory issue. If I simulate using a docker container with 8 CPUs and 16 GB RAM, running `skargo test --jobs 36` and monitoring with `htop` shows that memory is exhausted almost instantly, and all hell breaks loose then. That is only ~450MB per process so it makes sense.

beauby:
That makes sense. I have a branch somewhere that mitigates (some of) the memory hungriness of sktest, I'll dig it back.

jberdine:
It can't just be sktest though. Just trying to run 36 copies of skc, each with its own clang process, in 16GB is just not going to work.

PR #105, opened 2024-02-08
[skc] Clean up verbose (`-v`) mode.

beauby:
This commit cleans up the output of verbose mode. The formerly displayed information is now kept behind `--debug`. Moreover, the new verbose mode displays the linker invocation and its output.

### DIFF
--- a/compiler/src/AsmOutput.sk
+++ b/compiler/src/AsmOutput.sk
@@ -805,7 +805,7 @@ private fun defsIsomorphic(
 
 fun mergeIdenticalAsmDefs(
   defs: mutable AsmOutput.AsmDefIDToAsmDef,
-  verbose: Bool,
+  debug: Bool,
 ): void {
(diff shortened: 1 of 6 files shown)


compiler/src/AsmOutput.sk
```diff
@@ -805,7 +805,7 @@ private fun defsIsomorphic(
 
 fun mergeIdenticalAsmDefs(
   defs: mutable AsmOutput.AsmDefIDToAsmDef,
-  verbose: Bool,
+  debug: Bool,
```
jberdine, line 808: Maybe this should take the whole config?

jberdine (approved):
LGTM

compiler/src/AsmOutput.sk
beauby: This is arguable, but I personally favor limiting dependencies. I don't feel strongly about it though.

PR #111, opened 2024-02-13
Fix UTF8 in JS getChar

pikatchu:
The way getChar was implemented in JS was not matching the native implementation. The native version return a codepoint, the JS version was supposed to do the same (using a construction called .codePointAt), but it was not working on some utf8 encoded examples (for reasons that I don't understand).

This diff mimics the logic that we have natively. It would be probably worth revisiting at some point, because it would be nicer to use JS primitives to do the work, but it's not urgent. Let's fix our JS first.

The diff also includes tests on utf8 strings to make sure we don't regress.

### DIFF
--- a/prelude/ts/src/sk_types.ts
+++ b/prelude/ts/src/sk_types.ts
@@ -197,14 +197,32 @@ export interface WasmSupplier {
 }
 
 function utf8Encode(str: string): Array<number> {
-  const codePoints = new Array();
+  const utf8 = new Array();
 
-  for (const char of str) {
-    const codePoint = char.codePointAt(0);
-    codePoints.push(codePoint);
+  for (let i = 0; i < str.length; i++) {
+    let charcode = str.charCodeAt(i);
+
+    if (charcode < 0x80) {
+      utf8.push(charcode);
+    } else if (charcode < 0x800) {
+      utf8.push((charcode >> 6) | 0xc0);
+      utf8.push((charcode & 0x3f) | 0x80);
+    } else if (charcode < 0xd800 || charcode >= 0xe000) {
+      utf8.push((charcode >> 12) | 0xe0);
+      utf8.push(((charcode >> 6) & 0x3f) | 0x80);
+      utf8.push((charcode & 0x3f) | 0x80);
+    } else {
+      // surrogate pair
+      i++;
+      charcode = 0x10000 + (((charcode & 0x3ff) << 10) | (str.charCodeAt(i) & 0x3ff));
+      utf8.push((charcode >> 18) | 0xf0);
+      utf8.push(((charcode >> 12) & 0x3f) | 0x80);
+      utf8.push(((charcode >> 6) & 0x3f) | 0x80);
+      utf8.push((charcode & 0x3f) | 0x80);
+    }
   }
 
-  return codePoints;
+  return utf8;
 }
 
 export class Utils {
--- a/sql/ts/tests/tests.ts
+++ b/sql/ts/tests/tests.ts
@@ -1607,6 +1607,56 @@ export const tests = (asWorker: boolean) => {
         expect(res).toMatch(/^CREATE VIRTUAL VIEW v1/);
       },
     },
+    {
+      name: n("UTF8", asWorker),
+      fun: async (skdb: SKDB) => {
+        await skdb.exec("create table t1 (a TEXT);");
+        let objects = [
+          {data:"x\u2019s re"},
+          {data:"y\u2013 1"},
+          {data:"z\u00ae b"},
+          {data:"\u00c3\u00a9al P"},
+          {data:" \u2022 T"}
+        ];
+        for(let i in objects) {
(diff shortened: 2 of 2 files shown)


prelude/ts/src/sk_types.ts
```diff
@@ -197,14 +197,32 @@ export interface WasmSupplier {
 }
 
 function utf8Encode(str: string): Array<number> {
-  const codePoints = new Array();
-
-  for (const char of str) {
-    const codePoint = char.codePointAt(0);
```
jberdine, line 203: I'm not an expert, but I thought that "code point" in JS is a term that refers to a 16-bit integer, e.g. that could be an element of a UTF-16 sequence. And so explicitly getting UTF8 "code units" requires some other library / manual coding.

prelude/ts/src/sk_types.ts
```diff
@@ -197,14 +197,32 @@ export interface WasmSupplier {
...
+      utf8.push((charcode & 0x3f) | 0x80);
+    } else {
+      // surrogate pair
+      i++;
+      charcode = 0x10000 + (((charcode & 0x3ff) << 10) | (str.charCodeAt(i) & 0x3ff));
+      utf8.push((charcode >> 18) | 0xf0);
+      utf8.push(((charcode >> 12) & 0x3f) | 0x80);
+      utf8.push(((charcode >> 6) & 0x3f) | 0x80);
+      utf8.push((charcode & 0x3f) | 0x80);
+    }
```
jberdine, line 222: I don't understand why we use arithmetic operations in Char.utf8encode and bitwise operations here, but the functionality looks the same to me (assuming JS has sane semantics of bitwise ops).

jberdine (approved):
LGTM

prelude/ts/src/sk_types.ts
beauby: It's surprising that this is failing, as `for (... of ...)` + `codePointAt(0)` seems to be documented as working?

jberdine: Looking [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt) it seems clear that the returned "code points" are not bytes, but decode UTF-16 into (21-bit?) integers. On the other hand, [charCodeAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt) just gives the 16-bit int elements of a UTF-16 sequence, and leaves the decoding of the use of 1 or 2 of them per code point to the caller. Neither does anything to get to a UTF-8 byte sequence. I suppose that the decoding of UTF-16 surrogate pairs into a single number could be done with `codePointAt` and then that single int converted into 1-4 bytes manually in a way similar to the code in this PR. Either approach seems fine to me, not sure whether to prefer one over another.

pikatchu:
Thanks guys!

prelude/ts/src/sk_types.ts
gregsexton: > I'm not an expert, but I thought that "code point" in JS is a term that refers to a 16-bit integer, e.g. that could be an element of a UTF-16 sequence. And so explicitly getting UTF8 "code units" requires some other library / manual coding.

A code unit is one or more bytes in a character encoding (utf-8 and utf-16 are examples of char encodings. JS uses utf-16 internally for strings) that encode a (almost always unicode) code point. One or more code points are assembled together to actually get a glyph that is rendered to screen. utf-8 uses up to 4 bytes to represent a code point (historically 6 but that was deprecated) and utf-16 uses 1 or 2 (99% sure on this, it's been quite a while since I worked closely with utf-16).

gregsexton: > Looking [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt) it seems clear that the returned "code points" are not bytes, but decode UTF-16 into (21-bit?) integers.

Code points are never bytes. There are too many. There are roughly a million defined by unicode, so 21 bits sounds about right.

> On the other hand, [charCodeAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt) just gives the 16-bit int elements of a UTF-16 sequence, and leaves the decoding of the use of 1 or 2 of them per code point to the caller. Neither does anything to get to a UTF-8 byte sequence. I suppose that the decoding of UTF-16 surrogate pairs into a single number could be done with codePointAt and then that single int converted into 1-4 bytes manually in a way similar to the code in this PR.

To go from utf-16 to utf-8, you need to scan the byte sequence and convert to a code point and then back to a utf-8 representation of the code point.

gregsexton:
> It would be probably worth revisiting at some point, because it would be nicer to use JS primitives to do the work, but it's not urgent. Let's fix our JS first.

I pulled this down and played around. #113 seems to work for me locally. I'll check CI. LMK if you have concerns.

PR #112, opened 2024-02-13
Renaming "virtual views" into "reactive views"

pikatchu:
We originally called virtual view "virtual" because we did not want to use the term "materialized". The term materialize comes with baggage in the database world, those kind of queries are often very limited, experienced SQL engineers might make the wrong assumptions ...

I suggest we use the term "reactive" going forward. The term virtual is still supported in the parser until we have updated all the relevant places (docs, quickstart etc ...).


gregsexton:
You're a hero! It's going to take me ages to break the muscle memory of typing virtual, but I think this will help educate users and avoid an FAQ entry in the future.

gregsexton:
Ship it!

PR #113, opened 2024-02-13
Represent stdin as a utf8 byte array and use TextEncoder

gregsexton:
stdin is a byte stream and in my opinion should assume utf8. This
seems to be the assumption we have everywhere else. So I changed the
type to reflect this and then can use the built in TextEncoder to
convert strings.

Now this is JS so all bets are off, but I would hope this is likely
more correct and more efficient than anything we've hand rolled. But
it should definitely produce (slightly) smaller code.

All tests pass locally, including those added in pr #111.

### DIFF
--- a/compiler/runtime/runtime64_specific.cpp
+++ b/compiler/runtime/runtime64_specific.cpp
@@ -218,56 +218,6 @@ uint32_t SKIP_read_line_get(uint32_t i) {
   return lineBuffer[i];
 }
 
-uint32_t SKIP_getchar(uint64_t) {
-  int c1 = (uint32_t)getc(stdin);
-
-  if (c1 == EOF) {
-    SKIP_throw_EndOfFile();
-  }
-
-  if ((c1 & 0x80) == 0) {
-    return c1;
-  }
-
-  int c2 = (uint32_t)getc(stdin);
-
-  if (c2 == EOF) {
-    fprintf(stderr, "Invalid utf8");
-    exit(ERROR_INVALID_STRING);
-  }
-
-  if ((c1 & 0x20) == 0) {
-    return (c1 - 192) * 64 + (c2 - 128);
-  }
-
-  int c3 = (uint32_t)getc(stdin);
-
-  if (c3 == EOF) {
-    fprintf(stderr, "Invalid utf8");
-    exit(ERROR_INVALID_STRING);
-  }
-
-  if ((c1 & 0x10) == 0) {
-    return (c1 - 224) * 4096 + (c2 - 128) * 64 + (c3 - 128);
-  }
-
-  int c4 = (uint32_t)getc(stdin);
-
-  if (c4 == EOF) {
-    fprintf(stderr, "Invalid utf8");
-    exit(ERROR_INVALID_STRING);
-  }
-
-  if ((c1 & 0x8) == 0) {
-    return (c1 - 240) * 262144 + (c2 - 128) * 4096 + (c3 - 128) * 64 +
-           (c4 - 128);
-  }
-
-  fprintf(stderr, "Invalid utf8");
-  exit(ERROR_INVALID_STRING);
-  return 0;
-}
-
 thread_local void* exn;
 
 void* SKIP_getExn() {
--- a/prelude/src/stdlib/other/System.sk
+++ b/prelude/src/stdlib/other/System.sk
@@ -142,10 +142,6 @@ fun throwEndOfFile(): void {
 }
 
 @debug
-@cpp_extern("SKIP_getchar")
-native fun getChar(): Char;
-
-@debug
 @no_throw
 @cpp_runtime
 native fun print_stack_trace(): void;
--- a/prelude/ts/src/sk_runtime.ts
+++ b/prelude/ts/src/sk_runtime.ts
@@ -26,7 +26,6 @@ class LinksImpl implements Links {
   SKIP_read_line_fill!: () => int;
   SKIP_read_to_end_fill!: () => int;
   SKIP_read_line_get!: (index: int) => ptr;
-  SKIP_getchar!: () => int;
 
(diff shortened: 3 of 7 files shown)


pikatchu:
I don't understand how that can possibly work 😅
The getChar function returns a codepoint in native (if you look at runtime64_specific.cpp).
So could it be that the code works both for codepoints and bytes?

@beauby: thoughts?

beauby:
So `getChar()` (`SKIP_getchar()` in the C runtime) is only used in `sql/src/SqlCsv.sk`. Otherwise, skdb uses `read_line()` and `read_to_end()` which rely on the `SKIP_read_line_fill()`/`SKIP_read_line_get()` runtime functions (through `SKIP_read_line()` defined in `compiler/runtime/stdlib.c`, which seems to assume that `SKIP_read_line_get()` returns a single byte):
```
 122 char* SKIP_read_line() {                                                                                                                 
 123 │ int32_t size = SKIP_read_line_fill();                                                                                                  
 124 │                                                                                                                                         
 125 │ if (size < 0) {                                                                                                                        
 126 │ │ return NULL;                                                                                                                         
 127 │ }                                                                                                                                      
 128 │                                                                                                                                         
 129 │ uint32_t i;                                                                                                                            
 130 │ char* result = SKIP_Obstack_alloc(size);                                                                                               
 131 │                                                                                                                                         
 132 │ for (i = 0; i < size; i++) {                                                                                                           
 133 │ │ result[i] = SKIP_read_line_get(i);                                                                                                   
 134 │ }                                                                                                                                      
 135 │                                                                                                                                      
 136 │ result = sk_string_create(result, size);                                                                                               
 137 │ return result;                                                                                                                         
 138 } 
```

pikatchu:
Ahhhh ok.

pikatchu:
Shall file an issue to kill or rename getChar? I find this confusing.
Happy to merge this though.

jberdine:
This situation is confusing. I think that there are 2 sorts of code that we need to support:

1. Code that wants to just shuffle the representation of strings around, and so wants to operate on sequences of bytes (that happen to be UTF-8 encodings of strings).
2. Code that needs to know the semantics of the represented strings, such as checking if something is whitespace, converting to a different encoding, etc.

Right now, I don't think that we have clearly delineated APIs that support both of these. There are APIs that do each sort of thing, but they are "all mixed up" in the API. So I'd say the issue is not to kill getChar, but to make distinct APIs for manipulating the byte sequence representation of UFT-8 strings, and for operating on (and converting to and from) the represented Unicode code points and strings.

I'm no unicode expert, and have almost only ever just done code of sort 1 above, but this is my impression of the situation.

gregsexton:
Yes, looks like getChar is broken. I never looked at it, I just chased through where we actually read stdin and the functions that are called. As Lucas points out, it's not used.

getChar *is* only used in csv, but it's dead code - I forgot load csv even existed. So we can kill that easily.

For me, the fix should be to change one of the getChar implementations to be compatible (or kill them both) rather than making stdin a code point stream.

And agree with Josh, there are two layers. Java actually does this quite clearly. They have the concept of an Input/OutputStream - which is just bytes - and a Reader/Writer that layers over the top and talks in terms of characters (code points). But I doubt we need to do a bunch of modelling. We just need to make it clear what layer each function is operating at.

gregsexton:
For now, given that it's not used and is a potential sharp edge, I'll just quickly kill getChar.

jberdine:
Note though that it is called `getChar`, and `Char` in skip means a unicode code point, and definitely not a byte, so "broken" might be harsh. I'm not arguing to keep it though. Note that String::fromChars and StringIterator operate on code points and not bytes as well.

gregsexton:
Apologies, I meant: looks like getChar is broken now by me with this PR.

jberdine:
Ah! I see now.

gregsexton:
Ok, just cleaned up getChar and related functionality. Tests looked good locally but I admit I didn't wait for the last few diff ones, so will rely on CI. It's worth taking a look at the latest commit (https://github.com/SkipLabs/skdb/pull/113/commits/c3e02a1c442e6d06c163a7e9401c65b3e112be28) to check that this really is unneeded code.

jberdine:
Was the CSV code dead but wanted as (the basis for) an alternate mechanism to backup and restore a db? Otherwise LGTM.

gregsexton:
IIRC (@pikatchu?) it was meant for fast loading of data.

I've been thinking about backup and restore recently and much prefer the dump output and re-evaluating the statements. I like how explicit the text form is, which provides a lot of confidence.

ltique:
I *think* this PR is saying "stdin can be an arbitrary byte stream and if you want it to be UTF-8 you need to treat it as such?" That would be good (and I agree with the multi-layer approach that languages like Java and Rust have that does just this). Certainly, I regularly use stdin to stream non-UTF8 data, so I wouldn't want to rule that use case out!

gregsexton:
Yeah. stdin is a byte stream. Where skdb uses it for text data it assumes utf8 encoding.

> Certainly, I regularly use stdin to stream non-UTF8 data, so I wouldn't want to rule that use case out!

With skdb? Or you're talking about posix file interface?

ltique:
> With skdb? Or you're talking about posix file interface?

Oops, sorry: I was talking generally. But I'm pretty sure I have use cases where I'll want to do so with skdb (but I haven't got there yet and probably won't for some time TBH).

beauby:
I personally vote to get rid of `getChar()` and consolidate around other APIs (down the line it should ideally be those in the `IO` stdlib module, though they currently have limitations -- #11 should address some, and I'll add a proper UTF8 layer on top).

gregsexton:
This is what I've done. CI looks good so will merge now. But anyone late to the party, feel free to comment and I'll address in followups.

PR #114, opened 2024-02-14
Change string representation to support PIC

jberdine:
Change string representation to use bit 1 of the hash instead of the high bit to distinguish strings.


jberdine:
(CI is expected to fail, these changes will require a bootstrap and docker image dance.)

Just putting this up so others can see some of what is cooking. Individual commits should be readable more easily than the whole stack at once.

jberdine:
These changes are not all relevant to the string representation, but also include other fixes and cleanup that I encountered while working on strings.

mbouaziz:
Is there a prefix that can be reviewed and merged?

jberdine:
Yes. A number of these changes can't be tested in CI due to needing a more manual bootstrap procedure, and some will break CI for other PRs until the base docker image is updated. So I have been accumulating this to amortize the churn and risk. But since I'm not the only one working on these files now it makes sense to expidite merging more.

jberdine:
I need to go through and check that these commits don't undo @mbouaziz's recent improvements, and redo bootstraps after rebasing. But otherwise these commits up to but not including `[compiler] Change string hash tag bit from 31 to 3 in generated code` are ready for review and can be merged after that and another round of rebasing and testing.

The `[compiler] Change string hash tag bit from 31 to 3 in generated code` and following commits show the current state that works in native but not wasm. Using bit 3 for the string hash tag works in native since vtable pointers are 8-aligned, but not in wasm as there they are only 4-aligned. Using bit 2 does not currently work. Some other code is using bit 2 it seems. Bit 1 is used by copy and intern.

compiler/src/AsmOutput.sk
```diff
@@ -3408,8 +3408,8 @@ private fun llvmWrite(instr: Stmt, asm: mutable FunAsmDefBuilder): void {
   | StringCmpEq{lhs, rhs} ->
     llvmWriteIntrinsicCall(asm, instr, "SKIP_String_eq", Array[lhs, rhs])
   | v @ StringConcat{args} ->
-    // The runtime provides special versions for 2, 3 or 4 arguments.
-    varargs = v.args.size() < 2 || v.args.size() > 4;
+    // The runtime provides a special version for 2 arguments.
+    varargs = v.args.size() != 2;
```
mbouaziz, line 3428: `StringConcat` is produced at two places:
- in `intrinsics.sk`, `.IntrinsicsInliner.maybeInlineIntrinsic` for `String::concat` which is a method that takes only one parameter and hence will produce only `SKIP_String_concat2`
- in `rewrite.sk`, `.Rewrite.emitStringConcat`, which is never called.

Hence only the binary version is ever produced.

The suffix for the varargs version should be `"N"`, not `""`.

prelude/src/stdlib/other/System.sk
```diff
@@ -113,11 +113,6 @@ native fun print_error_raw(String): void;
 @cpp_extern
 native fun print_debug_raw(String): void;
 
-@debug
-@no_throw
-@cpp_runtime
-native fun flush_stdout(): void;
```
mbouaziz, line 119: Why doesn't the compiler give warnings for duplicates?

compiler/runtime/runtime64_specific.cpp
```diff
@@ -352,11 +349,8 @@ char* SKIP_open_file(char* filename_obj) {
   int status = fstat(fd, &s);
   size_t size = s.st_size;
 
-  char* result = nullptr;
   if (size == 0) {
-    result = sk_string_alloc(0);
-    sk_string_set_hash(result);
-    return result;
+    return sk_string_create("", 0);
```
mbouaziz: This should certainly be a static constant

compiler/runtime/runtime64_specific.cpp
```diff
@@ -365,9 +359,7 @@ char* SKIP_open_file(char* filename_obj) {
     fprintf(stderr, "Could not open file: %s\n", filename);
     exit(ERROR_FILE_IO);
   }
-  result = sk_string_alloc(size);
-  memcpy(result, f, size);
-  sk_string_set_hash(result);
+  char* result = sk_string_create(f, size);
```
mbouaziz: Sorry, I touched this function in #148

prelude/src/native/StringIterator.sk
```diff
@@ -13,8 +13,7 @@ module String;
 
 mutable class StringIterator private (
   // This object is manipulated by the native methods directly.  It's important
-  // that the fields match the native implementation
-  // (String.cpp:StringIterator).
+  // that the fields match the native implementation (string.c).
```
mbouaziz, line 16: Is there any native `StringIterator` implementation?

compiler/runtime/runtime.c
```diff
@@ -17,16 +17,24 @@
    tests this bit to determine if an object is a string or an object of some
    other type. */
 
+sk_string_t* get_sk_string(char* obj) {
+  return (sk_string_t*)(obj - sizeof(uint32_t) * 2);
```
mbouaziz: Should it be written
`(sk_string_t*)(obj - offsetof(sk_string_t, data))`?

mbouaziz (commented):
LGTM though I'm not sure I fully grasp the differences between `@cpp_extern`, `@cpp_runtime`, `@intrinsic` + what is in the preamble.

I think it's worth merging the prefix that works

jberdine:
Re the differences between `@cpp_extern`, `@cpp_runtime`, `@intrinsic` + what is in the preamble:

- `intrinsic` is for functions that the compiler itself knows about and will lower to some other code rather than just emitting a call to the named function. Sometimes a function is marked `intrinsic` if it is transformed to something by the compiler only conditionally or for some targets, in which case a default implementation is also needed. There are currently cases (e.g. `clzImpl`) where such default implementations exist but are never used. It's a TODO to clean them up.
- `cpp_runtime` and `cpp_extern` are very similar in that they declare a function for skip code and delegate the implementation to the runtime system. If a name is given, that is the C linkage name that will be used for the implementation, if no name is given, the default encoding of skip names to C linkage names is used (e.g. `SKIP_<module>_<class>_<function>`). The difference between `cpp_runtime` and `cpp_extern` is that compiler will emit a declaration for `cpp_extern` functions but will not for `cpp_runtime` ones. Therefore, the `cpp_runtime` ones need to be declared manually in the preamble. This is useful to e.g. give different but compatible types in the llvm declaration than would be generated.

compiler/src/AsmOutput.sk
jberdine: Yes, only the binary version is produced. I left it for later to decide whether to reintroduce the optimization and vararg version or to remove the support from the compiler. It was removed in https://github.com/SkipLabs/skfs/pull/142 but this seems to me to be a tangential change that we might still want. So I left it for now.

Also, my understanding is that the vararg version is a currently non-existent runtime function named `SKIP_String_concat` while the existing `SKIP_String_concatN` is a different function that concats an array of strings. I don't think that and the vararg version are / can be the same.

prelude/src/stdlib/other/System.sk
jberdine: *shrug*

compiler/runtime/runtime64_specific.cpp
jberdine: That seems like a good idea.

prelude/src/native/StringIterator.sk
jberdine: There is `SKIP_String_StringIterator__substring` in string.c, but I think that is all.

compiler/runtime/runtime.c
jberdine: I thought about that but held off for now since it would take a bit more work to be able to do it consistently that way for all of strings, arrays, and general skip objects. I was also thinking that the compiler code produces values with exact knowledge about the layout, and writing the runtime code using `offsetof` gives the impression that things are robust with respect to the compiler making different decisions about struct layout such as adding padding, when in fact that is not true. But the runtime code is much easier to read using actual C types and just have in mind that we are still sensitive to the exact layout the compiler currently uses. So after your other changes, it would make sense to change this to use `offsetof` as well.

jberdine:
To try to repro the failure I see on my laptop but not on an aws box:

1. pull down this branch (https://github.com/jberdine/skdb/tree/string)
2. checkout the `[compiler] Change string hash tag bit from 31 to 31 and 1 in generated code` commit: `git checkout string~3`
3. build _and install_ the stage1 compiler, I run: `set -a; SKARGO_PROFILE=release; STAGE=1; PATH=$PWD/compiler/stage1/bin:$PWD/compiler/stage0/bin:$PATH; set +a; make clean && make -C compiler clean install`
4. checkout the tip of this branch: `git checkout string`
5. build the stageC compiler, I run: `set -a; SKARGO_PROFILE="dev -v"; STAGE=C; PATH=$PWD/compiler/stageC/bin:$PWD/compiler/stageB/bin:$PWD/compiler/stageA/bin:$PATH; set +a; make clean && make -C compiler clean default`

Note that in the last step, I add `-v` to `SKARGO_PROFILE` in order to put `skargo` in verbose mode, which lets the error output get to the terminal.

The bad behavior I see is that when executing `skc --data stageC/bin/state.db --export-function-as main=skip_main -o stageC/bin/skc ...`, after the `clang` subprocess terminates, something like the following error message is printed:
```
bit 1 but not 31 set: 0x453f4c9c
make[1]: *** [Makefile:95: stageC/bin/skc] Error 67
```
This comes from line 43 or runtime.c.

jberdine:
This is ready now if anyone wants to make a review pass. I think/hope that the individual commits are each independeltly readable. The CI is expected to fail, the base image will need to be updated.

To test this I ran the following to test each commit in the stack one-by-one:
```
GIT_SEQUENCE_EDITOR=: git rebase -ix "dx 'set -a; SKARGO_PROFILE=dev; OLEVEL=-O0; STAGE=D; PATH=stageD/bin:\$PATH; set +a; make clean && make -C compiler clean install && make test-skfs'" main
```
where `dx` is my script to invoke a command in a docker container.
(Actually this was done only starting from the commit that adds the `stageD` build).

This process checks that a working compiler is produced at each step, when using the parent commit's compiler installed onto the host. This is true for the stack except for the step between `[runtime] Change to use bit 1 to distinguish strings` and `[compiler] Change to use bit 1 to distinguish strings in generated code` where the installed compiler generates code expecting the new runtime system, which only comes with the next commit. For this step, skc was manually installed onto the host with `set -a; SKARGO_PROFILE=release; STAGE=D; PATH=stageD/bin:$PATH; set +a; cd compiler; install -m 755 stage$STAGE/bin/skc /usr/local/bin`.

FInally, the tip was tested with a full `make test` using the tip compiler and runtime.

jberdine:
Note that the bootstrap commits, except for the last one, need not be committed. They mark points where going further in the stack requires the preceding runtime changes. I will filter them out before merging.

compiler/Makefile
```diff
@@ -62,25 +62,27 @@ stage3/lib/libstd.sklib: stage2
...
-	cp $(shell command -v skfmt) stageA/bin/skfmt
+stageD/lib/libstd.sklib: stageD/bin/skc
+	mkdir -p $(@D)
+	PATH=$$(realpath stageD/bin/):$(PATH) OUT_DIR=$$(realpath ./stageD/lib/) $(MAKE) -C ../prelude bootstrap
 
-stageB/bin/skc stageB/bin/skfmt stageB/bin/skargo: stageA
-	PATH=stageA/bin/:$(PATH) skargo build -r --bins --out-dir stageB/bin --target-dir stageB/target
-	PATH=stageA/bin/:$(PATH) skargo build -r --bins --out-dir stageB/bin --target-dir stageB/target --manifest-path ../skargo/Skargo.toml
+stageD/bin/skargo: stageD/bin/skc stageD/lib/libstd.sklib
+	PATH=stageD/bin/:$(PATH) skargo build --profile $(SKARGO_PROFILE) --bin skargo --out-dir stageD/bin --target-dir stageD/target --manifest-path ../skargo/Skargo.toml
```
mbouaziz, line 82: If you forget to clean the stageD binaries, this would use the latest stageD skargo to build the new one.

If you want to use skargo from the `PATH`, you should resolve the path to skargo before adding `stageD/bin/` to the `PATH`

jberdine: Yes, this will use a left-over stageD/bin/skargo if present. My experience building the compiler etc. in general has led me to always clean before a build. That's why my test procedure has e.g. `make clean && make -C compiler clean install && make test-skfs`. The current state of the Makefiles is, I think, that the declared dependencies suffice to determine an order in which to run the build commands from scratch, but there are many missing dependencies that would be needed to detect changes robustly. So far I have not spent the time to make non-clean builds robust. That is another project, and realistically I don't think that there is much to save since the build time is highly dominated by running skc and clang on pretty much all the files if anything has changed.

mbouaziz (approved):
LGTM :+1:

mbouaziz:
CI is red though

jberdine:
> CI is red though

Yes, CI can't handle PRs like this. It tries to build the current skdb with the old compiler and the current prelude.

I will strip out the extra bootstrap commits, land this, then rebuild the image for CI.

PR #115, opened 2024-02-15
Introduce compact FixedDir and FixedSingle representations for SKDB

bennostein:
This PR introduces much more efficient representations of the FixedDir and FixedSingle<Path, MInfo> structures for the special cases produced by SKDB, where a lot of data is shared between different key/val pairs.

The FixedDir stuff should be fairly self-explanatory, but I've written up some details about FixedSourceMap (which is a drop-in replacement for a FixedSingle<Path, MInfo>) in the commit message for sha 6931629.

Putting this up for visibility, but I'm still working on things here.  These more-compact representations look good when I inspect/dump data structures, but the top line size numbers are in fact slightly higher than before these changes.  Digging through some more components of FixedDirs/EagerDirs to see where else there is opportunity for improvement.

### DIFF
--- a/skfs/src/FixedDir.sk
+++ b/skfs/src/FixedDir.sk
@@ -140,66 +140,7 @@ class FixedDir<T: frozen> extends IFixedDir<T, FixedRow<T>> {
     row
   }
 }
-/*****************************************************************************/
-/* SQL. */
-/*****************************************************************************/
 
-
-module end;
-module SQLParser;
-
-base class Type uses Orderable {
-  children =
-  | FLOAT()
-  | INTEGER()
-  | TEXT()
-  | JSON()
-  | SCHEMA()
-}
-
-base class IKind uses Orderable {
-  children =
-  | INONE()
-  | IASC()
-  | IDESC()
-}
-
-module end;
-module SKDB;
-
-base class CValue uses Orderable {
-  children =
-  | CInt(value: Int)
-  | CFloat(value: Float)
-  | CString(value: String)
-}
-
-class RowKey(
-  row: RowValues,
-  kinds: Array<(Int, SQLParser.IKind, SQLParser.Type)>,
-) extends SKStore.Key
-
-class RowValues(
-  values: Array<?CValue>,
-  repeat: Int,
-) uses Orderable extends SKStore.File
-
-class IndexProjKey(
-  row: RowValues,
-  kinds: Array<(Int, SQLParser.IKind, SQLParser.Type)>,
-) extends SKStore.Key
-
-class ProjKey(
-  value: ?CValue,
-  colNbr: Int,
-  leftShard: Int,
-  rightShard: Int,
-) extends SKStore.Key
-
-module end;
-module SKStore;
-
-/*****************************************************************************/
 value class CompactRow<T>(
   values: Array<T>,
   repeat: Int,
--- a/skfs/src/FixedDir.sk
+++ b/skfs/src/FixedDir.sk
@@ -98,45 +98,7 @@ fun binSearch<T: Orderable>(get: Int ~> T, key: T, i: Int, j: Int): Int {
 }
 
 class FixedDir<T: frozen> extends FixedDirImpl<T, FixedRow<T>> {
-  static fun compress(x: FixedRow<T>): FixedRow<T> {
-    x
-  }
-  static fun decompress(x: FixedRow<T>): FixedRow<T> {
-    x
-  }
-}
-
-class CompressedFixedDir extends FixedDirImpl<Array<File>, FixedRow<File>> {
-  static fun compress(row: FixedRow<Array<File>>): FixedRow<File> {
-    size = row.value.i1.size();
-    invariant(size == 1);
-    FixedRow(row.key, (row.value.i0, row.value.i1[0]), row.tag);
-  }
-  static fun decompress(row: FixedRow<File>): FixedRow<Array<File>> {
-    FixedRow(row.key, (row.value.i0, Array[row.value.i1]), row.tag)
-  }
-}
-
-base class FixedDirImpl<T: frozen, Compressed: frozen> private final {
-  data: Array<Compressed> = Array[],
-} extends FixedData<T> {
-  static fun compress(FixedRow<T>): Compressed;
-  static fun decompress(Compressed): FixedRow<T>;
-
-  fun iterator(): mutable Iterator<FixedRow<T>> {
-    for (elt in this.data.iterator()) {
-      yield static::decompress(elt)
-    }
-  }
-
-  static fun isCompressable(data: mutable Vector<FixedRow<Array<File>>>): Bool {
-    for (elt in data) {
-      if (elt.value.i1.size() != 1) return false;
-    };
-    true
-  }
-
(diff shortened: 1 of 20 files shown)


gregsexton:
Is this ready for review?

bennostein:
> Is this ready for review?

Not quite yet -- still working through the issues I mentioned on thursday :(

gregsexton:
No sweat. Just checking. Feel free to ping the PR when ready and I'll make time to look over this. :)

bennostein:
Some interesting numbers here, @pikatchu , following up on our conversation earlier:

I've got three different test inputs now: (A) an inputs-only version (w/ all primary key constraints and reactive views removed), (B) an inputs/indexes-only version (w/ primary key constraints but no reactive views), and (C) the full dump.

I've tested the current `origin/main` and the tip of this stack on all three.  For inputs A and B, this stack reduces db size by ~19.5%.  For input C, it actually _increases_ db size by ~4%.

There's one broken test I'm chasing down (a diff test with a "group by" reactive view) but I don't expect a fix for that to change these results massively.

Full details as follows:
```
[ 15:20 docker:sk-dev /skdb ] cat ./test.sh
make build/skdb

rm -f test.db
build/skdb --init test.db

cat $1 | build/skdb --data test.db

echo Initial size:
build/skdb --data test.db size

build/skdb --data test.db compact

echo Compacted size:
build/skdb --data test.db size

[ 15:21 docker:sk-dev /skdb ] g co main
Already on 'main'
Your branch is up to date with 'origin/main'.

[ 15:21 docker:sk-dev /skdb ] ./test.sh overtone-dump.sql
cd sql && skargo build --release
mkdir -p build
cp sql/target/host/release/skdb build/skdb
Initial size:
116847344
Compacted size:
106550736

[ 15:22 docker:sk-dev /skdb ] ./test.sh overtone-dump-inputs-only.sql
make: 'build/skdb' is up to date.
Initial size:
15499696
Compacted size:
14122896

[ 15:23 docker:sk-dev /skdb ] ./test.sh overtone-dump-inputs-and-indices.sql
make: 'build/skdb' is up to date.
Initial size:
27206544
Compacted size:
16916304

[ 15:23 docker:sk-dev /skdb ] g co optimize_space
Switched to branch 'optimize_space'

[ 15:23 docker:sk-dev /skdb ] ./test.sh overtone-dump.sql
cd sql && skargo build --release
mkdir -p build
cp sql/target/host/release/skdb build/skdb
Initial size:
125412464
Compacted size:
110919728

[ 15:24 docker:sk-dev /skdb ] ./test.sh overtone-dump-inputs-only.sql
make: 'build/skdb' is up to date.
Initial size:
13923248
Compacted size:
11375888

[ 15:25 docker:sk-dev /skdb ] ./test.sh overtone-dump-inputs-and-indices.sql
make: 'build/skdb' is up to date.
Initial size:
28491376
Compacted size:
13627024
```

bennostein:
Marking this as ready for review!  Re-ran measurements and the topline is the same as my previous comment: ~20% improvement for input directories and/or indices, ~3-4% more space used for the join/view-dominated dump.

It's big but hopefully the commit history helps break things down.  The commit messages have more context on their changes where that's warranted, and I'm happy to chat with anyone on here or vc if that makes reviewing easier.

gregsexton:
Bear with me, I'll make time to look this over. Great wins! Out of curiosity, what was the cause of the issue that you were battling?

bennostein:
It was a bug in the new logic I wrote for `Extension`s here, which was only surfacing for those aggregation diff tests.  So not an underlying issue with sums etc., just a silly mistake that led me to bark up that wrong tree.

skfs/src/FixedDir.sk
```diff
@@ -20,19 +20,9 @@
 /*****************************************************************************/
 module SKStore;
 
-value class Source(dirName: DirName, baseName: Key) uses Orderable {
```
gregsexton, line 23: Glorious. This 'type conversion' has always been really annoying.

skfs/src/SKDBTypes.sk
```diff
@@ -0,0 +1,57 @@
+/*****************************************************************************/
+/* Some SQL types are exposed here to enable specializing SKFS operations to */
```
gregsexton, line 2: Perf doesn't come cheap, eh? Shame we had to do this, but it seems reasonable to specialise to get space savings. Have we tested skfs outside of skdb to find any bad assumptions? Maybe using skui? cc @skiplabsdaniel

skfs/src/FixedDir.sk
```diff
@@ -111,26 +122,155 @@ class FixedDir<T: frozen> private {
...
+    static{data => data.toArray()}
+  }
+
+  protected fun decode(row: FixedRow<T>): FixedRow<T> {
+    row
+  }
+}
+
+value class CompactRow<T>(
+  values: Array<T>,
```
gregsexton, line 145: [for another diff(s)] With understanding of the schema and some table stats (which could be collected/updated during purge), we should be able to do a lot better than this. e.g. 

- we could do some simple flat representation encoding to pack values together in to a binary blob.
- we could track most common values and reference a dictionary (very good for highly repeated strings used as an enum, which is very common)
- bool values (also very common, either as {0,1} int or 2 strings) could be a bit 
- nulls could be a small bitmask (most tables have few cols), and the values not represented at all
- etc, etc

cc @mbouaziz - I'm sure you'll think of this stuff and more.

skfs/src/FixedDir.sk
```diff
@@ -111,26 +122,155 @@ class FixedDir<T: frozen> private {
...
+  repeat: Int,
```
gregsexton, line 146: [for another diff?] In practice it's very rare for this to be anything other than 1, I believe. It can happen that this is larger than 1 if values are inserted atomically and replication sends them across - this is the only case I'm aware of. I'm not aware of anywhere where this can be < 1. In fact there might even be an invariant somewhere.

Then from a user perspective, when are you ever inserting duplicate rows in a transaction? I honestly can't manufacture even the most artificial scenario.

Given this, might make sense to assume a repeat of 1, save ourselves 4 bytes per row, and just pay an up front check that there are no repeats > 1. Or we might even just completely remove repeats. Replication can send a count still but we could expand this using different sources (possibly, would need to check logic here carefully).

gregsexton (commented):
Great work! Looks really solid.

This was a roller coaster to review. :D I went through the commits and then deleted basically all of my comments as you cleaned things up really nicely.

I didn't spot any issues, but given we're messing with and reconstructing sources, it would be a very very good idea to run the soak and the replication simulator. Replication makes heavy use of sources for achieving merge resolution and avoiding echoing your own changes back and creating feedback loops.

skfs/src/FixedDir.sk
bennostein: Instrumented things and ran `make test-native` to see how often we get non-1 repeats, and it actually happens quite a lot.  Can dig into things further in a follow up, but think we're better off continuing to pay this 4 bytes per row for now.

bennostein:
Soak and replication tests seem happy so I'm going to merge this.  Will keep an eye out but let me know if you run into anything that seems dodgy and potentially related!

PR #116, opened 2024-02-15
Several fixes to skdb js package

gregsexton:
Several of these make the recently added type constraints more
appropriate. The rest I've discovered the hard way trying to produce a
bundled web app.


gregsexton:
I need to quickly merge this so I can release a package and unblock myself. Will address any comments in follow ups.

bennostein:
Looks good to me -- ship it!

PR #117, opened 2024-02-16
Make SKDBTable more idiomatic and ergonomic

gregsexton:
We throw less and use undefined more - this makes it more idiomatic
and generally easier to use. E.g. `rows.scalarValue() ?? 0;`.
Typescript does generally do a pretty good job of making you 'unwrap
the optional'.

I also added a couple of methods I kept reaching for to get the first
and last row without throwing.

### DIFF
--- a/sql/ts/src/skdb_util.ts
+++ b/sql/ts/src/skdb_util.ts
@@ -35,14 +35,20 @@ export class ExternalFuns {
 }
 
 /* ***************************************************************************/
-/* Class for query results, extending Array<Object> with some common selectors
-   and utility functions for ease of use. */
+/* Class for query results, extending Array<Record<>> with some common
+   selectors and utility functions for ease of use. */
 /* ***************************************************************************/
 export class SKDBTable extends Array<Record<string, any>> {
-  scalarValue(): any {
-    const row = this.onlyRow();
+  scalarValue(): any | undefined {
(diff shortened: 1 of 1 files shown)


gregsexton:
cc @bennostein

bennostein:
LGTM but in a this-hurts-my-soul kind of way

gregsexton:
If you had an Option/Maybe type would you have returned that or would you still have chosen to throw?

bennostein:
I probably would have used an option!

I like the ocaml convention of using e.g. `foo` for the option-returning version and `foo_exn` for the throwing version of the same function.

gregsexton:
I try to avoid exceptions (in general but definitely) for anything unexceptional, (they cause bugs when not considered - often breaking invariants, it forces you to use exn catching for control flow which is usually orders of mag slower, can't be easily passed across thread boundaries, causes UB when threads crash and breaks invariants, and on and on). So I generally try to use option wherever possible. You can always do whatever the idiomatic equivalent of `x = getSomethingReturningOption().getOrThrow()` if you really want an exn.

Sorry, this comment started off harmless and after a few edits turned in to a rant. :D

gregsexton:
CI has a failure, but tests pass locally. Looks like flakiness to me so going to ship.

PR #122, opened 2024-02-19
Implementation of insertMany in JS

pikatchu:
It's the same as insert, but for more than one value. So we had:

skdb.insert([1, 2, 3]);

Now we can also write:

skdb.insertMany([[1, 2, 3], [4, 5, 6]]);

### DIFF
--- a/sql/ts/src/skdb_database.ts
+++ b/sql/ts/src/skdb_database.ts
@@ -280,6 +280,33 @@ export class SKDBSyncImpl implements SKDBSync {
     return this.runLocal(args1, stdin1) == "";
   };
 
+  insertMany = (tableName: string, valuesArray: Array<Array<any>>) => {
+    let params = new Map();
+    let valueIndex = 0;
+    while(valueIndex < valuesArray.length) {
+      let buffer = new Array();
+      buffer.push("insert into " + tableName + " values ");
+      for(let roundIdx = 0; roundIdx < 1000; roundIdx++) {
+        if(valueIndex >= valuesArray.length) break;
+        let values = valuesArray[valueIndex];
+        let keys = values.map((val, i) => {
+          let key = "@key" + i;
+          params.set(key, val);
+          return key;
+        });
+        if(roundIdx != 0) buffer.push(",");
+        buffer.push("(" + keys.join(", ") + ")");
+        valueIndex++;
+      }
+      buffer.push(";");
+      let [args1, stdin1] = this.addParams([], params, buffer.join(''));
+      if(this.runLocal(args1, stdin1) != "") {
+        return false;
+      }
+    };
+    return true;
+  }
+
   assertCanBeMirrored(tableName: string, schema: string): void {
     const error = this.runLocal(["can-mirror", tableName, schema], "");
     if (error === "") {
@@ -400,6 +427,10 @@ export class SKDBImpl implements SKDB {
 
   async insert(tableName: string, values: Array<any>) {
     return this.skdbSync.insert(tableName, values);
+  }
+
+  async insertMany(tableName: string, valuesArray: Array<Array<any>>) {
+    return this.skdbSync.insertMany(tableName, valuesArray);
   }
 
   async mirror(...tables: MirrorDefn[]) {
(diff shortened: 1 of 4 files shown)


sql/ts/src/skdb_types.ts
```diff
@@ -39,6 +39,7 @@ export interface SKDBSync {
     update: (added: SKDBTable, removed: SKDBTable) => void,
   ) => { close: () => void };
   insert: (tableName: string, values: Array<any>) => boolean;
+  insertMany: (tableName: string, valuesArray: Array<Array<any>>) => boolean;
```
gregsexton: You need to expose this in the SKDB type if you want it in the public interface. 

Insert was removed from the public facing interface because it wasn't really paying for itself now that we have params. So you might want to rename this 'insert' and kill the old one?

gregsexton (commented):
Could we change this to take an array of objects? Or maybe even an SKDBTable object (which is an array of objects)?

This would make what you get from us and what you give to us symmetric. But mostly it will be much easier to work with in JS. You almost never work with an array of arrays like this, it's not idiomatic in js libraries, but an array of objects is super common. So I expect in most situations users will have to write conversion logic before being able to use this.

Taking an array of objects should allow you to specify the col list too in the insert statement by looking at the keys.

sql/ts/src/skdb_database.ts
```diff
@@ -280,6 +280,33 @@ export class SKDBSyncImpl implements SKDBSync {
...
+    let valueIndex = 0;
+    while(valueIndex < valuesArray.length) {
+      let buffer = new Array();
+      buffer.push("insert into " + tableName + " values ");
+      for(let roundIdx = 0; roundIdx < 1000; roundIdx++) {
+        if(valueIndex >= valuesArray.length) break;
+        let values = valuesArray[valueIndex];
+        let keys = values.map((val, i) => {
+          let key = "@key" + i;
+          params.set(key, val);
```
bennostein, line 330: Why indirect through a query parameter here instead of interpolating the value directly into the query string?

I'm also not sure how/whether this works -- won't each `params.set` call overwrite the previous row's in the `params` map, such that an insertMany of n different rows instead inserts n copies of the last row?

gregsexton: > Why indirect through a query parameter here instead of interpolating the value directly into the query string?

You would need to carefully escape the values - very dangerous. Much better this way so the parser/compiler never interprets these as part of the query.

bennostein: Fair, better to have that all the escaping logic in one place.  I haven't tested performance with big query parameter objects, though -- are we confident that everything will be happy with a query with ~5-10k params? (this will generate params objects of size up to 1000 * number-of-columns)

bennostein: Once the issue in my initial comment is fixed, that is.  As is, the param object is just the last row.

gregsexton: Agree with Benno. Logic looks broken. The test is only checking the count rather than what is written.

pikatchu: Yes you are right. My bad. I should have tested this more carefully.

Thanks for the thorough review!

pikatchu:
So I took into account your remarks bar one. I don't think calling "insertMany" "insert" is a good idea. I think it's confusing. Happy to further discuss if you feel strongly about it @gregsexton.

Thanks for taking the time to have a look at this! @bennostein @gregsexton

sql/ts/src/skdb_types.ts
```diff
@@ -80,6 +80,11 @@ export interface SKDB {
     update: (added: SKDBTable, removed: SKDBTable) => void,
   ) => Promise<{ close: () => Promise<void> }>;
 
+  insertMany: (
+    tableName: string,
+    valuesArray: Array<Array<any>>
```
gregsexton, line 85: Array<Record<string, any>>?

gregsexton:
> if you feel strongly about it

Nah, I don't. I felt more strongly about taking in a sequence of objects. Thanks for doing that. :)

sql/ts/src/skdb_types.ts
pikatchu: Yes good call thanks.

PR #123, opened 2024-02-19
Change CSV string en/decoding

gregsexton:
If we insert a string with a newline replication becomes permanently
broken for the client. This is because there is a single w-csv process
per client and the parsing state machine becomes broken. We're looking
for tokens that won't arrive instead of seeing the checkpoint that
indicates end of txn.

The chosen solution is to properly escape and encode strings.

## The encoding itself

The encoding deviates from 'standard' csv -- if there is such a thing.
Instead we use the json escaping algorithm. This is mostly for 2
reasons:

1. this approach is very well battle tested and accounts for more
things than csv usually considers.

2. we have logic that assumes that a row is represented on a single
line (e.g. looking for checkpoint markers, the csv parser itself!).
Standard CSV allows for line breaks in strings and this wreaks havoc.
It'll also hurt line-based buffering. This is all ok as the csv we
produce is not meant to be public facing. And ultimately we plan to
replace this with a binary format.

And it helped that we already have the json en/decoding code written
and tested. I just adapted it a little for the expected interface.

### DIFF
--- a/sql/src/SqlCAst.sk
+++ b/sql/src/SqlCAst.sk
@@ -221,6 +221,7 @@ base class CValue uses Orderable, Show {
     this match {
     | CInt(n) -> n.toString()
     | CFloat(f) -> floatToString(f)
+    | CString(str) if (format is SKStore.OCSV _) -> SKCSV.escapeString(str)
     | CString(str) if (format.usesDoubleQuotes()) ->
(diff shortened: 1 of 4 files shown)


gregsexton:
This fixes critical bugs. Merging.

PR #125, opened 2024-02-20
Get skdb package to work with node, bundler server, and bundler optimised build

gregsexton:
Previously we had the first two, but the optimised build didn't work.

When building for production, the bundler statically analyses the code
to find the dep. tree. It then concatenates, minifies, tree shakes,
etc the js files to produce the smallest set of files possible and
reduce http round trips.

We were being too clever and the bundler couldn't see some of our
module loading so files were not being included. It also broke the
worker and wasm loading.

I've tested the cross product of:

- running skdb on the main thread
- running skdb on a worker thread

- in a node setup
- in a vite dev setup - the clever hot module reloading server thing
- in a vite prod setup - doing a vite build and then serving the files statically

Webpack is another very popular bundler. I looked at it's docs when
considering solutions and it seems to be compatible with this, but I
haven't tested.

### DIFF
--- a/packages/react/src/skdb-react.tsx
+++ b/packages/react/src/skdb-react.tsx
@@ -1,6 +1,7 @@
 import { useState, useEffect } from "react";
 import { createContext, useContext } from "react";
 import type { SKDB } from "skdb";
+import { SKDBTable } from "skdb";
 import { Params } from "skdb/dist/skdb_types.js";
 
 export const SKDBContext = createContext<SKDB | undefined>(undefined);
@@ -29,9 +30,9 @@ export function useQuery(
   query: string,
   params: Params = {},
   defaultRows: Array<any> = [],
-): any {
+): SKDBTable {
   const skdb = useSKDB();
-  const [state, setState] = useState(defaultRows);
+  const [state, setState] = useState<SKDBTable>(new SKDBTable(...defaultRows));
 
   const deps = Object.values(params);
   deps.push(skdb);
@@ -41,7 +42,7 @@ export function useQuery(
     let removeQuery = false;
     const closeable = { close: () => {} };
     skdb
-      .watch(query, params, (rows: Array<any>) => {
+      .watch(query, params, (rows) => {
         setState(rows);
       })
       .then((handle) => {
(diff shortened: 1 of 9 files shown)


gregsexton:
This PR nearly killed me. Definitely going greyer earlier because of it. :/

But I think it finally achieves a holy grail. I swear we must be the only library in the world trying to target browser and node, using wasm, optionally running in a worker.

gregsexton:
One other thing. To test I created a minimal vite+react+skdb package. And the same for node. These could evolve in to templates to quickstart with skdb.

gregsexton:
cc @skiplabsdaniel

gregsexton:
Oh, and last thing, for real this time. Some of the solution here is super dumb. But honestly, I tried so many things. This was really the least dumb that I could make work.

skiplabsdaniel:
I have a branch with similar changes.
With import for wasm, it's no longer possible to use the package as is, it's requires the use of a bundler.

gregsexton:
> I have a branch with similar changes.

Oh sorry, I didn't realise you were working on this. Anything worth merging together? I imagine you might have some better ideas on this stuff. :)

> With import for wasm, it's no longer possible to use the package as is, it's requires the use of a bundler.

You can use as is with node. But otherwise yeah you need a bundler. Do you mean using a <script> tag in the browser? Or something else? Let me know and maybe I can figure out a way of checking the environment and getting this to work.

gregsexton:
Ah, I see the tests are very unhappy. I'll fix these and maybe that fixes your concern too.

skiplabsdaniel:
> Oh sorry, I didn't realise you were working on this..

You couldn't have known. As your branch solves some issues, I'll wait for it to be merged.
I'm thinking of adding an option to sknpm to run vite or/and webpack, so I can check that I'm not breaking anything and that it works on my other projects.

> Ah, I see the tests are very unhappy. I'll fix these and maybe that fixes your concern too.

Yes, fixing the tests fixes my concern too.

gregsexton:
> I'm thinking of adding an option to sknpm to run vite or/and webpack, so I can check that I'm not breaking anything and that it works on my other projects.

This sounds like a good idea. I was actually curious about what if we bundle everything as part of the build so that anything downstream is working with something pre-bundled. This way we wouldn't have to worry about different bundlers and making sure we work with everything. Not sure how this would work with wasm though.

gregsexton:
Latest commit should fix everything. It's regretfully quite hacky, so I'd be very open to other approaches. But it's the best I could come up with.

skiplabsdaniel:
LGTM.
Great job.

PR #126, opened 2024-02-21
[skstore] Allow function call on parent obstack

### LINKED ISSUE #106: Segfault in withRegionFold under certain conditions
Not using line numbers because they might change over time.

Inside Context.sk, inside the function withRegionFold, the "next" call on the iterator segfaults. It only segfaults on one specific code path, the one coming from EagerDir.sk, inside the function "update", the last call to withRegionFold.

I added a temporary fix that consists in rewriting the iterator from:
it = dirty.iterator()

To:
it = dirty.iterator().collect(Array).iterator()

But we should investigate why this segfaults without it.

Under ARM M3 (apple)

In the following version of ubuntu:
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy

With the following version of clang:
Ubuntu clang version 15.0.7
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

And the following version of llvm:
Ubuntu LLVM version 15.0.7
  Optimized build.
  Default target: aarch64-unknown-linux-gnu
  Host CPU: (unknown)

skiplabsdaniel:
Add a way to call a function in the parent Obstack of given one, to allow mutable data to cross withRegion.

Fix #106

### DIFF
--- a/compiler/runtime/obstack.c
+++ b/compiler/runtime/obstack.c
@@ -253,30 +253,44 @@ void* SKIP_destroy_Obstack_with_value(sk_saved_obstack_t* saved, void* toCopy) {
   return result;
 }
 
-void SKIP_switch_to_parent(sk_saved_obstack_t* saved) {
+sk_obstack_t* SKIP_switch_to_parent(sk_saved_obstack_t* saved) {
+  // Gather current obstack data
+  sk_obstack_t* first = page;
+  while (first != NULL && first->previous != saved->page) {
+    first = first->previous;
+  }
+  
   sk_obstack_t* saved_page = page;
   char* saved_head = head;
   char* saved_end = end;
 
+  // Switch to parent obstack
   page = saved->page;
   head = saved->head;
   end = saved->end;
 
+  // Store obstack dat for restauration
   saved->head = saved_head;
-  saved->page = saved_page;
+  saved->page = first;
   saved->end = saved_end;
+
+  return saved_page;
 }
 
-void SKIP_restore_from_parent(sk_saved_obstack_t* saved) {
-  sk_obstack_t* saved_page = saved->page;
+void SKIP_restore_from_parent(sk_saved_obstack_t* saved, sk_obstack_t* leading) {
+  // Save the obstack restauration data
+  sk_obstack_t* first_page = saved->page;
   char* saved_head = saved->head;
   char* saved_end = saved->end;
 
+  // Update the sub obstack with new parent obstack positions
+  first_page->previous = page;
   saved->page = page;
   saved->head = head;
   saved->end = end;
 
-  page = saved_page;
+  // Switch to sub obstack
+  page = leading;
   head = saved_head;
   end = saved_end;
 }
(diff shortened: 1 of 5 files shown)


compiler/runtime/obstack.c
```diff
@@ -248,6 +248,34 @@ void* SKIP_destroy_Obstack_with_value(sk_saved_obstack_t* saved, void* toCopy) {
 
   return result;
 }
+
```
gregsexton, line 262: [super nit]: you've got some trailing whitespace here.

skfs/src/Context.sk
```diff
@@ -1691,6 +1702,14 @@ fun withRegionValue<T>(f: () ~> T): T {
   }
 }
 
+@debug
+fun callOnParentObstack<T>(saved: Obstack, f: () -> T): T {
+  switchToParent(saved);
+  res = f();
```
gregsexton, line 1716: I guess we get in to all sorts of trouble here if f creates or destroys an obstack? Probably not an easy way of guarding against this. But maybe a big warning comment couldn't hurt?

skfs/src/Context.sk
```diff
@@ -1718,7 +1737,7 @@ fun withRegionFold<K, T>(
     !optSaved = Some(newObstack());
     loop {
       saved = optSaved.fromSome();
-      it.next() match {
+      callOnParentObstack(saved, it.next) match {
```
gregsexton, line 1748: If I remember the data structure correctly, an obstack is a linked list of pages, so this shouldn't be an issue if it.next allocates causing a new page to be created? It won't corrupt the child obstack? Pretty sure this is true in native where we just malloc pages, but wasm has a simpler allocation scheme I think that's based on bumping, could that cause a problem?

gregsexton (commented):
Ah ha! I certainly never considered the iterator! Nicely done.

skfs/src/Context.sk
gregsexton: I'm trying to remember the wasm implementation. I don't think this is an issue. It only ever increases the bump right?

skiplabsdaniel: Yes, It increases the bump or takes a previously freed page in the free table so It don't affect the child obstack.

jberdine:
Stepping back from this particular code change, is there a concise and general characterization of what was wrong with the old withRegionFold implementation? I understand how it leads to creating "old to young" pointers, and that the runtime system is not prepared to handle them. But is this potentially dangerous code pattern specific to withRegion, or is it something that we need to be on the watch for all over the code base? Maybe it is just "if you ever call newObstack, be extremely careful"?

gregsexton:
My mental model is: don't pass anything mutable (and therefore could allocate and update internal references) across a new obstack barrier. You're in for a world of pain.

jberdine:
Ok that makes sense. This makes me think that we should try hard to get rid of all the calls to newObstack other than in the implementation of withRegion*, and if we can't in some cases, create whatever variant of withRegion is needed. (I don't know of other ways in the language than accepting an immutable callback to enforce not passing mutable things across newObstack.)

gregsexton:
I didn't realise this was still open. You should ship! But there is a CI failure, is this related to the diff?

mbouaziz:
Waiting on @pikatchu to test on his machine

gregsexton:
Let's ship this?

PR #129, opened 2024-02-22
[stdlib] Fix `Time.time()`/`Time.time_ms()`.

beauby:
These runtime functions not being marked `@debug` meant they were eligible to CSE, causing the following code to always return zero:
```
t0 = Time.time_ms();
f();
t1 = Time.time_ms();

t1 - t0
```

### DIFF
--- a/prelude/src/stdlib/other/Time.sk
+++ b/prelude/src/stdlib/other/Time.sk
@@ -1,10 +1,12 @@
 module Time;
 
+@debug
 @cpp_extern("SKIP_time")
 native fun time(): Int;
 
 // Current time expressed as the number of milliseconds since
 // 1970-01-01T00:00:00.000Z, the beginning of the Unix epoch.
+@debug
 @cpp_extern("SKIP_time_ms")
 native fun time_ms(): Int;
 


mbouaziz:
IIUC `@debug` means _Please don't eliminate this because I use it for debugging_.
Here, the goal is not debugging. Should there be another annotation, e.g. `@volatile`?

Also I would expect `@debug` function calls to be eliminated in a release build.
`@volatile` can easily be used by any kind of purity analysis.

jberdine:
Yes, this is a misuse of `@debug` that is currently a work-around. Ideally `native` functions would be treated specially, see https://github.com/SkipLabs/skfs/issues/675.

ltique:
It looks like a compiler barrier (i.e. "compiler, you are not allowed to move things around this barrier") is what is ideally needed? If the problem is that LLVM is moving stuff, you can use asm black boxes as a hack that in practise act as a compiler barrier of sorts.

jberdine:
Expanding a bit, the current interpretation of `immutable` types and functions is that they are observationally pure, and so they should satisfy transformations such as CSE. The system uses `immutable` as an indicator that values and computations are safe to not track in the incremental computation graph. Some `native` functions expose an `immutable` type but are in fact not pure. This use of `native` (or, similarly, `debug`) achieves an intermediate point between `mutable` (which must be tracked by the incremental system) and `immutable` (that isn't). This intermediate point is essentially "reads/writes mutable state, but changes to that state should not be considered to require any incremental recomputations". The best way to support that usage in the (type) system is not currently clear. For now, the approach is to use `debug native` functions with `immutable` types.

jberdine (approved):

jberdine:
And the point of making `native` functions ineligible for CSE by default, as if they were tagged `debug`, is that the common case for native functions is that they are in fact impure. It would be safer to treat all of them as impure by default, and have an annotation that could be consciously added to select native functions that we want the compiler to treat as pure and perform e.g. CSE on them.

PR #130, opened 2024-02-22
Introduce `semver` package.

beauby:
This package implements various structures and helpers around semantic
versioning (operators, comparisons, matching, etc.).


semver/src/parse.sk
```diff
@@ -0,0 +1,190 @@
...
+  loop {
+    pos.peek() match {
+    | '.' ->
+      pos.advance();
+      res.push(identifier_segment(pos))
+    | _ -> break void
+    }
+  };
+
+  if (res.isEmpty()) {
```
mbouaziz: This cannot happen since you start with one element

semver/src/lib.sk
```diff
@@ -0,0 +1,267 @@
...
+  | Greater()
+  | GreaterEq()
+  | Less()
+  | LessEq()
+  | Tilde()
+  | Caret()
+  | Wildcard()
+}
+
+class Prerelease(identifier: String) uses Show, Orderable {
```
mbouaziz, line 169: Would it be simpler to keep an array of `Numeric | NonNumeric`?

semver/src/lib.sk
```diff
@@ -0,0 +1,267 @@
...
+  // Cf. https://semver.org/:
+  // > Build metadata MUST be ignored when determining version precedence.
+  fun compare_precedence(other: Version): Order {
+    (this.major, this.minor, this.patch, this.pre).compare(
+      (other.major, other.minor, other.patch, other.pre),
+    )
+  }
+}
+
+class VersionReq(comparators: Array<Comparator>) uses Show {
```
mbouaziz, line 43: Is there a spec for this?

skargo/src/manifest.sk
```diff
@@ -1,43 +1,28 @@
...
-  static fun fromTOML(basePath: String, toml: TOML.Value): VersionConstraint {
+// TODO: Add support for `git` repositories as a package location.
+class ManifestDependency{
+  version: ?Semver.VersionReq = None(),
+  path: ?String = None(),
+} {
+  static fun fromTOML(basePath: String, toml: TOML.Value): ManifestDependency {
     toml match {
-    | TOML.String(v) -> static::fromString(v)
+    | TOML.String(v) ->
```
mbouaziz, line 10: Don't you want to allow `{ version = ... }` as well?

skargo/src/manifest.sk
```diff
@@ -1,43 +1,28 @@
...
-          Path.join(basePath, path),
-          t.maybeGetString("version").map(Version::fromString),
-        )
+        ManifestDependency{
+          // FIXME: This breaks on absolute paths.
+          path => Some(Path.join(basePath, path)),
+          version => t
+            .maybeGetString("version")
+            .map(Semver.VersionReq::fromString),
+        }
```
mbouaziz, line 21: Fail if `t` has non-`path`, non-`version` fields?

semver/src/lib.sk
beauby: That's an option, although there is a tradeoff between processing the pre-release tag upfront (possibly for a large number of elements, for instance when parsing a package repository index), and doing it dynamically (and each time, though it could be cached) during comparison. Since in the vast majority of package version comparisons the pre-release tag won't even come into play, I opted to not pay the cost upfront (not that perfs matter all that much here at the moment).

semver/src/lib.sk
beauby: It is (heavily) inspired from [cargo](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html), but no official spec that I know of.

semver/src/parse.sk
beauby: Good catch.

skargo/src/manifest.sk
beauby: Yes, the current grammar for `Skargo.toml` is a bit limited.

skargo/src/manifest.sk
beauby: Yes we should somehow validate the schema of the whole file.

PR #133, opened 2024-02-22
Fix: the context used by lwwEdits should be mutable

gregsexton:
The goal of trying to keep the context readonly here was laudable but
/indexbycol/ is created lazily. By cloning and creating the dir in
the clone we end up permenantly marking /indexbycol/ as a DeletedDir.
This happens because we call update immediatley after in the w-csv hot
loop.

The overall effect is that: if replication is the first to get to
creating /indexbycol/ then it becomes permanently deleted for all
users afterward. This means that after the initial replication all
future replication is broken, and likely other things too.

### DIFF
--- a/sql/src/SqlCsv.sk
+++ b/sql/src/SqlCsv.sk
@@ -210,23 +210,22 @@ fun lwwEdits(
   snapshot: Int,
   writer: SKStore.Path,
   primaryIdxOpt: ?Int,
-  context: readonly SKStore.Context,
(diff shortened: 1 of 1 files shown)


gregsexton:
This has been like this for some time. No idea how we got through all the tests and dogfooding without hitting it. But in the app I'm working on now, replication is just immediately permanently broken.

gregsexton:
Can't see how anyone would object to this. Merging quickly as it's such a serious bug.

PR #135, opened 2024-02-27
[runtime] Type for obstack

mbouaziz:
This is a cosmetic PR that should just make `obstack.c` easier to read.

Tested with:
```shell
set -a; SKARGO_PROFILE=dev; STAGE=C; PATH=$PWD/compiler/stage$STAGE/bin:$PATH; set +a; make clean && make -C compiler clean test && make test
```

### DIFF
--- a/compiler/runtime/obstack.c
+++ b/compiler/runtime/obstack.c
@@ -267,7 +267,7 @@ uint32_t SKIP_should_GC(sk_saved_obstack_t* saved) {
     cursor = cursor->previous;
     nbr_page++;
     if (nbr_page > 3 ||
-        (nbr_page > 1 && (head - (char*)page > (2 * PAGE_SIZE / 3)))) {
+        (nbr_page > 1 && (head - page->user_data > (2 * PAGE_SIZE / 3)))) {
       return 1;
     }
   }
--- a/compiler/runtime/obstack.c
+++ b/compiler/runtime/obstack.c
@@ -38,7 +38,7 @@
 // end: The end of the page.
 
 #ifdef SKIP32
-char* page = NULL;
+struct sk_obstack* page = NULL;
 char* head = NULL;
 char* end = NULL;
 #endif
@@ -48,7 +48,7 @@ char* end = NULL;
 // well).
 
 #ifdef SKIP64
-__thread char* page = NULL;
+__thread struct sk_obstack* page = NULL;
 __thread char* head = NULL;
 __thread char* end = NULL;
 #endif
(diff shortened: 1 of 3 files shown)


compiler/runtime/obstack.c
```diff
@@ -340,11 +340,11 @@ size_t sk_get_nbr_pages(void* saved_page) {
 sk_cell_t* sk_get_pages(size_t size) {
   sk_cell_t* result = (sk_cell_t*)sk_malloc(sizeof(sk_cell_t) * size);
   int i = 0;
-  void* cursor = page;
+  sk_obstack_t* cursor = page;
   for (i = 0; i < size; i++) {
     result[i].key = cursor;
-    result[i].value = (uint64_t)(cursor + *(size_t*)(cursor + sizeof(char*)));
-    cursor = *(char**)cursor;
+    result[i].value = (uint64_t)(cursor + cursor->size);
```
jberdine: I'm not absolutely positive, but I think this line needs careful checking. My understanding is that the previous code that has `cursor + ...` where `cursor` has type `void*` is illegal, but likely to be treated by compilers as if it was `char*` instead of `void*`. I expect that the new code will add `cursor->size * sizeof(sk_obstack_t)` bytes while the old added only `cursor->size` bytes. I'm tempted to read the llvm asm that clang produces for this function.

compiler/runtime/obstack.c
```diff
@@ -340,11 +340,11 @@ size_t sk_get_nbr_pages(void* saved_page) {
 sk_cell_t* sk_get_pages(size_t size) {
```
jberdine: While you are here, it would probably be easier to read if the `size` argument was renamed to `nbr_pages`.

compiler/runtime/obstack.c
```diff
@@ -59,63 +59,63 @@ __thread char* end = NULL;
...
-  char* page;
+  struct sk_obstack* page;
   char* end;
 } sk_saved_obstack_t;
 
+typedef struct sk_obstack {
+  struct sk_obstack* previous;
+  size_t size;
+  sk_saved_obstack_t saved;
+  char contents[0];
```
jberdine: nit: the comments at the top of obstack.c use the term "user data" for this, maybe change either here or there to use the same term in both places.

compiler/runtime/obstack.c
```diff
@@ -197,19 +196,19 @@ void SKIP_destroy_Obstack(sk_saved_obstack_t* saved) {
     saved_head = saved->head;
     saved_end = saved->end;
   }
-  while (saved_head < page || saved_head > end) {
-    char* tofree = page;
+  while (saved_head < (char*)page || saved_head > end) {
```
jberdine: Not new in this diff, but my understanding is that the comparisons between the various `head`s and `page` would be more accurately done with `page->contents` instead, since `head` should never be able to point into the part of a page before the start of the contents. Also, I think the while condition is trying to test saved_head ∉ [page,end) and `saved_head == end` can't end up happening only because `head` can't point to the beginning of a page that happens to get allocated adjacent. So I think checking `saved_head >= end` would be clearer, or even rewritten as `!(page <= saved_head && saved_head <= end)`.

compiler/runtime/obstack.c
```diff
@@ -264,11 +263,12 @@ uint32_t SKIP_should_GC(sk_saved_obstack_t* saved) {
...
   size_t size = 0;
-  void* cursor = page;
+  sk_obstack_t* cursor = page;
   while (cursor != NULL && cursor != saved->page) {
-    cursor = *(char**)cursor;
+    cursor = cursor->previous;
     nbr_page++;
-    if (nbr_page > 3 || (nbr_page > 1 && (head - page > (2 * PAGE_SIZE / 3)))) {
+    if (nbr_page > 3 ||
+        (nbr_page > 1 && (head - (char*)page > (2 * PAGE_SIZE / 3)))) {
```
jberdine: As above, `(char*)page` here might be better as `page->contents`.

jberdine (commented):
Nice, this helps! I made a few comments, and there is one thing that I think needs more checking.

It would be good to comment on how this has been tested. For runtime changes I run `set -a; SKARGO_PROFILE=dev; STAGE=C; PATH=$PWD/compiler/stage$STAGE/bin:$PATH; set +a; make clean && cd compiler && make clean test && cd .. && make test`. Or possible with `STAGE=2` instead of `C` for changes that should not introduce any incompatibilities between the runtime system before and after the diff.

compiler/runtime/obstack.c
mbouaziz: The condition may be always false for small values of `PAGE_SIZE` but we know it won't happen...

It's just a heuristics so fine changing it.

compiler/runtime/obstack.c
mbouaziz: I agree

compiler/runtime/obstack.c
mbouaziz: I'll change to `user_data` to avoid introducing a new term

compiler/runtime/obstack.c
mbouaziz: I'll do that

compiler/runtime/obstack.c
gregsexton: Good spot! Godbolt suggests you're right: https://godbolt.org/z/v1o4M6YT4 gcc just adds bytes to cursor as if it were char*. Clang fails to compile on godbolt, which is a puzzle - don't we use clang for building the runtime??

jberdine: Clang will compile it as C but not C++... See https://godbolt.org/z/7E1oKh1Mr for a version with void** compared to void*. The asm and bitcode is different.

mbouaziz: Good catch! I went on fixing these kinds of things in #138 first

mbouaziz: Turned to `(uint64_t)cursor + cursor->size`

jberdine (approved):

PR #136, opened 2024-02-28
Dump did not output table indices

gregsexton:
This meant that rebuilds would drop indices - which is bad - but also
that indices checked for by privacy logic would be missing and cause
runtime exceptions, permanently breaking the service after upgrade.


sql/src/SqlTables.sk
```diff
@@ -747,10 +751,14 @@ fun dumpTables(origContext: readonly SKStore.Context, suffix: ?String): void {
       dirDescr = values.first;
       if (!dirDescr.isInput) return void;
       printDirDescr(dirDescr, suffix);
+      tableName = dirDescr.name.lower;
       for (indexDescr in getIndexByName(context).getArray(
         context,
-        SKStore.SID(dirDescr.dirName.toString()),
+        SKStore.SID(tableName),
       )) {
+        if (indexDescr.isSystemIndex()) {
```
gregsexton, line 759: Test from my fun

PR #137, opened 2024-03-03
[sqlparser] Replace const `Map` with static function.

beauby:
The initialization of the map incurs a high cost at start up (as found out by @gregsexton). Using a function with a pattern match should fix that, and allow more optimizations.

### DIFF
--- a/sqlparser/src/Keyword.sk
+++ b/sqlparser/src/Keyword.sk
@@ -303,156 +303,159 @@ base class Keyword uses Equality, Show, Hashable {
   | TWith() -> "WITH"
   | TWithout() -> "WITHOUT"
 
-  const kwMap: Map<String, Keyword> = Map[
-    "ABORT" => TAbort(),
-    "ACTION" => TAction(),
-    "ADD" => TAdd(),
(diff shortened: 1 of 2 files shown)


mbouaziz:
> The initialization of the map incurs a high cost at start up
Weird...

LGTM anyway

jberdine (approved):
It is concerning that creating a map is expensive, given that they are used many places, but LGTM.

gregsexton:
Nice! Thanks for doing this, I'll test it out.

Yeah the map is constructed at runtime during skip init called from main. Running `skdb --init` goes from an average of 0.0128 secs to 0.0179 if I checkout before and after the new parser is merged in. So in absolute terms it's not exactly expensive, but this is a 40% regression. It's across the board for all startup. Which _really_ hurts the simulator as it starts a huge number of processes. This map accounts for ~half of the regression. The rest is the const values in SQLPrivacy.

bennostein:
> The rest is the const values in SQLPrivacy.

The PredefinedTable singletons? Could look into moving away from Maps there too if that helps startup speeds.

PR #138, opened 2024-03-04
[runtime] Fix warnings

mbouaziz:
In #135, @jberdine found that we used pointer arithmetic on `void`. This could have been avoided by enabling compiler warnings. So let's do it and fix a few more oddities.

I've run
```shell
set -a; SKARGO_PROFILE=dev; STAGE=2; PATH=$PWD/compiler/stage$STAGE/bin:$PATH; set +a; make clean && make -C compiler clean test && make test
```
as suggested by @jberdine

### DIFF
--- a/compiler/runtime/Makefile
+++ b/compiler/runtime/Makefile
@@ -1,6 +1,6 @@
 MEMSIZE32=1073741824
 
-COMMONFLAGS=-Wpointer-arith
+COMMONFLAGS=-Wpointer-arith -Wunused-variable
 OLEVEL=-O2
 CC32FLAGS=-DSKIP32 --target=wasm32 -emit-llvm -nostdlibinc $(COMMONFLAGS)
 CC64FLAGS=$(OLEVEL) -DSKIP64 $(COMMONFLAGS)
--- a/compiler/runtime/Makefile
+++ b/compiler/runtime/Makefile
@@ -1,6 +1,6 @@
 MEMSIZE32=1073741824
 
-COMMONFLAGS=-Wall -Wno-sometimes-uninitialized -Wunused-parameter -Wno-c2x-extensions
+COMMONFLAGS=-Wall -Wno-sometimes-uninitialized -Wunused-parameter -Wno-c2x-extensions -Wsign-compare
 OLEVEL=-O2
 CC32FLAGS=-DSKIP32 --target=wasm32 -emit-llvm -nostdlibinc $(COMMONFLAGS)
 CC64FLAGS=$(OLEVEL) -DSKIP64 $(COMMONFLAGS)
--- a/compiler/runtime/Makefile
+++ b/compiler/runtime/Makefile
@@ -1,6 +1,6 @@
 MEMSIZE32=1073741824
 
-COMMONFLAGS=-Wpointer-arith -Wunused-variable
+COMMONFLAGS=-Wpointer-arith -Wunused-variable -Wsometimes-uninitialized
 OLEVEL=-O2
 CC32FLAGS=-DSKIP32 --target=wasm32 -emit-llvm -nostdlibinc $(COMMONFLAGS)
 CC64FLAGS=$(OLEVEL) -DSKIP64 $(COMMONFLAGS)
--- a/compiler/runtime/Makefile
+++ b/compiler/runtime/Makefile
@@ -1,6 +1,6 @@
 MEMSIZE32=1073741824
 
-COMMONFLAGS=-Wall -Wno-sometimes-uninitialized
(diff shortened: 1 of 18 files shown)


compiler/runtime/obstack.c
```diff
@@ -79,14 +79,15 @@ void sk_obstack_attach_page(char* lpage) {
 }
```
jberdine, line 79: I can't comment on the commit summary, but does "My VSCode is configured with format-on-save and uses the standard Microsoft C/C++ Extension Pack." mean that it runs clang-format, or does its own thing?

compiler/runtime/runtime.c
```diff
@@ -17,7 +17,6 @@ SKIP_gc_type_t* get_gc_type(char* skip_object) {
   return *slot1;
 }
 
-
```
jberdine, line 20: Aren't the *local_context* functions below also dead? And then also the comment right below here.

compiler/runtime/runtime.c
```diff
@@ -83,32 +82,6 @@ void* SKIP_llvm_memcpy(char* dest, char* val, SkipInt len) {
 /* Global context synchronization. */
 /*****************************************************************************/
 
-#ifdef SKIP32
```
jberdine, line 86: The comment above goes with this removed code.

compiler/runtime/runtime32_specific.c
```diff
@@ -93,7 +93,7 @@ void sk_context_set_unsafe(char* obj) {
   context = obj;
 }
 
-SkipInt SKIP_genSym(SkipInt n) {
+SkipInt SKIP_genSym(SkipInt /* n */) {
```
jberdine, line 96: It isn't clear to me why this arg being unused isn't a bug. The native version of this function ensures to return a value that is larger than the arg, but the wasm one does not. @pikatchu do you know why?

jberdine (commented):
Overall LGTM. There is one change noted inline that I don't know why the warning isn't pointing to a bug. It would be good to understand that before silencing the warning.

compiler/runtime/obstack.c
mbouaziz: I was too lazy to check...
Looks like it's indeed clang-format: https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting and it's using the `.clang-format` file already present.

compiler/runtime/runtime.c
mbouaziz: Hmm, indeed!

compiler/runtime/runtime.c
mbouaziz: There's still code in this section that seem to be related to the **global** context

compiler/runtime/runtime.c
mbouaziz: Now removed

compiler/runtime/runtime32_specific.c
mbouaziz: Opened https://github.com/SkipLabs/skdb/issues/141 so we can keep moving

jberdine (approved):

compiler/runtime/Makefile
```diff
@@ -1,8 +1,12 @@
 MEMSIZE32=1073741824
 
+COMMONFLAGS=-Wall -Wextra -Wno-sign-conversion -Wno-sometimes-uninitialized -Wno-c2x-extensions -Wsign-compare -Wextra-semi-stmt
```
gregsexton, line 3: I would probably throw in -Wconversion too. That tends to catch a lot of useful things.

gregsexton:
This is great! I've wanted to do this for a long time :D. If you've got cycles and fancy it: I was also thinking of throwing in ubsan and asan for debug builds.

compiler/runtime/Makefile
mbouaziz: :+1: 

I actually did `-Weverything` and looked at warnings that appeared and could be worth fixing later.

mbouaziz:
> I was also thinking of throwing in ubsan and asan for debug builds.

Created https://intern.skiplabs.io/tasks/3bHdAtQiSR_wcnxJtJ9uX_XdsUC for it ;-)

gregsexton:
Do we have -werror somewhere? Or are these all just warnings?

mbouaziz:
They are just warnings for now. werror would be good if people agree.

bennostein:
> They are just warnings for now. werror would be good if people agree.

Bring on the `-Werror`!

jberdine:
> > They are just warnings for now. werror would be good if people agree.
> 
> Bring on the `-Werror`!

At least for the release profile, but I have to manually set CXXFLAGS when iterating on the compiler anyhow so adding `-Wno-error` ought to work.

PR #139, opened 2024-03-04
[test] Remove use of load-csv from test_json

jberdine:


bennostein:
There are a few in `sql/test/TPC-h/makeSqliveDb.sh` as well -- want to get those too while we're at it?

bennostein:
LGTM!

sql/test_json.sh
```diff
@@ -86,9 +86,6 @@ else
     fail "COUNT_PATHS"
 fi
 
-echo "\"id555\", \"{\"\"c\"\":33}\"" | $SKDB load-csv t1
-echo "\"id555\", \"{\"\"c\"\":33}\"" | $SKDB load-csv t2
```
gregsexton, line 90: We're missing the test data now? We should load the data in using an insert IMO, will be clearer.

It looks like below statements will fail, but none of the exit codes are checked.

jberdine: Is the current version what you had in mind?

sql/test/TPC-h/makeSqliveDb.sh
```diff
@@ -119,13 +119,4 @@ CREATE TABLE SUPPLIER (
...
 " | $SKDB --data /tmp/test.db
 
-echo "select * from customer;" | sqlite3 $TPCHDB -csv | $SKDB load-csv CUSTOMER --data /tmp/test.db
-echo "select * from lineitem;" | sqlite3 $TPCHDB -csv | head -n 500000 | $SKDB load-csv LINEITEM --data /tmp/test.db
-echo "select * from nation;" | sqlite3 $TPCHDB -csv | $SKDB load-csv NATION --data /tmp/test.db
-echo "select * from orders;" | sqlite3 $TPCHDB -csv | $SKDB load-csv ORDERS --data /tmp/test.db
-echo "select * from part;" | sqlite3 $TPCHDB -csv | $SKDB load-csv PART --data /tmp/test.db
-echo "select * from partsupp;" | sqlite3 $TPCHDB -csv | $SKDB load-csv PARTSUPP --data /tmp/test.db
-echo "select * from region;" | sqlite3 $TPCHDB -csv | $SKDB load-csv REGION --data /tmp/test.db
-echo "select * from supplier;" | sqlite3 $TPCHDB -csv | $SKDB load-csv SUPPLIER --data /tmp/test.db
```
jberdine, line 129: Is there a replacement for this sort of usage?

sql/test_json.sh
gregsexton: Yeah I think that looks good. I guess the test is happy now?

sql/test/TPC-h/makeSqliveDb.sh
gregsexton: Can we get sqlite to dump out sql statements instead of csv? Then could just load the whole thing.

gregsexton: Looks like there's a .dump command. https://stackoverflow.com/a/7526055

sql/test_json.sh
jberdine: Yes, but it doesn't mean much as I still have to figure out how to make the test unhappy when just deleting these lines...

sql/test/TPC-h/makeSqliveDb.sh
jberdine: I'll have a look

sql/test/TPC-h/test_tpch.sh
```diff
@@ -15,7 +15,7 @@ rm -f /tmp/nation_count
 START=$SECONDS
 ./makeSqliveDb.sh
 TOTAL=$(($SECONDS - $START))
-if (( TOTAL > 180 ));
+if (( TOTAL > 1800 ));
```
jberdine, line 18: It seems that load-csv was a lot faster

bennostein: Am I reading this right as a ~10x slowdown going from (a) sqlite CSV output piped into `SKDB load-csv` to (b) sqlite SQL-dump output piped into `SKDB`?

That makes sense to me -- when we're dealing with csv input we don't do any (sql) parsing, batch those writes into reasonable chunks, etc.

But if the test is taking 30 minutes now, that seems like an issue.  We could pipe more tables through `head` to cut down the input sizes as a bandaid fix I suppose, and revisit once we do some more work on bulk insertions?

jberdine: On my machine the actual time I see is 1334, but I didn't go back and check what a recent time with load-csv was. I did run this test before and had it pass though, so it still is roughly in the region of 10x. This test is not run by default though, it takes far more memory than the CI machines have.

PR #140, opened 2024-03-04
Date tests: force locale

mbouaziz:
These tests call the `date` command line which would print values in your current locale and would make the tests `formats 1/2/3/4` fail.

`cd sql && ./test_data.sh` now works even if your `LC_TIME` is set to something as weird as `fr_FR.UTF-8` before


jberdine (approved):

mbouaziz:
@skiplabsdaniel?

skiplabsdaniel:
LGTM. Thanks for doing that.

PR #142, opened 2024-03-04
Palloc cleanups

mbouaziz:
A bunch of small cleanups followed by other simplifications changing the file mapping format.

Tested with
```shell
set -a; SKARGO_PROFILE=dev; STAGE=2; PATH=$PWD/compiler/stage$STAGE/bin:$PATH; set +a
make clean && make -C compiler clean test && make test
```

### DIFF
--- a/compiler/runtime/intern.c
+++ b/compiler/runtime/intern.c
@@ -329,15 +329,6 @@ void* SKIP_intern_shared(void* obj) {
     }
   }
 
-  {
-    unsigned int i;
-    for (i = 0; i < nbr_pages; i++) {
-      if (large_pages[i]) {
-        pages[i].value = (uint64_t)pages[i].key;
-      }
-    }
-  }
-
   sk_free_size(pages, sizeof(sk_cell_t) * nbr_pages);
   sk_free_size(large_pages, sizeof(int) * nbr_pages);
   sk_stack_free(st);
--- a/compiler/runtime/intern.c
+++ b/compiler/runtime/intern.c
@@ -249,14 +249,6 @@ void* SKIP_intern_shared(void* obj) {
   sk_stack3_t* st3 = &st3_holder;
   size_t nbr_pages = sk_get_nbr_pages(NULL);
   sk_cell_t* pages = sk_get_pages(nbr_pages);
-  int* large_pages = sk_malloc(sizeof(int) * nbr_pages);
-
-  {
-    unsigned int i;
-    for (i = 0; i < nbr_pages; i++) {
-      large_pages[i] = 0;
-    }
-  }
 
   sk_stack_init(st, STACK_INIT_CAPACITY);
   sk_stack3_init(st3, STACK_INIT_CAPACITY);
@@ -275,10 +267,6 @@ void* SKIP_intern_shared(void* obj) {
       }
 
       continue;
-    }
-
-    if (sk_is_large_page(pages[obstack_idx].key)) {
-      large_pages[obstack_idx] = 1;
     }
 
     void* interned_ptr;
@@ -330,7 +318,6 @@ void* SKIP_intern_shared(void* obj) {
   }
 
   sk_free_size(pages, sizeof(sk_cell_t) * nbr_pages);
-  sk_free_size(large_pages, sizeof(int) * nbr_pages);
   sk_stack_free(st);
   sk_stack3_free(st3);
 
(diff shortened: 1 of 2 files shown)


compiler/runtime/intern.c
```diff
@@ -249,14 +249,6 @@ void* SKIP_intern_shared(void* obj) {
   sk_stack3_t* st3 = &st3_holder;
   size_t nbr_pages = sk_get_nbr_pages(NULL);
   sk_cell_t* pages = sk_get_pages(nbr_pages);
-  int* large_pages = sk_malloc(sizeof(int) * nbr_pages);
```
gregsexton, line 252: I see you're removing this because it's not read. Was it ever? Is it actually a bug that we're not reading it?

gregsexton:
I haven't gone through this carefully. But looks full of wins to me. :)

compiler/runtime/intern.c
mbouaziz: It was read in a piece of code that I removed because it was dead as well.

I'm not sure why it was there, I supposed it was a remain from the past...

I see how `copy` benefits from large pages: it reuses them instead of copying them. But I don't see how `intern` could benefit from large pages.

mbouaziz: Looks like @skiplabsdaniel did something related in 5a3176f02a286efe122609c30132f6f7f9eb5648

Edit: but it seems already dead at that time

mbouaziz: It was added by @pikatchu in the tiny commit 14f86be2a9b3ef127c48f0a3fed33813fce8a84a

mbouaziz: I think the goal was either:

1. to mark the large page as empty so that it can be reused, which kinda makes sense, but:
- to me, nothing guarantees there are no other pointers to the object in the large page, but I haven't looked at (hidden?) invariants on the usage of `intern` which could guarantee that;
- this was modifying only the local array of pages which doesn't survive `intern`.

2. or to make the binary search skip over the large pages that have already been treated, but I don't see the point. Did I miss something?

gregsexton: Looks like it's probably just dead code from a failed experiment. Just thought I'd double check. :)

jberdine: Could also be left-over debugging code hunting for a memory leak.

compiler/runtime/palloc.c
```diff
@@ -636,32 +636,29 @@ void* sk_get_ftable(size_t size) {
...
-  }
-  *ginfo = malloc(sizeof(ginfo_t));
-  if (*ginfo == NULL) {
+  no_file_t* no_file = malloc(sizeof(no_file_t));
+  if (no_file == NULL) {
     perror("malloc");
     exit(1);
   }
+  ginfo = &no_file->ginfo;
+  *ginfo = &no_file->ginfo_data;
```
jberdine: Would `no_file->ginfo = &no_file->ginfo_data` be easier to read?

compiler/runtime/palloc.c
```diff
@@ -438,9 +438,21 @@ void sk_commit(char* new_root, uint32_t sync) {
 
 typedef struct {
   int64_t version;
-  void* bottom_addr;
+  struct file_mapping* bottom_addr;
```
jberdine: Would `file_mapping_t` instead of `struct file_mapping` work here?

compiler/runtime/palloc.c
```diff
@@ -506,7 +500,6 @@ void sk_create_mapping(char* fileName, char* static_limit, size_t icapacity) {
   memcpy(persistent_fileName, fileName, fileName_length);
 
   *ginfo = ginfo_data;
-
   (*ginfo)->ginfo_array = ginfo_data;
```
jberdine: Would it be clearer to move these up to where ginfo_data is set?

jberdine: Ah, but I see you later remove these, nm.

jberdine (approved):
Nice!

compiler/runtime/palloc.c
mbouaziz: I removed the `ginfo` field later anyway

compiler/runtime/palloc.c
mbouaziz: I thought you couldn't refer to a type defined later, I'll double-check, it may have changed since I learned C

jberdine: Isn't file_mapping also only defined later? I'm not positive, C has funky distinctions between just referring to names that might later get defined vs actually declaring names.

mbouaziz: Indeed, I could move the typedef earlier and keep the declaration below.

PR #144, opened 2024-03-05
Fix test failures with simulator and make it run 2-3x faster

gregsexton:
The failures were horribly non-deterministic and turned out to be a
problem with how the simulator models the system, rather than the
system itself.


jberdine (commented):
LGTM. The changes to SqlCsv.sk make sense to me.

PR #145, opened 2024-03-06
Runtime/compiler cleanups related to gc types

mbouaziz:
Here are a few cleanups pertaining to `SKIP_gc_type_t`, its creation and its usage.

Follows simplifications and code sharing between the only two kinds of non-string skip objects: class instances and arrays.

Test with `STAGE=C make clean && make -C compiler clean test && make test`

### DIFF
--- a/compiler/runtime/Makefile
+++ b/compiler/runtime/Makefile
@@ -7,7 +7,6 @@ COMMONFLAGS=-Werror -Wall -Wextra -Wno-sign-conversion -Wno-sometimes-uninitiali
 OLEVEL=-O2
 CC32FLAGS=-DSKIP32 --target=wasm32 -emit-llvm -nostdlibinc $(COMMONFLAGS)
 CC64FLAGS=$(OLEVEL) -DSKIP64 $(COMMONFLAGS)
-CXX64FLAGS=$(OLEVEL) $(COMMONFLAGS)
 
 # NB: this MUST be kept in sync with CRELFILES in prelude/build.mk
 # and CRELFILES in prelude/build_wasm32.mk
@@ -62,7 +61,7 @@ libbacktrace: .libs/libbacktrace.a
 	$(MAKE) -C libbacktrace
 
 runtime64_specific.o: runtime64_specific.cpp .libs/libbacktrace.a
-	clang++ -std=c++11 $(CXX64FLAGS) -g3 -o $@ -c -Ilibbacktrace/ $<
+	clang++ -std=c++11 $(CC64FLAGS) -g3 -o $@ -c -Ilibbacktrace/ $<
 
 %.o: %.c
 	clang $(CC64FLAGS) -g3 -o $@ -c $<
--- a/compiler/runtime/Makefile
+++ b/compiler/runtime/Makefile
@@ -1,12 +1,12 @@
 MEMSIZE32=1073741824
 
-COMMONFLAGS=-Werror -Wall -Wextra -Wno-sign-conversion -Wno-sometimes-uninitialized -Wno-c2x-extensions -Wsign-compare -Wextra-semi-stmt
+OLEVEL=-O2
+COMMONFLAGS=$(OLEVEL) -Werror -Wall -Wextra -Wno-sign-conversion -Wno-sometimes-uninitialized -Wno-c2x-extensions -Wsign-compare -Wextra-semi-stmt
 # To add: -Wcast-align -Wcast-qual -Wimplicit-int-conversion -Wmissing-noreturn -Wpadded -Wreserved-identifier -Wshorten-64-to-32 -Wtautological-unsigned-zero-compare
 # -Watomic-implicit-seq-cst ?
 # If using -Weverything, you may want to add -Wno-declaration-after-statement -Wno-missing-prototypes -Wno-missing-variable-declarations -Wno-shadow -Wno-strict-prototypes -Wno-zero-length-array -Wno-unreachable-code-break -Wno-unreachable-code-return
-OLEVEL=-O2
 CC32FLAGS=-DSKIP32 --target=wasm32 -emit-llvm -nostdlibinc $(COMMONFLAGS)
-CC64FLAGS=$(OLEVEL) -DSKIP64 $(COMMONFLAGS)
+CC64FLAGS=-DSKIP64 $(COMMONFLAGS)
 
 # NB: this MUST be kept in sync with CRELFILES in prelude/build.mk
 # and CRELFILES in prelude/build_wasm32.mk
@@ -47,7 +47,7 @@ libskip_runtime32.bc: $(BCFILES32)
 	llvm-link $(BCFILES32) -o $@
 
 %.bc: %.c
-	clang $(OLEVEL) $(CC32FLAGS) -o $@ -c $<
+	clang $(CC32FLAGS) -o $@ -c $<
 
 libskip_runtime64.a: $(OFILES) runtime64_specific.o $(ONATIVE_FILES)
 	$(AR) rcs $@ $^
--- a/compiler/runtime/native_eq.c
+++ b/compiler/runtime/native_eq.c
@@ -41,13 +41,13 @@ SkipInt SKIP_native_eq_class(sk_stack_t* st, char* obj1, char* obj2) {
   }
 
   size_t size = ty1->m_userByteSize / sizeof(void*);
-  size_t bitsize = sizeof(void*) * 8;
+  const size_t refMaskWordBitSize = sizeof(ty1->m_refMask[0]) * 8;
   size_t mask_slot = 0;
   unsigned int i;
   while (size > 0) {
-    for (i = 0; i < bitsize && i < size; i++) {
-      void* ptr1 = *(((void**)obj1) + (mask_slot * bitsize) + i);
-      void* ptr2 = *(((void**)obj2) + (mask_slot * bitsize) + i);
+    for (i = 0; i < refMaskWordBitSize && i < size; i++) {
+      void* ptr1 = *(((void**)obj1) + (mask_slot * refMaskWordBitSize) + i);
+      void* ptr2 = *(((void**)obj2) + (mask_slot * refMaskWordBitSize) + i);
 
       if (ty1->m_refMask[mask_slot] & (1 << i) && ptr1 != ptr2) {
         sk_stack_push(st, ptr1, ptr2);
@@ -57,10 +57,10 @@ SkipInt SKIP_native_eq_class(sk_stack_t* st, char* obj1, char* obj2) {
         }
       }
     }
-    if (size < bitsize) {
+    if (size < refMaskWordBitSize) {
       break;
     }
-    size -= bitsize;
+    size -= refMaskWordBitSize;
     mask_slot++;
   }
 
@@ -84,11 +84,11 @@ SkipInt SKIP_native_eq_array(sk_stack_t* st, char* obj1, char* obj2) {
 
   size_t memsize = ty1->m_userByteSize * len1;
 
-  size_t bitsize = sizeof(void*) * 8;
-
   if ((ty1->m_refsHintMask & 1) == 0) {
     return (SkipInt)memcmp(obj1, obj2, memsize);
   }
+
+  const size_t refMaskWordBitSize = sizeof(ty1->m_refMask[0]) * 8;
 
   char* ohead1 = obj1;
   char* ohead2 = obj2;
@@ -99,7 +99,7 @@ SkipInt SKIP_native_eq_array(sk_stack_t* st, char* obj1, char* obj2) {
     size_t mask_slot = 0;
     while (size > 0) {
       unsigned int i;
-      for (i = 0; i < bitsize && size > 0; i++) {
+      for (i = 0; i < refMaskWordBitSize && size > 0; i++) {
         void* ptr1 = *((void**)ohead1);
         void* ptr2 = *((void**)ohead2);
 
--- a/compiler/runtime/native_eq.c
+++ b/compiler/runtime/native_eq.c
@@ -150,14 +150,14 @@ SkipInt SKIP_native_eq_helper(sk_stack_t* st, char* obj1, char* obj2) {
   }
 
   switch (ty1->m_kind) {
-    case 0:
+    case kSkipGcKindClass:
       return SKIP_native_eq_class(st, obj1, obj2);
       break;
-    case 1:
+    case kSkipGcKindArray:
       return SKIP_native_eq_array(st, obj1, obj2);
       break;
     default:
-      // NOT SUPPORTED
+      // IMPOSSIBLE
       SKIP_exit((SkipInt)-1);
   }
 
--- a/compiler/runtime/native_eq.c
+++ b/compiler/runtime/native_eq.c
@@ -103,10 +103,10 @@ SkipInt SKIP_native_eq_array(sk_stack_t* st, char* obj1, char* obj2) {
         void* ptr1 = *((void**)ohead1);
         void* ptr2 = *((void**)ohead2);
 
-        if (ty1->m_refMask[mask_slot] & (1 << i) && ptr1 != ptr2) {
-          sk_stack_push(st, ptr1, ptr2);
-        } else {
-          if (ptr1 != ptr2) {
+        if (ptr1 != ptr2) {
+          if (ty1->m_refMask[mask_slot] & (1 << i)) {
+            sk_stack_push(st, ptr1, ptr2);
+          } else {
             return 1;
           }
         }
--- a/compiler/runtime/native_eq.c
+++ b/compiler/runtime/native_eq.c
@@ -26,7 +26,7 @@
  */
 /*****************************************************************************/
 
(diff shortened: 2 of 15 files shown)


jberdine:
To avoid duplicate work, see https://github.com/SkipLabs/skdb/pull/114. I also have a branch removing (some of) the padding. I'll push updated versions of them shortly.

mbouaziz:
Fortunately, not much duplicate work

compiler/runtime/runtime.h
```diff
@@ -146,25 +146,57 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
...
-  void* unused2;
-#endif
-  void* unused;
-  size_t m_refMask[0];
+  uint8_t m_kind;  // either kSkipGcKindClass or kSkipGcKindArray
+  uint16_t m_unused_padding1;
+  uint32_t m_unused_padding2;
+  SkipInt m_userByteSize;
+  SkipInt m_refMask[0];
+  // follows a 0-terminated name
```
jberdine:
```suggestion
  // a 0-terminated name follows if m_hasName is true
```

compiler/runtime/runtime.h
```diff
@@ -146,25 +146,57 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
...
 
+#define kSkipGcKindClass 0
+#define kSkipGcKindArray 1
+
 // vtable entries created by createGCType in vtable.sk
 typedef struct {
   uint8_t m_refsHintMask;
-  uint8_t m_kind;
-  uint8_t m_tilesPerMask;
-  uint8_t m_hasName;
```
jberdine: These are set in vtable.sk, but maybe never read. I think the tilesPerMask could be left over from the old GC.

compiler/runtime/runtime.h
```diff
@@ -146,25 +146,57 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
...
-  uint16_t m_uninternedMetadataByteSize;
-  uint16_t m_internedMetadataByteSize;
-  size_t m_userByteSize;
-#ifdef SKIP32
-  void* unused1;
-  void* unused2;
-#endif
```
jberdine, line 161: IIUC m_refMask is meant to be 8-aligned even in 32 bit mode, so some of this might still be needed.

compiler/src/vtable.sk
```diff
@@ -660,21 +651,14 @@ private fun createGCType(
...
-        });
-      }
+      // Possible optimization: tile the mask so the inner loop of the GC is tighter.
+      slotInfo.types.eachWithIndex((idx, x) -> {
+        scalarType = x.getScalarType(env);
+        if (scalarType.isGCPointer()) {
+          bitOffset = slotInfo.bitOffsets[idx];
+          setBit(bitOffset, !isAllDeepFrozen && !x.isDeepFrozen())
+        }
+      })
```
jberdine: I don't understand why to make this change. This is setting the refMask, which is still used e.g. in copy.c. I suppose that tilesPerMask need not be stored in the vtable entry, but it still seems like the mask should be tiled out. Am I confused?

compiler/src/vtable.sk
```diff
@@ -572,8 +572,8 @@ private fun createGCType(
   // m_unused_tilesPerMask
   pushField(kByteConstants[0]);
 
-  // m_hasName
-  pushField(kByteConstants[if (provideName) 1 else 0]);
+  // m_unused_hasName
+  pushField(kByteConstants[0]);
```
jberdine: If this is made constantly false, the code in vtable.sk that puts the string there should be removed at the same time. But it might be better to leave it in non-release mode to make debugging vtable things easier. Not sure, it seems odd to have a string there and no way to know from the code.

compiler/runtime/runtime.h
```diff
@@ -150,10 +150,8 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
 typedef struct {
   uint8_t m_refsHintMask;
   uint8_t m_kind;
-  uint8_t m_unused_tilesPerMask;
-  uint8_t m_unused_hasName;
   uint16_t m_uninternedMetadataByteSize;
-  uint16_t m_unused_internedMetadataByteSize;
+  uint32_t m_unused_padding;
```
jberdine: Won't doing this make accesses to uninternedMetadataByteSize unaligned, without a reduction in size?

compiler/runtime/native_eq.c
```diff
@@ -69,17 +69,21 @@ SkipInt SKIP_native_eq_class(sk_stack_t* st, char* obj1, char* obj2) {
...
   size_t len1 = skipArrayLen(obj1);
-  size_t len2 = skipArrayLen(obj2);
+  {
+    size_t len2 = skipArrayLen(obj2);
 
-  if (len1 != len2) {
-    return 1;
+    if (len1 != len2) {
+      return 1;
+    }
```
jberdine: Not objecting, but why do this?

jberdine (commented):
Nice! Very good to get rid of more of the stale support for frozen. A couple questions inline. I or someone else should probably take another look over this with fresh eyes.

compiler/runtime/runtime.h
mbouaziz: Yes, but I remove `m_hasName` later. I didn't see any piece of code using the name actually. I suspect it could be useful for debugging, right?

compiler/runtime/runtime.h
mbouaziz: I keep it 8-byte-aligned and it is dynamically checked in `vtable.sk`:
```
  if (fieldBitOffset % 64 != 0) {
    sc.die("bad vtable entry alignment")
  };

  // m_refMask
  ```

File: compiler/src/vtable.sk (hunk 4 above)
Comment (mbouaziz) [reply compiler/src/vtable.sk]: If _the inner loop of the GC_ refers to what is in `SKIP_intern_array` then the tiling is not used as of today. I could rewrite it so that it's used btw, I doubt tough that it makes things really faster.

File: compiler/runtime/runtime.h (hunk 6 above)
Comment (mbouaziz) [reply compiler/runtime/runtime.h]: Will it?

Comment (mbouaziz) [reply compiler/runtime/runtime.h]: `clang` produces the offsets 0, 1, 2, 4, 8 that I expected for this struct:
```c
struct {
    uint8_t a;
    uint8_t b;
    uint16_t c;
    uint32_t d;
    uint64_t e;
}
```

compiler/runtime/native_eq.c
mbouaziz: Right, using `len2` instead of `len1` wouldn't be an issue... I'll revert

compiler/src/vtable.sk
mbouaziz: IIUC, there used to be, or was an attempt to, tile out the refMask as much as possible in a 64-bit mask to avoid rereading it from the beginning when processing arrays. But no runtime function processing arrays make use of it.

compiler/src/vtable.sk
mbouaziz: Let's keep the `m_hasName` field then. It's not costing any space given all the paddings we have.

Note: we only use one bit of the `refsHintMask` and one bit for the `kind`, so we could save another 8 bytes by storing the size on a 62 bits. But I don't think it's worth optimizing this struct.

mbouaziz:
I've left only the commits that make all combinations of (old runtime, new runtime) x (old compiler, new compiler) work together. There's only a couple minor commits left for later.

mbouaziz:
I wrongly assumed that only the runtime accessed stuff but the compiler may produce low-level accesses as well that I need to check.

Edit: I checked and it looks like the compiler does not generate accesses to the gc type, only other vtable entries.

compiler/runtime/runtime.h
jberdine: Ah yes, I overlooked the size change implied by the change of type for m_userByteSize.

compiler/runtime/Makefile
```diff
@@ -1,12 +1,13 @@
...
 
-COMMONFLAGS=-Wall -Wextra -Wno-sign-conversion -Wno-sometimes-uninitialized -Wno-c2x-extensions -Wsign-compare -Wextra-semi-stmt
+COMMONFLAGS=-Werror -Wall -Wextra -Wno-sign-conversion -Wno-sometimes-uninitialized -Wno-c2x-extensions -Wsign-compare -Wextra-semi-stmt
 # To add: -Wcast-align -Wcast-qual -Wimplicit-int-conversion -Wmissing-noreturn -Wpadded -Wreserved-identifier -Wshorten-64-to-32 -Wtautological-unsigned-zero-compare
 # -Watomic-implicit-seq-cst ?
 # If using -Weverything, you may want to add -Wno-declaration-after-statement -Wno-missing-prototypes -Wno-missing-variable-declarations -Wno-shadow -Wno-strict-prototypes -Wno-zero-length-array -Wno-unreachable-code-break -Wno-unreachable-code-return
 OLEVEL=-O2
 CC32FLAGS=-DSKIP32 --target=wasm32 -emit-llvm -nostdlibinc $(COMMONFLAGS)
 CC64FLAGS=$(OLEVEL) -DSKIP64 $(COMMONFLAGS)
+CXX64FLAGS=$(OLEVEL) $(COMMONFLAGS)
```
jberdine: Is it done this way to not pass `-DSKIP64` when compiling c++ code? It might not be necessary, but I think it would be surprising later to find that `SKIP64` isn't defined in code that is part of the 64 bit runtime.

I also have not understood why the parts that are compiled as c++ aren't c, or vice versa. Maybe that non-understanding is related here.

compiler/runtime/runtime.h
```diff
@@ -146,25 +146,60 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
...
+} sk_array_t;
+
+#define skipArrayLen(obj) \
+  (((sk_array_t*)((char*)(obj)-offsetof(sk_array_t, data)))->length)
+
+/*****************************************************************************/
+/* SKIP objects: arrays and class instances. */
+/*****************************************************************************/
+
+#define skipObjLen(ty, obj) (tyIsArray(ty) ? skipArrayLen(obj) : 1)
```
jberdine: I don't have strong views, but here and elsewhere, AFAIU the existing style is to use snake case names in the C/C++ runtime code except when they need to be the same as those exposed to skip code, or to emphasize a connection to the skip names.

compiler/runtime/runtime.h
```diff
@@ -146,25 +146,60 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
...
+   - 1 word for kSkipGcKindClass, its vtable pointer
+   - 2 words for kSkipGcKindArray, its vtable pointer preceded by its size on
+       32 bits, itself preceded by an unused padding of 32 bits on 64-bits arch
+       (see below).
+*/
+#define uninternedMetadataWordSize(ty) ((ty)->m_kind + 1)
+#define uninternedMetadataByteSize(ty) \
+  (uninternedMetadataWordSize(ty) * sizeof(void*))
+
+#define tyIsArray(ty) ((ty)->m_kind)
```
jberdine: This would be clearer and more robust to future changes if it was expressed as a comparison to `kSkipGcKindArray` instead of relying on the macro definitions interpretation as booleans.

compiler/runtime/runtime.h
```diff
@@ -146,25 +146,60 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
...
+/* The uninternedMetadataByteSize is the size preceding the pointer to a
+   non-string GC value. It is:
+   - 1 word for kSkipGcKindClass, its vtable pointer
+   - 2 words for kSkipGcKindArray, its vtable pointer preceded by its size on
+       32 bits, itself preceded by an unused padding of 32 bits on 64-bits arch
+       (see below).
+*/
+#define uninternedMetadataWordSize(ty) ((ty)->m_kind + 1)
+#define uninternedMetadataByteSize(ty) \
+  (uninternedMetadataWordSize(ty) * sizeof(void*))
```
jberdine: Eventually it would be nice if we could write this along the lines of:
```
#define uninternedMetadataWordSize(ty) \
  ((ty)->m_kind == kSkipGcKindArray ? offsetof(sk_array_t, data) : offsetof(sk_class_t, ...))
```

compiler/runtime/runtime.h
```diff
@@ -146,25 +146,60 @@ sk_value3_t sk_stack3_pop(sk_stack3_t* st);
...
+#ifdef SKIP64
+  uint32_t unused_padding;
+#endif
+  uint32_t length;
+  void** vtable;
+  unsigned char data[0];
+} sk_array_t;
+
+#define skipArrayLen(obj) \
+  (((sk_array_t*)((char*)(obj)-offsetof(sk_array_t, data)))->length)
```
jberdine: I would add the following:
```
#define container_of(ptr, type, member) ((type *)((char *)(1 ? (ptr) : &((type *)0)->member) - offsetof(type, member)))
```
and then use it to define this.

compiler/runtime/runtime64_specific.cpp
```diff
@@ -1,3 +1,7 @@
+#ifndef SKIP64
+#define SKIP64
+#endif
```
jberdine: As mentioned on the Makefile, it seems odd to do this here for cpp files and using the build system for c files.

compiler/runtime/runtime64_specific.cpp
```diff
@@ -349,27 +353,32 @@ char* SKIP_open_file(char* filename_obj) {
...
-  char* result = nullptr;
-  if (size == 0) {
-    result = sk_string_alloc(0);
-    sk_string_set_hash(result);
-    return result;
+  if (fd == -1) {
+    perror("ERROR (open failed)");
+    fprintf(stderr, "Could not open file: %s\n", filename);
+    exit(ERROR_FILE_IO);
   }
```
jberdine: While I understand that the control flow regarding testing size for 0 in the new version is equivalent, I think the old code organization --- where this special case is handled by a test for the special case followed by an early return --- is easier to read. It lets you think about the two functionally significantly different cases separately from each other as their implementations aren't in intermingled lines of code. And the style of putting the special cases / error checks first and using early returns is very common and idiomatic.

compiler/src/vtable.sk
jberdine: Ok thanks for checking, it's totally plausible that this is a remnant of the old GC.

compiler/runtime/runtime.h
jberdine: github seems to make it impossible to see what the code was when I wrote that comment, and I don't remember exactly. The current version is good. For reference it is not the offsets being wrong that I'm concerned with, but more that from a performance perspective, it is better if more of the members can be made to start at offsets which are multiples of the word size. So if combining and reordering two padding fields into one results in the same overall size, but fewer fields with offsets that are multiples of the word size, then I would by default avoid making that change.

jberdine (approved):
This looks good. I made another full pass over the final state of the PR. I have a few suggested changes, but nothing that needs another round of review I think.

compiler/runtime/runtime.h
mbouaziz: Anyway I reverted this change for now, because of boostrapping issues

compiler/runtime/Makefile
mbouaziz: I have no idea either.

I agree, there was actually no need for a C++-specific variable, especially because we do want `-DSKIP64` as well but I enforced it in `runtime64_specific.cpp`.

~~Btw this commit was part of #148 on which I rebased this PR. I won't change anything there.~~

compiler/runtime/runtime.h
mbouaziz: I'll snake case

compiler/runtime/runtime.h
mbouaziz: :+1:

compiler/runtime/runtime64_specific.cpp
mbouaziz: Ok, I'll do something in this PR then

compiler/runtime/runtime64_specific.cpp
mbouaziz: The special case is size 0 is an optimization, no need to mmap and memcpy though it would be correct to do it, maybe we could even do it because it won't cost much.

The goal was to avoid duplicating the code that closes the file and frees the filename.

compiler/runtime/Makefile
mbouaziz: I changed to share the flags + ensure `runtime64_specific.cpp` is given `-DSKIP64`

compiler/runtime/runtime.h
mbouaziz: You're killing all the fun :stuck_out_tongue_closed_eyes:

compiler/runtime/runtime64_specific.cpp
jberdine: Ok I see. I plan to change strings to avoid the need to allocate in sk2c_string, so hopefully that will get better soon. But closing will remain. And I guess that currently we even leak the fd for size = 0.

PR #149, opened 2024-03-08
[runtime] Fix segfault upon uncaught exception.

beauby:
In some cases, libbacktrace does not manage to extract the function name for a frame, leading to a `strcmp(NULL, foo)`.


jberdine (approved):
Ah yes, nice

mbouaziz:
<3
/approve

PR #150, opened 2024-03-11
[sql] Implement create-table-as-select and add parser/AST support for alter-table

bennostein:
This is in preparation for some larger schema migration changes, but mostly stands on its own, adding a full implementation for create-table-as-select SQL statements and syntactic support for alter-table statements.

### DIFF
--- a/sql/src/SqlEval.sk
+++ b/sql/src/SqlEval.sk
@@ -476,38 +476,26 @@ class Evaluator{options: Options, user: ?UserFile} {
         this.createTable(context, pos, name, dirDescr, ifNotExists);
       })
     | P.CreateTableAs{pos, name, select, ifNotExists} ->
-      core: P.SelectCoreQuery = select.core match {
-      | P.SelectCoreValues _ ->
-        error(
-          pos,
-          "CREATE TABLE AS requires a SELECT query, not a VALUES literal",
-        )
-      | core @ P.SelectCoreQuery _ -> core
+      compiler = SKDB.Compiler::create(false, this.options, select.pos, params);
+      selectAst = select with {
+        core => select.core match {
+        | vals @ P.SelectCoreValues _ -> compiler.expandValuesClause(vals)
+        | qry @ P.SelectCoreQuery _ -> qry
+        },
       };
 
-      compiler = SKDB.Compiler::create(false, this.options, select.pos, params);
       cselect = compiler.compileSelect(
         context,
-        select,
-        genSelectId(select),
+        selectAst,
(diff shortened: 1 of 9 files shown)


sql/src/SqlEval.sk
```diff
@@ -475,8 +475,63 @@ class Evaluator{options: Options, user: ?UserFile} {
...
       })
-    | P.CreateTableAs _ ->
-      invariant_violation("CREATE TABLE AS SELECT not implemented")
+    | P.CreateTableAs{pos, name, select, ifNotExists} ->
+      core: P.SelectCoreQuery = select.core match {
+      | P.SelectCoreValues _ ->
+        error(
+          pos,
+          "CREATE TABLE AS requires a SELECT query, not a VALUES literal",
+        )
```
beauby, line 484: Note that SQLite accepts the following:
```
CREATE TABLE bar AS VALUES(1, 'hello')
```
and gives the anonymous columns names like `column1`, `column 2`.

sql/src/SqlEval.sk
```diff
@@ -475,8 +475,63 @@ class Evaluator{options: Options, user: ?UserFile} {
...
+
+      (_, cols) = SKDB.Compiler::getCols(core.params, cselect.from);
+
+      schema = cselect.params.mapWithIndex((idx, param) -> {
+        P.ColumnDefinition{
+          name => cols.maybeGet(idx) match {
+          | None() ->
+            error(
+              pos,
+              "Anonymous columns not allowed in CREATE TABLE AS SELECT",
```
beauby, line 504: SQLite also accepts anonymous expressions here, and simply sets the column name to the string representation of the expression:
```
CREATE TABLE foo AS SELECT 2+2;
# Is equivalent to:
# CREATE TABLE foo ("2+2" INT);
# INSERT INTO foo VALUES(2+2);
```

bennostein: Right, I thought that being more restrictive than sqlite made sense for the schema migration application -- one can imagine mistakenly creating a table with anonymous columns and then being stuck with it.

But I can easily remove those restrictions for general uses and just enforce non-anonymous columns for schema migrations in particular.

beauby: I think it's fine as is, especially since it's failing with a clear error message.

sql/src/SqlEval.sk
gregsexton: Bleh! I don't think we should support that. Particularly if we don't support renaming cols afterward.

gregsexton:
LGTM

PR #152, opened 2024-03-14
Simple transaction API for typescript client

bennostein:
I found myself wanting this at times yesterday so thought I'd put something together quickly.

Not married to these particular design choices if people have strong opinions.

Note that I set things up so that either an effectful or a fluent style is possible, i.e. both

```
new SKDBTransaction(skdb)
  .addStatement(...)
  .addStatement(...)
  .params(...)
  .flush()
```

and

```
let txn = new SKDBTransaction(skdb);
for (const stmt of statements)
  txn.addStatement(stmt);
txn.params(...);
txn.flush();
```

should work just fine.

### DIFF
--- a/sql/ts/src/skdb_util.ts
+++ b/sql/ts/src/skdb_util.ts
@@ -111,28 +111,28 @@ export class SKDBTable extends Array<Record<string, any>> {
 /* ***************************************************************************/
 
 export class SKDBTransaction {
-  private query_params: Params;
+  private params: Params;
   private stmts: Array<string>;
   private db_handle: SKDB;
 
   constructor(skdb: SKDB) {
-    this.query_params = {};
+    this.params = {};
     this.stmts = [];
     this.db_handle = skdb;
   }
 
-  addStatement(stmt: string): SKDBTransaction {
+  add(stmt: string): SKDBTransaction {
     this.stmts.push(stmt.trim().replace(/;$/, ""));
     return this;
   }
 
-  params(params: Params): SKDBTransaction {
-    Object.assign(this.query_params, params);
+  addParams(params: Params): SKDBTransaction {
+    Object.assign(this.params, params);
     return this;
   }
 
-  flush(params: Params = {}): Promise<SKDBTable> {
-    this.params(params);
+  commit(additionalParams: Params = {}): Promise<SKDBTable> {
+    this.addParams(additionalParams);
 
     let txn = "BEGIN TRANSACTION;";
     for (const stmt of this.stmts) {
@@ -140,6 +140,6 @@ export class SKDBTransaction {
     }
     txn = txn + "\nCOMMIT;";
 
-    return this.db_handle.exec(txn, this.query_params);
+    return this.db_handle.exec(txn, this.params);
   }
 }
--- a/sql/ts/src/skdb_util.ts
+++ b/sql/ts/src/skdb_util.ts
@@ -1,3 +1,5 @@
+import type { SKDB, Params } from "./skdb_types.js";
+
 /* ***************************************************************************/
 /* The type used to represent callable external functions. */
 /* ***************************************************************************/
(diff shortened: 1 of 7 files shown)


sql/ts/src/skdb_util.ts
```diff
@@ -102,3 +104,42 @@ export class SKDBTable extends Array<Record<string, any>> {
...
+  private stmts: Array<string>;
+  private db_handle: SKDB;
+
+  constructor(skdb: SKDB) {
+    this.query_params = {};
+    this.stmts = [];
+    this.db_handle = skdb;
+  }
+
+  addStatement(stmt: string): SKDBTransaction {
```
gregsexton: Could we call this 'add'? Just because you might want to add queries or statements, and it's shorter to type. :D

sql/ts/src/skdb_util.ts
```diff
@@ -102,3 +104,42 @@ export class SKDBTable extends Array<Record<string, any>> {
...
+    this.stmts.push(stmt.trim().replace(/;$/, ""));
+    return this;
+  }
+
+  params(params: Params): SKDBTransaction {
+    Object.assign(this.query_params, params);
+    return this;
+  }
+
+  flush(params: Params = {}): Promise<SKDBTable> {
```
gregsexton: Let's call this 'commit'? Just to be inline with usual txn keywords.

sql/ts/src/skdb_util.ts
```diff
@@ -102,3 +104,42 @@ export class SKDBTable extends Array<Record<string, any>> {
...
+    this.stmts = [];
+    this.db_handle = skdb;
+  }
+
+  addStatement(stmt: string): SKDBTransaction {
+    this.stmts.push(stmt.trim().replace(/;$/, ""));
+    return this;
+  }
+
+  params(params: Params): SKDBTransaction {
```
gregsexton: It might be a lot of work I guess and I worry about how robust it would be, but it would be really cool if addStatement took the sql and a params object. Then for each query you could specify params independently. The logic here could then do variable renaming to ensure there's no conflict. Maybe there's a global params call too for shared params. Not sure how valuable this is but thought I'd throw out the idea.

sql/ts/src/skdb_util.ts
gregsexton: The params 'set union' behaviour and the argument to flush unioning is a little unexpected imo, but desirable. Maybe we could call the method 'addParams' and the arg here 'additionalParams' or something like that? I read 'params()' as replacement.

I made the opposite mistake to this with `mirror()` and I regret it. I wish mirror were called 'mirrorOnly' or something like that. The mirror verb suggests being additive to me.

gregsexton:
Good idea! Hopefully left some useful feedback on the interface.

sql/ts/src/skdb_util.ts
bennostein: I thought about this (and discussed a bit with Lucas+Josh) but decided against having a `Params` argument to `addStatement` to avoid variable renaming/conflict woes, as this is all happening before any sql parsing.

My intuition is that transactions will often reuse the same param across multiple queries/statements, where a single shared param namespace is desirable.  And I worry about the conceptual complexity of separate per-query params and transaction-wide params with different semantics.

But, one can also imagine wanting to write something like

```
for(const user_id in users)
   txn.addStatement(
     "update users set ... where id=@user_id",
     {user_id}
   )
```

which is not possible in this interface. The parsing and renaming of variables to support that wouldn't be too bad -- just scan for @-variables, taking care to avoid substitutions within string literals.

sql/ts/src/skdb_util.ts
bennostein: Sounds good to me on the shorter-to-type front!  Though I'd quibble that queries are (syntactically) just statements, not some separate thing :)

sql/ts/src/skdb_util.ts
gregsexton: Agree on all fronts.

> taking care to avoid substitutions within string literals.

This - and I'm sure there are other situations - is what I was worried about. Probably end up writing a parser.

sql/ts/src/skdb_util.ts
gregsexton: Yeah, you're right, they are. And you can argue the other way around too that statements are queries that return an empty result set. But in _every_ sql api I've ever used they're split and referred to separately. So this is what people are primed for and why I wrote my comment. 

Incidentally this is why we (after lots of debate) chose 'exec' as the verb for handling both categories. This is really unusual. But much cleaner and simpler IMO.

PR #153, opened 2024-03-14
Implement SQL LENGTH function

bennostein:
This is another thing I found myself wishing for yesterday, pretty straightforward to add.

### DIFF
--- a/sql/src/SqlCAst.sk
+++ b/sql/src/SqlCAst.sk
@@ -451,6 +451,7 @@ class CAnd(CExpr<Int>, CExpr<Int>) extends CExpr<Int>
 class CUp<T: frozen>(CExpr<T>) extends CExpr<T>
 class CCoalesce<T: frozen>(Array<CExpr<T>>) extends CExpr<T>
 class CIIsNull(CExpr<Int>) extends CExpr<Int>
+class CILength(CExpr<String>) extends CExpr<Int>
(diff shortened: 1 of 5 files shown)


sql/test/unit/test_length.sql
```diff
@@ -0,0 +1,5 @@
+create table t (x text, y text, z text);
+
+insert into t values ('', null, 'test');
+
+select x, length(x), y, length(y), z, length(z) from t;
```
gregsexton: could we also test a literal? `length('foo')`

gregsexton (commented):
Awesome, thanks for adding this!

sql/src/SqlCAst.sk
```diff
@@ -451,6 +451,7 @@ class CAnd(CExpr<Int>, CExpr<Int>) extends CExpr<Int>
 class CUp<T: frozen>(CExpr<T>) extends CExpr<T>
 class CCoalesce<T: frozen>(Array<CExpr<T>>) extends CExpr<T>
 class CIIsNull(CExpr<Int>) extends CExpr<Int>
+class CILength(CExpr<String>) extends CExpr<Int>
```
mbouaziz, line 454: Should it be named `CSLength`?

bennostein: Ahh, good catch, I misinterpreted the convention here as that letter denoting the return type rather than the arg type.

PR #154, opened 2024-03-15
[sknpm:ts] Small cleaning

skiplabsdaniel:


bennostein:
All looks good to me!  Some nice cleanups and improvements, thanks

Makefile
```diff
@@ -153,3 +153,14 @@ run-dev-server: sql/target/host/dev/skdb build/init.sql
...
 	cp $^ $@
+
+.PHONY: check-vite
+check-vite: npm
+	rm -rf build/vitejs
+	git clone https://github.com/SkipLabs/skdb_vite_react_starter.git build/vitejs
+	cd build/vitejs && npm install;
+	rm -r build/vitejs/node_modules/skdb
+	cp -r build/package/skdb build/vitejs/node_modules/
+	cd build/vitejs && npm run build
```
gregsexton: Yes! It might also be worth trying to test `npm run dev`. I found often enough that one could be working while the other was broken. But this might be awkward to test actually as you'll need to load the page. :/

gregsexton: But then, to be fair, you probably need to load the page with the prod build too. Due to the dynamic loading, I was often able to build successfully but then it wouldn't actually have resolved everything and page load would fail.

sknpm/src/main.sk
```diff
@@ -885,25 +885,20 @@ fun copyAndConvertImport(
     print_string(">> Copy js files and convert to valid imports.");
   };
   nameLine = "var wasm64 = \"/*--NAME--*/\";";
-  modulesLines = "var modules = [ /*--MODULES--*/];";
+  modulesLines = "/*--MODULES--*/";
   base64Line = "export var base64 = \"\";";
   versionLine = "const npmVersion = \"\";";
-  extensionsLine = "/*--EXTENSIONS--*/";
```
gregsexton, line 894: Ah, cool. I considered doing this, but wasn't sure what it was intended for. I figured best to leave it alone. 🤣

gregsexton (commented):
Looks great. Using URL everywhere is much nicer. :) Did this fix your issue building for prod? If so, what was the problem?

skiplabsdaniel:
> Did this fix your issue building for prod? If so, what was the problem?

It was nothing, my vitejs configuration was no good.

PR #156, opened 2024-03-15
[cli] Fix parsing of boolean options starting with `--no-`.

beauby:
Before this change, it was not possible to declare a boolean option whose name started with `--no-` (unless it was generated as the negation of an other option).

### DIFF
--- a/cli/src/parser.sk
+++ b/cli/src/parser.sk
@@ -199,6 +199,12 @@ private fun parse(
 
       // Use arg name as default long flag for convenience.
       longFlag = arg._long.default(arg.name);
+      if (arg is BoolArg _ && longFlag.startsWith("no-")) {
+        positiveFlag = longFlag.stripPrefix("no-");
+        invariant_violation(
+          `Invalid boolean argument name ${longFlag}. Declare a negatable ${positiveFlag} instead.`,
+        )
+      };
       options.add(`--${longFlag}`, arg);
       if (arg._negatable) {
         options.add(`--no-${longFlag}`, arg);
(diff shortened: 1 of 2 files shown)


jberdine:
For an arg `--no-<foo>`, don't you always want there to be an arg `--<foo>`? The default can be set to either value right?

beauby:
@jberdine I think both ways could be argued, but imagine a user without knowledge of the `--foo`/`--no-foo` mechanism declaring a flag `--no-foo` and having it be always `false` regardless of user input.

jberdine:
Yes, that interpretation is not good. I would say that it would be good to throw an error when trying to manually create an arg starting with `--no-` that says that the unprefixed one should be created instead.

beauby:
That's an option, though I just thought of something: one might want to have the positive flag not be a boolean: `--foo=bar` vs `--no-foo`. What do you think?

bennostein:
> That's an option, though I just thought of something: one might want to have the positive flag not be a boolean: `--foo=bar` vs `--no-foo`. What do you think?

This seems very similar to just having an optional argument foo, whose absence is the `--no-foo` case, no? I'm struggling to think of a case where the different semantics would be helpful -- maybe if the default value were `Some(value)` and you want to be able to override as `Some(other-value)` or disable as `None()`.  But that seems niche, I can't think of a real-world instance.

beauby:
@bennostein Yes, that seems to be the only difference.

There seems to be consensus against standalone `--no-foo` boolean flags, so I'll make it fail upon trying to register such a flag.

PR #159, opened 2024-03-15
[stdlib] Add missing `@debug` annotations in `FileSystem`.

beauby:


jberdine (approved):
Just curious, but is CSE for realpath definitely problematic, or just following the pattern?

beauby:
Just following the pattern at this stage, though there is a real risk for the return value of `realpath()` to change between two calls.

PR #160, opened 2024-03-15
[runtime] Fix warnings (discarded `write()` return value).

beauby:


jberdine (approved):

jberdine:
On the other hand, maybe we should check the return value for -1 to detect errors.

PR #161, opened 2024-03-15
[runtime] Fix inconsistencies in preamble declarations.

beauby:


jberdine (approved):

compiler/runtime/palloc.c
```diff
@@ -208,8 +208,8 @@ int SKIP_cond_timedwait(pthread_cond_t* x, pthread_mutex_t* y, uint32_t secs) {
   return pthread_cond_timedwait(x, y, &ts);
 }
 
-void SKIP_cond_broadcast(void* c) {
-  pthread_cond_broadcast(c);
+int32_t SKIP_cond_broadcast(void* c) {
+  return pthread_cond_broadcast(c);
```
mbouaziz:
```suggestion
  return (int32_t)pthread_cond_broadcast(c);
```
Just to make it explicit we're not keeping the same type.

compiler/runtime/palloc.c
```diff
@@ -208,8 +208,8 @@ int SKIP_cond_timedwait(pthread_cond_t* x, pthread_mutex_t* y, uint32_t secs) {
   return pthread_cond_timedwait(x, y, &ts);
```
mbouaziz:
```suggestion
  return (int32_t)pthread_cond_timedwait(x, y, &ts);
```

compiler/runtime/runtime.c
```diff
@@ -23,13 +23,13 @@ SKIP_gc_type_t* get_gc_type(char* skip_object) {
...
-void SKIP_print_last_exception_stack_trace_and_exit() {
+void SKIP_print_last_exception_stack_trace_and_exit(void*) {
   todo();
 }
-void SKIP_unreachableMethodCall() {
+void SKIP_unreachableMethodCall(void*, void*) {
   todo();
 }
-void SKIP_unreachableWithExplanation() {
+void SKIP_unreachableWithExplanation(void*) {
```
mbouaziz, line 32: We should probably declare them `noreturn`

beauby: They're technically not `noreturn` at the moment, as `todo()` throws a `RuntimeException` which can be caught.

mbouaziz: `noreturn` means the function will not finish its execution with `return` or reaching the end of the function, which is the case.

The exception can only be caught higher up in the stack, it doesn't prevent this function from being marked `noreturn`

PR #162, opened 2024-03-15
[skc] Add support for sklib format.

beauby:
The sklib format contains a compiled skip package. While the skip code
itself cannot be compiled in isolation, possible C extensions
can. Thus, we store arbitrary object files along with the skip source
code within a ranlib archive.
An extra `lib.skmeta` file is stored, containing various metadata
about the package, including the name of the package, its direct
dependencies, and the required linker arguments.

Dependencies to sklib packages are expressed with the `skc` `--dep`
flag, which has two forms:
  - `--dep foo=/path/to/some.sklib`, and
  - `--dep foo`.

The latter will search for a package named `foo` within the search
path (comprised of the `sysroot`, and any path provided with the `-L`
flag). The `sysroot` (which can be overridden through the `--sysroot`
flag) defaults to `../lib/` relative to the location of the `skc`
binary.

Only direct dependencies should be specified with `--dep`. This is
mainly for two reasons: 1. ergonomics, as it lifts the burden of
determining the dependency graph manually, and 2. we might want to
allow multiple versions of a given package to coexist in a dependency
graph (though this will require name-spaced packages). The exact
dependencies of a given package are stored in the `lib.skmeta`,
allowing `skc` to fetch all dependencies in a deterministic way.

In order to improve ergonomics, `--dep std` is implicitly added unless
the `--no-std` flag, or an explicit dependency to `std` is
supplied. This enables building a skip binary through a simple
invocation: `skc foo.sk -o foo`, compared to the previously required
`skc foo.sk path/stdlib/src/**/*.sk -o foo`

Building sklib packages is done by specifying the `--sklib-name=foo`
flag to `skc`.
E.g. `skc --sklib-name=sqlparser -o sqlparser-dev.sklib src/**/*.sk`.
This will run the front end on the provided sources, and output an
`sklib` archive.

Due to the reactive nature of the compiler, we have to take care of
cache invalidation for source files. The simple heuristic used up to
this point consists of keeping only the analysis for the intersection
of collections of provided source files across two successive `skc`
invocations, and re-triggering analysis for files that have
changed (checked by their `mtime`).
For instance, `skc foo.sk bar.sk baz.sk --data state.db` followed by `skc
foo.sk bar.sk foobar.sk --data state.db` would get rid of the analysis
of `bar.sk` (and that of `foo.sk` if its `mtime` has changed between
the two invocations).
This is an issue when building a whole dependency graph by
successively invoking `skc` to generate sklibs in topological order,
as analysis for sources of sklibs in independent sub-graphs would get
invalidated, causing redundant work.

Hence, we need to invalidate more diligently. The current
implementation uses the following heuristic:
- when building an sklib, only invalidate files from the *current
library*, keeping analysis around for all other libraries. This begs
the question of what is the current library? For now, we rely solely
on the value of `--sklib-name` provided on the command line. This
would need to change if we want to support compilation graphs with
several versions of a given library (which would require name-spacing
packages), or if we want to support sharing analysis between target
architectures (which is currently not possible because the front end
uses the pointer width).
- When building a binary, pretend to be building a library with an
special name. (Most binaries are not built this way, as the
corresponding library is generated first, and generating the binary
piggy-backs on the front end analysis of the library and triggering
the back end by picking an entry point. Building binaries by directly
invoking `skc foo.sk -o foo` is supported for convenience but most
projects should rely on `skargo` to handle the build.)

Final remark: building the standard library manually would look like:
```
SKC_PREAMBLE=/path/to/preamble64.ll \
  skc --no-std --sklib-name=std -o libstd.sklib \
      -l static=skip_runtime64 -l static=backtrace \
      -L ../compiler/stage1/lib/ \
      src/**/*.sk
```

Based on #11, #23.


jberdine (commented):
Some comments and questions prior to digging through the code in detail:

- If only for curiousity, is the choice of the ranlib format a free choice or are there external considerations that make it a good idea?

- Any comments one the expectations for what will eventually happen with the dependency cycle between std and skfs?

- Minor, but regarding the name of `--dep`, is it maybe too generic since these are library dependencies, while there might at some point be other sorts of dependencies, such as data to be packed into the binaries. Also, something like `--lib` might go better since it searches in the path set up by `-L`.

- I wonder what happens if some lib is both a direct dep as well as a dep of a dep, and it is mistakenly declared as a dep only of the dep. Do the uses of it as a direct dep lead to resolution failures? Ideally they would, but maybe not something to insist on in the first step.

- I don't understand "For instance, skc foo.sk bar.sk baz.sk --data state.db followed by skc foo.sk bar.sk foobar.sk --data state.db would get rid of the analysis
of bar.sk (and that of foo.sk if its mtime has changed between
the two invocations).". Is there a typo where baz.sk was meant instead of bar.sk?

- Maybe there is something basic that I have missed, but I'm confused by the sklib files containing skip sources and then the invalidation logic using modification times of files. How are those two going to get along?

- Also on invalidation, can you explain a bit about why it has anything to do with libraries and not just source files at all? I'm missing that.

beauby:
> Some comments and questions prior to digging through the code in detail:
> 
> * If only for curiousity, is the choice of the ranlib format a free choice or are there external considerations that make it a good idea?

The idea behind using the ranlib format is that it can be passed directly to clang for linking. With hindsight, it is nice but not really necessary.

> * Any comments one the expectations for what will eventually happen with the dependency cycle between std and skfs?

Ideally, `skfs` should depend on `std`, but not the other way around. Currently, there are things related to exceptions in `std` that make use of some `SKStore` types, plus some C functions defined in the runtime. It should be straightforward to split the two but I haven't gotten to it yet.

> * Minor, but regarding the name of `--dep`, is it maybe too generic since these are library dependencies, while there might at some point be other sorts of dependencies, such as data to be packed into the binaries. Also, something like `--lib` might go better since it searches in the path set up by `-L`.

Agreed, I am a bit uneasy about `--dep`. I thought about `--lib` but it seems confusing to have both `--lib` and `-l` with different meanings. 

> * I wonder what happens if some lib is both a direct dep as well as a dep of a dep, and it is mistakenly declared as a dep only of the dep. Do the uses of it as a direct dep lead to resolution failures? Ideally they would, but maybe not something to insist on in the first step.

Currently we do not have a way to do that AFAIK. Once we namespace the dependencies we can make only explicit deps namespaces available, which would allow this.

> * I don't understand "For instance, skc foo.sk bar.sk baz.sk --data state.db followed by skc foo.sk bar.sk foobar.sk --data state.db would get rid of the analysis
>   of bar.sk (and that of foo.sk if its mtime has changed between
>   the two invocations).". Is there a typo where baz.sk was meant instead of bar.sk?

Indeed, apologies for that.

> * Maybe there is something basic that I have missed, but I'm confused by the sklib files containing skip sources and then the invalidation logic using modification times of files. How are those two going to get along?

Ranlib archives have a `mtime` attributes for files, so this should not be an issue. Or did you have something else in mind?

> * Also on invalidation, can you explain a bit about why it has anything to do with libraries and not just source files at all? I'm missing that.

This is an important point: the idea is that when building a package, we first generate a DAG (this will be done by skargo), and then build dependencies in topological order. Suppose the DAG has two disconnected sub-graphs: while building the second sub-graph, we do not have any knowledge of the first one, yet we do not want to discard analysis for it.

jberdine:
> > * Minor, but regarding the name of `--dep`, is it maybe too generic since these are library dependencies, while there might at some point be other sorts of dependencies, such as data to be packed into the binaries. Also, something like `--lib` might go better since it searches in the path set up by `-L`.
> 
> Agreed, I am a bit uneasy about `--dep`. I thought about `--lib` but it seems confusing to have both `--lib` and `-l` with different meanings.

Personally I'm more concerned that `-l` and `-L` don't talk about the same sort of libraries, which I expect to be surprising to people.

A related question is whether it is necessary to have separate/duplicate names for `--sklib-name` and for the actual file name. Is it really necessary to have multiple names, for what seems to be the same thing? OCaml has multiple levels of names for libraries and packages, etc. and it is one of the biggest pain points. And if the names could be unified, maybe it would work to use `skc foo.sklib` instead of `skc --dep foo` and thereby get rid of `--dep` completely, and use something like `-I` for extending the path searched for such files, leaving `-l` and `-L` for the system libraries.

> > * I wonder what happens if some lib is both a direct dep as well as a dep of a dep, and it is mistakenly declared as a dep only of the dep. Do the uses of it as a direct dep lead to resolution failures? Ideally they would, but maybe not something to insist on in the first step.
> 
> Currently we do not have a way to do that AFAIK. Once we namespace the dependencies we can make only explicit deps namespaces available, which would allow this.

So, paraphrasing to check my understanding: it will initially be possible to make the mistake of using a transitive dep that is not declared as an explicit dep, but later when a layer of namespacing is added, this mistake will be catchable.

> > * Maybe there is something basic that I have missed, but I'm confused by the sklib files containing skip sources and then the invalidation logic using modification times of files. How are those two going to get along?
> 
> Ranlib archives have a `mtime` attributes for files, so this should not be an issue. Or did you have something else in mind?

Ah, ok thanks, I missed that.

> > * I don't understand "For instance, skc foo.sk bar.sk baz.sk --data state.db followed by skc foo.sk bar.sk foobar.sk --data state.db would get rid of the analysis
> >   of bar.sk (and that of foo.sk if its mtime has changed between
> >   the two invocations).". Is there a typo where baz.sk was meant instead of bar.sk?
> 
> Indeed, apologies for that.
> 
> > * Also on invalidation, can you explain a bit about why it has anything to do with libraries and not just source files at all? I'm missing that.
> 
> This is an important point: the idea is that when building a package, we first generate a DAG (this will be done by skargo), and then build dependencies in topological order. Suppose the DAG has two disconnected sub-graphs: while building the second sub-graph, we do not have any knowledge of the first one, yet we do not want to discard analysis for it.

There is still something I'm missing. It makes sense to me that we want to avoid discarding what has been computed from source files that have not changed in the meantime. But in the descriptions of the behaviors for both before and after this change, there is mention of discarding computations for files that have not changed but weren't passed as inputs for the latest compiler invocation. Why is such discarding ever necessary? That is, why does it matter whether or not a file has been given as an input, if it has not changed, then the previous computation is still valid, no?

beauby:
> > > * Minor, but regarding the name of `--dep`, is it maybe too generic since these are library dependencies, while there might at some point be other sorts of dependencies, such as data to be packed into the binaries. Also, something like `--lib` might go better since it searches in the path set up by `-L`.
> > 
> > 
> > Agreed, I am a bit uneasy about `--dep`. I thought about `--lib` but it seems confusing to have both `--lib` and `-l` with different meanings.
> 
> Personally I'm more concerned that `-l` and `-L` don't talk about the same sort of libraries, which I expect to be surprising to people.

I agree on that point. We could have `-l skip=foo`/`-l skip=foo=/path/to/foo/sklib` and `-L skip=/path/to/libs/` (considering we would also have `-l static=foo`/`-L native=/path/to/sys/lib` for disambiguation). I'm not too fond of the `...=...=...` thing, but I could live with it. What do you think?

> A related question is whether it is necessary to have separate/duplicate names for `--sklib-name` and for the actual file name. Is it really necessary to have multiple names, for what seems to be the same thing? OCaml has multiple levels of names for libraries and packages, etc. and it is one of the biggest pain points. And if the names could be unified, maybe it would work to use `skc foo.sklib` instead of `skc --dep foo` and thereby get rid of `--dep` completely, and use something like `-I` for extending the path searched for such files, leaving `-l` and `-L` for the system libraries.

That one got me thinking for a while. I, too, dislike the seemingly redundant aspect of specifying `--sklib-name foo` on top of `-o foo.sklib`.
Having an explicit `--sklib-name` allows later invocations of `skc` to inspect the name of the provided library. This provides a few benefits:
  - for now (as we do not support multiple versions of the same library being used (mainly for namespacing and linking issues)), it allows ensuring that no two versions of the same `sklib` is provided/required;
  - when/if we support namespaced packages, knowing the actual library name can be used for disambiguation. For instance, if `foo.sklib` and `bar.sklib` both define a `FooBar` module (at the moment, the result is uncertain: in the best case the two modules are merged, but definitions can also conflict), we could extend the `module alias` construction to support `module alias FB = foo@FooBar` (syntax up for debate). Without an explicit `--sklib-name`, we're left to rely on build time choices to ensure code correctness (i.e. if we later decide to build lib `foo` as `libfoo-wasm32-v1.3.0.sklib` rather than `foo.sklib`, the above `module alias` changes meaning), which seems really bad.
  - We can support `--dep foo`, and have it being resolved to the correct `.sklib` file without specifying the full path, which in turn allows for having an implicit `--dep std` by default (and therefore being able to do `skc foo.sk -o foo` rather than requiring the user to know the path of the system installed `libstd.sklib`). This also allows specifying only the direct dependencies of a package, rather than all transitive dependencies (unless we start storing full paths to dependencies in `.sklib`s, which means they cannot be moved anymore). For the resolution, we could rely on a convention on file names for this, but it would be brittle, especially when dealing with multiple versions of a same package:
  - while we currently do not support having the same package with two different versions being used when building the same target, it is not uncommon in general that package dependencies (used to build the actual `.sklib` or binary) and build dependencies (used only to build the custom `build.sk` script that will generate artifacts such as C libraries or resources) to differ in version requirements. In that case, relying on a file name convention would likely break.
  - Regarding package-level invalidation of source file analysis (detailed below), it would not be possible without a way to tell whether two `.sklib` files represent the same library (so nothing would get invalidated unless the two invocations share the exact same output path).

> > > * I wonder what happens if some lib is both a direct dep as well as a dep of a dep, and it is mistakenly declared as a dep only of the dep. Do the uses of it as a direct dep lead to resolution failures? Ideally they would, but maybe not something to insist on in the first step.
> > 
> > 
> > Currently we do not have a way to do that AFAIK. Once we namespace the dependencies we can make only explicit deps namespaces available, which would allow this.
> 
> So, paraphrasing to check my understanding: it will initially be possible to make the mistake of using a transitive dep that is not declared as an explicit dep, but later when a layer of namespacing is added, this mistake will be catchable.

Correct.
> > > * Also on invalidation, can you explain a bit about why it has anything to do with libraries and not just source files at all? I'm missing that.
> > 
> > 
> > This is an important point: the idea is that when building a package, we first generate a DAG (this will be done by skargo), and then build dependencies in topological order. Suppose the DAG has two disconnected sub-graphs: while building the second sub-graph, we do not have any knowledge of the first one, yet we do not want to discard analysis for it.
> 
> There is still something I'm missing. It makes sense to me that we want to avoid discarding what has been computed from source files that have not changed in the meantime. But in the descriptions of the behaviors for both before and after this change, there is mention of discarding computations for files that have not changed but weren't passed as inputs for the latest compiler invocation. Why is such discarding ever necessary? That is, why does it matter whether or not a file has been given as an input, if it has not changed, then the previous computation is still valid, no?

Clashing definitions are the main issue: if `class Foo` is defined in file `a.sk` and then `class Foo` is redefined in file `b.sk`, then there is no way to discard the previous definition unless we invalidate stale files.

One could think overriding previous definitions would be enough, but things like class extensions and sub typing would still cause issues: if we do not discard analysis for stale files, we can have stale sub classes definitions:
```
// a.sk
base class Foo {
  children = 
  | Bar1(Int)
  | Bar2(String)
}
// invoking `skc a.sk`: all good

// b.sk
class Bar3(Array<Int>) extends Foo
// invoking `skc b.sk`: all good

// c.sk
fun foo(f: Foo): void {
  f match {
  | Bar1(x) -> ...
  | Bar2(x) -> ...
  }
}
// invoking `skc c.sk`: missing pattern
// no way to discard the definition of `Bar3`, apart from killing `state.db`.
```

jberdine:
> > > > * Minor, but regarding the name of `--dep`, is it maybe too generic since these are library dependencies, while there might at some point be other sorts of dependencies, such as data to be packed into the binaries. Also, something like `--lib` might go better since it searches in the path set up by `-L`.
> > > 
> > > 
> > > Agreed, I am a bit uneasy about `--dep`. I thought about `--lib` but it seems confusing to have both `--lib` and `-l` with different meanings.
> > 
> > 
> > Personally I'm more concerned that `-l` and `-L` don't talk about the same sort of libraries, which I expect to be surprising to people.
> 
> I agree on that point. We could have `-l skip=foo`/`-l skip=foo=/path/to/foo/sklib` and `-L skip=/path/to/libs/` (considering we would also have `-l static=foo`/`-L native=/path/to/sys/lib` for disambiguation). I'm not too fond of the `...=...=...` thing, but I could live with it. What do you think?

First, is more done with the system libraries than just to pass them on through to the system linker? I haven't checked closely, but expect that they end up involved with ranlib, etc. If not, and they are only passed on, then maybe the clearest thing to do would be to add a mechanism to pass flags to the system linker. Maybe something like a `--linkopts="-L /usr/local/lib -lfoo"` argument that causes `-L /usr/local/lib -lfoo` to be added to the system linker invocation. Then the skip tools could just deal with skip code and libs.

> > > > * Also on invalidation, can you explain a bit about why it has anything to do with libraries and not just source files at all? I'm missing that.
> > > 
> > > 
> > > This is an important point: the idea is that when building a package, we first generate a DAG (this will be done by skargo), and then build dependencies in topological order. Suppose the DAG has two disconnected sub-graphs: while building the second sub-graph, we do not have any knowledge of the first one, yet we do not want to discard analysis for it.
> > 
> > 
> > There is still something I'm missing. It makes sense to me that we want to avoid discarding what has been computed from source files that have not changed in the meantime. But in the descriptions of the behaviors for both before and after this change, there is mention of discarding computations for files that have not changed but weren't passed as inputs for the latest compiler invocation. Why is such discarding ever necessary? That is, why does it matter whether or not a file has been given as an input, if it has not changed, then the previous computation is still valid, no?
> 
> Clashing definitions are the main issue: if `class Foo` is defined in file `a.sk` and then `class Foo` is redefined in file `b.sk`, then there is no way to discard the previous definition unless we invalidate stale files.

Ok, I see now, thanks. Skip the language has an inherent closed-world view, and so the build and library tooling needs to have some concept outside the language itself of the "scope" of things that go into each closed-world. Naively, each root in the DAG of dependencies could determine its own separate closed-world, and each could be built from everything it transitively depends on, but doing that would be suboptimal due to redundant computation for the files that are dependencies of multiple roots. This makes me think that the right solution is to implement this with one combined persistent state for the entire DAG, and use skstore mappers that call the compiler frontend instead of invoking skc processes. Does that make sense? Did you think about that before and did it look to be very involved?

beauby:
> First, is more done with the system libraries than just to pass them on through to the system linker? I haven't checked closely, but expect that they end up involved with ranlib, etc. If not, and they are only passed on, then maybe the clearest thing to do would be to add a mechanism to pass flags to the system linker. Maybe something like a --linkopts="-L /usr/local/lib -lfoo" argument that causes -L /usr/local/lib -lfoo to be added to the system linker invocation. Then the skip tools could just deal with skip code and libs.

I like this idea. I need to think a bit more about possible edge cases, but it seems like a satisfying and workable solution. 
For reference, what we need to enable is
- passing linking args to the system linker (as you mentioned) : `-lfoo` and `-L/path/` mainly,
- passing (generated) static libraries to `skc` so that they get embedded into the `.sklib` (and the `.sklib` can then further be passed as is to the system linker). Those should be at known locations, so maybe lookup does not need to be involved, and we can rely on `skargo` to pass the explicit path to `skc`,
- passing (direct) dependencies to `.sklib` libraries. This one needs lookup, but if we get system libraries out of the way, we can fully use `-l` and `-L` at `skc` level (vs `--linkopts="-lfoo -L/path/to/foo/`).

beauby:
> Ok, I see now, thanks. Skip the language has an inherent closed-world view, and so the build and library tooling needs to have some concept outside the language itself of the "scope" of things that go into each closed-world. Naively, each root in the DAG of dependencies could determine its own separate closed-world, and each could be built from everything it transitively depends on, but doing that would be suboptimal due to redundant computation for the files that are dependencies of multiple roots. This makes me think that the right solution is to implement this with one combined persistent state for the entire DAG, and use skstore mappers that call the compiler frontend instead of invoking skc processes. Does that make sense? Did you think about that before and did it look to be very involved?

This is, to some extent, what I envisioned about the next step for namespaced libraries. I am not sure what CLI API you have in mind for providing the whole DAG to a single `skc` invocation, but several successive `skc` invocations do not seem to be an issue as long as things are cached (and namespaced) properly across runs. We can discuss this offline a bit if you feel it would be more efficient.

.circleci/base.yml
```diff
@@ -22,8 +22,8 @@ jobs:
           no_output_timeout: 30m
           command: |
             mkdir -p ~/test-results
-            cd compiler && make STAGE=B
-            PATH=./stageB/bin:$PATH skargo test --jobs 8 --junitxml ~/test-results/skc.xml
+            cd compiler && make
```
jberdine: it might be good to leave `STAGE=1` to avoid silently relying on the default here and then setting `PATH` explicitly on the next line

compiler/Makefile
```diff
@@ -10,54 +10,32 @@ libdir   ?= $(exec_prefix)/lib
...
-	cp $< $@
-
-stage%/lib/libskip_runtime32.bc: runtime/libskip_runtime32.bc
-	mkdir -p $(@D)
-	cp $< $@
+	$(MAKE) $@/bin/skc $@/bin/skfmt $@/bin/skargo $@/lib/libstd-x86_64-pc-linux-gnu.sklib $@/lib/libstd-wasm32-unknown-unknown.sklib
 
-stage%/lib/skip_preamble64.ll: preamble/preamble64.ll
+# TODO: Build fresh stdlib for stage >=1?
+stage%/lib/libstd-x86_64-pc-linux-gnu.sklib: bootstrap/libstd-x86_64-pc-linux-gnu.sklib
```
jberdine: Could you explain a bit how libstd....sklib is built? I'm only seeing a dependency on the bootstrap version, but I haven't worked out how it gets rebuilt. I see the compile_sklib function in compile.sk, but I don't see how the native objects get compiled.

I'm also worried by the x86_64 stuff. There is native object code in the sklib files, right? I expect that we will need an arm64 version too, and have logic to select between then, or else it will be broken for all of us using the docker on macs workflow.

jberdine (commented):
Looking good. A couple questions and comments as I go. Also, it would help if you updated the PR summary for the latest iteration of the CLI.

compiler/Makefile
beauby: Good point. This is a bit of a magic leap for now: the (future) normal way of generating `libstd-*.sklib` will be using `skargo` (once it handles sklibs, cf. #163). For now, it is manually generated by invoking `make`+`skc` directly. This will remain a fallback option in the absence of a usable bootstrap (either on new platforms or in the possible case of botched bootstrap). I will update the PR to make it an explicit `make` target to document it.

Regarding `arm64`, you are correct indeed, will add an `arm64` version.

jberdine: Ok, sounds good. Having a `make` target would be good, and useful in additional to documentation I think. For info, I'm thinking about how I will need to adapt the current workflow I use to test each commit on a branch. Currently I do the following:
```
GIT_SEQUENCE_EDITOR=: git rebase -ix "dx 'set -a; CXXFLAGS=\"-O0 -mllvm -inline-threshold=0\"; SKARGO_PROFILE=dev; STAGE=B; PATH=$PWD/compiler/stageB/bin:$PWD/compiler/stageA/bin:$PATH; set +a; make clean && make -C compiler clean install && make test'" main
```
where `dx` is my script to execute a command in a running docker container. Hopefully with your changes I will be able to add a step to build the runtime+stdlib and change `install` to `promote`.

compiler/Makefile
```diff
@@ -10,54 +10,40 @@ libdir   ?= $(exec_prefix)/lib
...
+bootstrap/libstd.sklib:
+	PATH=$(realpath stage0/bin/):$(PATH) OUT_DIR=$(realpath ./bootstrap) $(MAKE) -C ../prelude bootstrap
 
-stage0/bin/skargo: bootstrap/skargo_out64.ll stage0/lib/libskip_runtime64.a stage0/lib/libbacktrace.a
+stage0/bin/skc: bootstrap/skc_out64.ll bootstrap/libskip_runtime64.a bootstrap/libbacktrace.a
 	mkdir -p $(@D)
 	clang++ -no-pie $(CXXFLAGS) -o $@ $^ -lpthread -g3
 
-stage0/bin/skfmt: bootstrap/skfmt_out64.ll stage0/lib/libskip_runtime64.a stage0/lib/libbacktrace.a
+stage0/bin/skargo: bootstrap/skargo_out64.ll stage0/lib/libstd.sklib
```
jberdine, line 42: Just to check, the dependencies here are different because bootstrap/skc_out64.ll contains the bitcode not only of skc but also that of the stdlib and preamble, right? And this skc pattern is the reason that there are targets above for bootstrap/libskip_runtime64.a and bootstrap/libbacktrace.a separate from bootstrap/libstd.sklib?

compiler/Makefile
```diff
@@ -10,54 +10,40 @@ libdir   ?= $(exec_prefix)/lib
...
+bootstrap/%.ll: bootstrap/%.ll.gz
+	gunzip -c $< > $@
 
-stage%/lib/skip_preamble64.ll: preamble/preamble64.ll
-	mkdir -p $(@D)
-	cp $< $@
+.PHONY: bootstrap/libskip_runtime64.a bootstrap/libbacktrace.a
+bootstrap/libskip_runtime64.a:
+bootstrap/libbacktrace.a:
+	OUT_DIR=$(realpath ./bootstrap) $(MAKE) -C ../prelude
```
jberdine, line 32: I may be getting confused between this and the skargo 2 PR, but this looks like it is putting host native object code in the bootstrap directory. Am I not understanding, or aren't we going to need the bootstrap dir to contain both x86_64 and arm64 native, as well as wasm, versions of the runtime?

compiler/Makefile
beauby: The dependencies are different here for bootstrap reasons: in order to generate a .sklib (which would contain the object files of `libskip_runtime64.a` and `libbacktrace.a`), we need `skc`, but in order to build `skc` from `skc_out64.ll`, we need `libskip_runtime64.a` and `libbacktrace.a`.

jberdine: Maybe what I really mean is, since there is bootstrap/libstd.sklib, couldn't stage0/bin/skc be built from that? That way skc.ll would not need to be expanded with the stdlib ll, and maybe bootstrap/libskip_runtime64.a bootstrap/libbacktrace.a would not be needed separate from bootstrap/libstd.sklib.

beauby: A bit of context: `bootstrap/libstd.sklib` is currently not being checked in, and is being rebuilt from scratch. This changes from a previous iteration. The reason is that generating both the amd64 and arm64 versions would be a pain in practice.

jberdine: And thinking of the scenario where we need to rebuild the runtime from C source but otherwise use the bootstrap, since sklib is in ar format, the bitcode inside could be extracted and used with freshly built native .a files. Not a super easy workflow, but possible and not something that we expect to do more than a few times ever.

jberdine: Ah, ok, I see. Nevermind my last comments then.

PR #163, opened 2024-03-15
Skargo 2.0

beauby:
This commit introduces skargo 2.0, a rewrite of skargo leveraging                                                  
sklibs to improve developer experience and pave the way to building                                                
shared skip libraries with FFI.                                                                                    
Most changes should be invisible to the user, but some CLI                                                         
improvements were made: proper support for `-v`/`-vv`, with an accent                                              
on providing all the information necessary to debug builds (one                                                    
notable issue was that environment variables passed to `build.sk` were                                             
hidden, they are now explicitly logged when using `-vv`).                                                          
Internally, skargo now handles build/dev/regular dependencies in a                                                 
clean way by building an explicit dependency graph (which opens the                                                
possibility of parallel invocations of `skc`). Most of the logic                                                   
resides in `src/build_context.sk` for building the graph, and                                                      
`src/build_runner.sk` for actually running the build steps.

Based on #162.


prelude/src/stdlib/other/Chars.sk
```diff
@@ -135,6 +135,19 @@ fun octDigitsToInt(digits: String): Int {
   digits.foldl((current, digit) -> current * 8 + hexDigitToInt(digit), 0)
 }
 
+fun intToOctDigits(n: Int): String {
+  digits = mutable Vector[];
+  while (n > 0) {
+    digits.push(hexDigitChars[n.and(0x8)]);
+    !n = n.ushr(8)
```
mbouaziz: Did you mean
```
    digits.push(hexDigitChars[n.and(7)]);
    !n = n.ushr(3)
```

beauby: Hum yes. I hastily rewrote some code that was initially using modulo/div. This function is only used (so far) to serialize the file mode in `arparser`, which we do not inspect so it went unnoticed. Thanks!

beauby: Fixed in #23.

gregsexton:
Could you describe a little what has changed? And particularly what changes for users?

beauby:
> Could you describe a little what has changed? And particularly what changes for users?

Of course, just waiting on #162 to settle first.

compiler/Makefile
```diff
@@ -67,6 +67,13 @@ stageC/bin/skc stageC/bin/skfmt stageC/bin/skargo: stageB
 	PATH=stageB/bin/:$(PATH) skargo build -r --bins --out-dir stageC/bin --target-dir stageC/bin
 	PATH=stageB/bin/:$(PATH) skargo build -r --bins --out-dir stageC/bin --target-dir stageC/target --manifest-path ../skargo/Skargo.toml
 
+.PHONY: promote
+promote: stage$(STAGE)
+	cp stage1/target/host/release/deps/skc-*.ll bootstrap/skc_out64.ll && gzip -f bootstrap/skc_out64.ll
```
jberdine: these hardcode stage1

skargo/src/commands/check.sk
```diff
@@ -0,0 +1,68 @@
...
+      ),
+    execCheck,
+  )
+}
+
+fun execCheck(gctx: GlobalContext, args: Cli.ParseResults): void {
+  // TODO: Reinstate config for format-on-check?
+  // if (env.config.formatOnCheck) {
+  //   execFmt(args, gctx)
+  // };
```
jberdine: is supporting this problematic?

beauby: Not really, I just had to ditch the config file logic temporarily, I'll reinstate it.

compiler/Makefile
beauby: Good catch, I'll change it to `stage$(STAGE)` (though I doubt anyone bothers with stage 2 in general).

jberdine: I build stage 2 reasonably often, but not so much in order to promote it.

gregsexton:
Anything to know as a user? Do I use it exactly as before?

beauby:
> Anything to know as a user? Do I use it exactly as before?

Same as before, with the exception of verbosity levels (default, -v, -vv, --quiet)

PR #166, opened 2024-03-19
[sql] Evaluate rather than (fail to) compile arg-less VId and VSeqNum

### LINKED ISSUE #164: Cannot insert a generated id using a select
I want to do this:

```
INSERT INTO pk_single_row_hist SELECT local_sequence_number(), datetime(), client, value FROM pk_single_row;
```

But I cannot.

```
Error: line 1, character 0:
local_sequence_number expects a string literal
```

This is true of id() and some other things that seem to behave differently if they're an expression in a select's col expression list or an insert's value list.

jberdine:
This results in them being evaluated once per row rather than being
compiled once per query. As a result queries such as
```
SELECT local_sequence_number(), id(), v FROM t1;
```
will yield distinct values for each row of t1.

closes #164

### DIFF
--- a/sql/src/SqlCAst.sk
+++ b/sql/src/SqlCAst.sk
@@ -469,6 +469,9 @@ class CStrftime(
 
 class COn(P.JoinKind, CExpr<Int>) extends CExpr<Int>
 
+class CISeqNum(Bool) extends CExpr<Int>
+class CSId() extends CExpr<String>
+
 /*****************************************************************************/
 /* JSON primitives */
 /*****************************************************************************/
--- a/sql/src/SqlCompile.sk
+++ b/sql/src/SqlCompile.sk
@@ -995,8 +995,7 @@ mutable class Compiler{
         error(this.pos, "Unknown local_sequence_number name " + varName)
       };
       CIExpr(CILiteral(globals[varName]))
-    | P.VSeqNum(_, None()) ->
-      error(this.pos, "local_sequence_number expects a string literal")
+    | P.VSeqNum(isPositive, None()) -> CIExpr(CISeqNum(isPositive))
     | P.VId(Some(varName)) ->
       namedIDs = context.getGlobal(NAMEDIDS) match {
       | None() -> error(this.pos, "Unknown id " + varName)
@@ -1006,7 +1005,7 @@ mutable class Compiler{
         error(this.pos, "Unknown id " + varName)
       };
       CSExpr(CSLiteral(namedIDs[varName]))
-    | P.VId(None()) -> error(this.pos, "id expects a string literal")
+    | P.VId(None()) -> CSExpr(CSId())
     | P.VPermission(arg) ->
       this.compileExpr(context, arg) match {
       | CSExpr(CSLiteral(str)) ->
--- a/sql/src/SqlExpr.sk
+++ b/sql/src/SqlExpr.sk
@@ -291,6 +291,8 @@ class ExprEvaluator(
           "ok"
         })
       })
+
+    | CSId() -> ADef(Ksuid::create().toString())
     }
   }
 
@@ -949,6 +951,13 @@ class ExprEvaluator(
       | AUndef() -> AUndef()
       | ANull() -> ANull()
       | ADef(s) -> ADef(s.length())
+      }
+
+    | CISeqNum(isPositive) ->
+      if (isPositive) {
+        ADef(SKStore.genSym(0))
+      } else {
+        ADef(-SKStore.genSym(0))
       }
     }
   }
--- a/sql/test/unit/insert_select_id.sql
+++ b/sql/test/unit/insert_select_id.sql
@@ -1,0 +1,8 @@
+CREATE TABLE t1(i INTEGER, t TEXT, v INTEGER);
+
+INSERT INTO t1 VALUES (local_sequence_number(), id(), 1);
+INSERT INTO t1 VALUES (local_sequence_number(), id(), 2);
+INSERT INTO t1 VALUES (local_sequence_number(), id(), 3);
+INSERT INTO t1 SELECT local_sequence_number(), id(), v FROM t1;
+SELECT i FROM t1;
+SELECT t FROM t1;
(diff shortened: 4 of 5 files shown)


jberdine:
Note that other "functions" that might appear in such `SELECT`s, such as `datetime()`, still get evaluated only once per query, not once per row. So e.g.
```
SELECT local_sequence_number(), id(), datetime(), v FROM t1;
```
will produce rows which all have exactly the same 3rd column.

jberdine:
FTR, the test output is:
```
$ cd sql; cat test/insert_select_id.sql | ../build/skdb
------------------------------------------------------------
1|uid|1
9|f7sKiTYD65iPunMPdjC2WNGyMa9|2
13|f7sKiQGbqBt_k54k2cvriONkAcE|3
------------------------------------------------------------
25|f7sKiNfRJ9FDl0HcCGKPRsPtnCE|1
26|f7sKiP9TNdH4Mib1Qov-DxwwxQ9|2
27|f7sKiJRADlcbHt3iLz3TlDyP34D|3
------------------------------------------------------------
1|uid|1
9|f7sKiTYD65iPunMPdjC2WNGyMa9|2
13|f7sKiQGbqBt_k54k2cvriONkAcE|3
33|f7sKiKJr22_zN7doV712LkdGKU0|1
34|f7sKiQQtpiI1WM4qoyaxiiHyrS5|2
35|f7sKiTU-ZxUVlahWs_G7ns2z-G1|3
------------------------------------------------------------
```

gregsexton:
> still get evaluated only once per query, not once per row

[To be clear, I think this is out of scope for this pr.]
Is this how we evaluate functions? I'm not sure this is correct. I would expect them to be per row. 

And then for those functions known to produce a constant value (such as perhaps `datetime`) this can be folded in before evaluation. This is how a lot of engines work. They have a query rewrite phase before the planner works out how to execute the query.

sql/test/insert_select_id.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE t1(i INTEGER, t TEXT, v INTEGER);
+
+INSERT INTO t1 VALUES (1, 'uid', 1);
+INSERT INTO t1 VALUES (local_sequence_number(), id(), 2);
```
gregsexton: I assume we have somewhere where we do an insert like this but using a named key for e.g. id and a select with that name within a txn?

bennostein:
> FTR, the test output is:
> 
> ```
> $ cd sql; cat test/insert_select_id.sql | ../build/skdb
> ------------------------------------------------------------
> 1|uid|1
> 9|f7sKiTYD65iPunMPdjC2WNGyMa9|2
> 13|f7sKiQGbqBt_k54k2cvriONkAcE|3
> ------------------------------------------------------------
> 25|f7sKiNfRJ9FDl0HcCGKPRsPtnCE|1
> 26|f7sKiP9TNdH4Mib1Qov-DxwwxQ9|2
> 27|f7sKiJRADlcbHt3iLz3TlDyP34D|3
> ------------------------------------------------------------
> 1|uid|1
> 9|f7sKiTYD65iPunMPdjC2WNGyMa9|2
> 13|f7sKiQGbqBt_k54k2cvriONkAcE|3
> 33|f7sKiKJr22_zN7doV712LkdGKU0|1
> 34|f7sKiQQtpiI1WM4qoyaxiiHyrS5|2
> 35|f7sKiTU-ZxUVlahWs_G7ns2z-G1|3
> ------------------------------------------------------------
> ```

Could we get this hooked into `make test-native`? I get that it'll change each time and we don't want to be too sensitive to specific sequence numbers, but perhaps a `grep` into `wc -l` to check that we get the right number of rows of the right shape?

EDIT: could also use `uniq` to check that we're evaluating different sequence numbers per row, now that I think of it...

gregsexton (commented):
Looks like a good step in the right direction. Thanks a lot for fixing this quickly. :)

sql/test/insert_select_id.sql
jberdine: Like https://github.com/SkipLabs/skdb/blob/main/sql/test/unit/test_id.sql, https://github.com/SkipLabs/skdb/blob/main/sql/test/unit/test_seqnum.sql, and https://github.com/SkipLabs/skdb/blob/main/sql/test/unit/test_seqnum2.sql, or something more?

jberdine:
> Could we get this hooked into `make test-native`? I get that it'll change each time and we don't want to be too sensitive to specific sequence numbers, but perhaps a `grep` into `wc -l` to check that we get the right number of rows of the right shape?
> 
> EDIT: could also use `uniq` to check that we're evaluating different sequence numbers per row, now that I think of it...

Yes, I agree that the testing setup is pretty limited. Have to think a bit on a better approach. Anything with ids is time-sensitive, and maybe we don't want to be manually hacking scripts to chop out the stable parts, and having to consider that the volatile parts are sometimes the point of the test...

jberdine:
> > still get evaluated only once per query, not once per row
> 
> [To be clear, I think this is out of scope for this pr.] Is this how we evaluate functions? I'm not sure this is correct. I would expect them to be per row.
> 
> And then for those functions known to produce a constant value (such as perhaps `datetime`) this can be folded in before evaluation. This is how a lot of engines work. They have a query rewrite phase before the planner works out how to execute the query.

Most functions, etc. get simplified somewhat in the compilation phase and get lowered to a form that the evaluator then evaluates once per row. Some things get simplified all the way to literal values though, like`id('key')` will be looked up once.

bennostein:
> Have to think a bit on a better approach.

fwiw I was thinking along the lines of what's in `sql/unit_tests.sh`, which use various unix utility pipelines to check what they want to check.

e.g.

```
if cat <that-file> | $SKDB | grep <strip out non-sequence-number stuff> | uniq | wc -l | grep -q <the-right-number-of-lines> ; then
    pass "SEQUENCE NUMBERS"
else
    fail "SEQUENCE NUMBERS"
fi
```

sql/test/insert_select_id.sql
gregsexton: Yep, perfect.

PR #167, opened 2024-03-19
Configure and run ktfmt over kotlin sources

bennostein:
This adds gradle targets to autoformat kotlin sources under `./sql/server` using `ktfmt` and `spotless`.

I also formatted everything with
```
(cd ./sql/server &&  ./gradlew :spotlessApply :core:spotlessApply :dev:spotlessApply)
```

There's some minor churn as we last ran ktfmt on 0.42 and I used the latest 0.47 here.

@gregsexton, I'm not sure what (if anything) you're using for autoformatting. I'm not picky about what we use as long as it's something -- happy to drop this PR if you've got some other conflicting setup in place that you would rather point me to :)

### DIFF
--- a/sql/server/build.gradle.kts
+++ b/sql/server/build.gradle.kts
@@ -1,8 +1,3 @@
 plugins {
   id("org.jetbrains.kotlin.jvm") version "1.9.0"
-  id("com.diffplug.spotless") version "6.12.1"
 }
-
-repositories { mavenCentral() }
-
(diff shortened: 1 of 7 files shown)


gregsexton:
LGTM. 

Spotless/ktfmt is what I use. Introduced here https://github.com/SkipLabs/skdb/commit/601041de03b0538ef295680f7909abeaf2d2ec32. But it didn't survive the refactor in to modules/repos.

Thanks for sorting this out :)

PR #168, opened 2024-03-20
[cli] Add option to avoid failing on invalid arg.

beauby:
The previous behavior of `ParseResults` is to raise when requesting the value of an argument that was not defined. This is a good thing in general, as inspecting value of an option that was not specified is often a sign of a bug. In certain instances, however, it is desirable not to raise and return `None()` instead for `maybeGetString()` etc. It is the case when a helper function builds an object from `ParseResults` (that may originate from different `Command` definitions) with default values for undefined arguments.

### DIFF
--- a/cli/src/parser.sk
+++ b/cli/src/parser.sk
@@ -37,58 +37,96 @@ class ParseResults{
   subcommand: ?String = None(),
   error: ?ArgumentError = None(),
 } {
-  fun getBool(name: String): Bool {
-    this.values[name] match {
-    | BoolValue(v) -> v
+  fun maybeGetBool(name: String, fail_on_undefined_arg: Bool = true): ?Bool {
+    this.getValue(name, fail_on_undefined_arg) match {
+    | BoolValue(v) -> Some(v)
+    | MissingValue() -> None()
     | _ -> invariant_violation(`Argument ${name} is not boolean`)
     }
   }
-
-  fun maybeGetString(name: String): ?String {
-    this.values.maybeGet(name) match {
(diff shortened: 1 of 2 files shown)


cli/src/parser.sk
```diff
@@ -37,61 +37,71 @@ class ParseResults{
   subcommand: ?String = None(),
   error: ?ArgumentError = None(),
 } {
-  fun getBool(name: String): Bool {
-    this.values[name] match {
+  fun getBool(name: String, fail_on_invalid_arg: Bool = true): Bool {
```
mbouaziz: Should `fail_on_invalid_arg` be named `fail_on_undefined_arg`?

cli/src/parser.sk
```diff
@@ -37,61 +37,71 @@ class ParseResults{
...
+    this.getValue(name, fail_on_invalid_arg) match {
     | BoolValue(v) -> v
+    | MissingValue() -> false
```
mbouaziz: I haven't looked at how this interacts with the rest of `cli` but I suspect you may not always want `false` as the default. Should `getBool` take an optional `default: Option<Bool> = None()` instead of `fail_...`?

mbouaziz: `getBool` is the only one for which you have a default

mbouaziz: Oh, I see that `BoolArg` already defaults its default value to `false`...

beauby: I updated the API so that the `maybeGetFoo()` methods take a `fail_on_undefined_arg` boolean, and the `getFoo()` methods take an optional `default`.

PR #169, opened 2024-03-20
[cli] Allow multiple short flags in one.

beauby:
Example: `-abc` is now equivalent to `-a -b -c` as long as `a`, `b`,                                        
and `c` are all boolean flags.                                                                              
Additionally, the last flag can be non-boolean: `-kj2` is now                                               
equivalent to `-k -j 2`.

Based on #173.

### DIFF
--- a/cli/src/parser.sk
+++ b/cli/src/parser.sk
@@ -319,22 +319,12 @@ mutable class Parser(
         | Failure(err) ->
           !error = Some(err);
           break void
-        | Success((arg, value)) ->
-          arg match {
-          | BoolArg _ ->
-            if (!this.values.containsKey(arg.name)) {
-              this.values.set(arg.name, BoolValue(!str.startsWith("--no-")))
-            } else {
-              !error = Some(DuplicateValueError(arg.name));
-              break void
-            }
-          | _ ->
-            updateValue(this.values, arg, value) match {
-            | Failure(err) ->
-              !error = Some(err);
-              break void
-            | Success _ -> void
-            }
+        | Success(args) ->
+          this.updateValues(args) match {
+          | Failure(err) ->
+            !error = Some(err);
+            break void
+          | Success _ -> void
           }
         }
       }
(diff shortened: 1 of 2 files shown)


cli/src/parser.sk
```diff
@@ -253,7 +260,7 @@ private fun parse(
         (str, None())
       };
       options.maybeGet(argName) match {
-      | Some(arg) if (arg._takesValue()) ->
+      | Some(arg) if (arg is ValuedArg _) ->
```
mbouaziz: Can this be written `| Some(arg @ ValuedArg _) ->`?

cli/src/parser.sk
```diff
@@ -286,29 +286,43 @@ private fun parse(
...
+              if (!values.containsKey(arg.name)) {
+                values.set(arg.name, BoolValue(true))
+              } else {
+                !error = Some(DuplicateValueError(arg.name))
+              }
+            | _ ->
+              !error = Some(InvalidArgumentError(str));
+              break void
+            }
+          }
```
mbouaziz: Alternatively: eat characters as long as they are `BoolArg`s, the possibly remaining one is a `ValuedArg` in short form. E.g. `-abcO2`

cli/src/parser.sk
beauby: Indeed. Updated.

cli/src/parser.sk
beauby: I thought of this but I am not sure how common a use case this is. Do you some tools that do it?

jberdine: One example is `make -kj16`

beauby: Alright, I'll update!

cli/src/parser.sk
```diff
@@ -371,10 +361,23 @@ mutable class Parser(
...
+    args: Array<(Arg, Value)>,
+  ): Result<void, ArgumentError> {
+    for ((arg, value) in args) {
+      updateValue(this.values, arg, value) match {
+      | Failure(err) -> return Failure(err)
+      | Success _ -> void
+      }
+    };
+
+    Success(void)
```
jberdine, line 374: Aside: This makes me think we should have something like `Iterator.iter_result` analogous to:
```
val fold_result : 
  'a t ->
  init:'accum ->
  f:('accum -> 'a -> ('accum, 'e) Result.t ->
  ('accum, 'e) Result.t
fold_result t ~init ~f is a short-circuiting version of fold that runs in the Result monad. If f returns an Error _, that value is returned without any additional invocations of f.
```
I guess it would be almost the same as `Iterator.each` but wrapping the inner function call with a check for `Failure`.

cli/src/parser.sk
```diff
@@ -391,24 +394,49 @@ mutable class Parser(
...
       this.options.maybeGet(argName) match {
-      | Some(arg @ StringArg _)
-      | Some(arg @ IntArg _) ->
-        parse_value(arg, strValue).map(v -> (arg, v))
-      | _ -> Failure(InvalidArgumentError(str))
+      // Check for short forms that take values such as `-O2`.
+      | Some(arg @ ValuedArg _) ->
+        strValue = str.substring(str.getIter().forward(2));
+        parse_value(arg, strValue).map(v -> Array[(arg, v)])
+      // Check for short boolean flags stringed together such as `-abc`,
```
jberdine: strung

jberdine (approved):
LGTM

cli/src/parser.sk
beauby: Oh yeah that would be nice.

PR #171, opened 2024-03-20
[cli] Support repeatable bool args.

beauby:
This commit replaces the dubious concept of `ArrayArg` with a
`repeatable` property on scalar args. This enables repeatable bool
args, both in the `-v -v -v` form and in the `-vvv` form.

Based on #170 and #168.


jberdine (approved):
LGTM

PR #172, opened 2024-03-20
[stdlib] Support `env` in `Posix.Popen` interface.

beauby:


prelude/src/stdlib/other/System.sk
```diff
@@ -287,7 +287,17 @@ fun subprocess(
...
+}
+
+fun popen{
+  args: Array<String>,
+  env: Map<String, String> = Map[],
+  clear_env: Bool = false,
+  stdout: (String) -> void = _ -> void,
+  stderr: (String) -> void = _ -> void,
+}: CompletedProcess {
+  p = Posix.Popen::create{args, env, clear_env, stdout => true, stderr => true};
```
jberdine, line 300: Why not also `stdin` while you are at it?

jberdine (approved):
LGTM

prelude/src/stdlib/other/System.sk
beauby: The `popen` interface should be made more general (and support for instance `cwd`, though this is notoriously cumbersome with `posix_spawn()`). For `stdin`, do you mean a boolean option to forward to `Posix.Popen::create`?

jberdine: Yes, just thinking that if it is already there in `Popen::create` then might as well support it here too.

beauby: I agree on the principle, but I figured then we should also support "just establishing pipes" for `stdout`/`stderr` (rather than the closures being invoked upon reads). My take away is that current `popen()`/`subprocess()` APIs are incomplete and not great, and we should rethink them through.

jberdine: Ok yes, that makes sense.

PR #173, opened 2024-03-21
[cli] Refactor parser.

beauby:
This will make further PRs easier to review.

### DIFF
--- a/cli/src/parser.sk
+++ b/cli/src/parser.sk
@@ -150,32 +150,27 @@ fun parseArgsFrom(cmd: Command, argv: Sequence<String>): ParseResults {
 private fun updateValue(
   values: mutable Map<String, Value>,
   arg: Arg,
-  value: String,
+  value: Value,
 ): ?ArgumentError {
   arg match {
   | ArrayArg _ ->
+    strValue = value match {
+    | StringValue(v) -> v
+    | _ -> invariant_violation(`Unexpected value type for ArrayArg`)
+    };
+
     values.maybeGet(arg.name) match {
     | Some(ArrayValue(arr)) ->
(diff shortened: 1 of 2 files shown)


cli/src/parser.sk
```diff
@@ -150,61 +150,63 @@ fun parseArgsFrom(cmd: Command, argv: Sequence<String>): ParseResults {
...
-  values: mutable Map<String, Value> = mutable Map<String, Value>[],
-  globals: Array<Arg> = Array[],
-): ParseResults {
-  subcommands = mutable Map[];
-  for (subcmd in cmd._subcommands) {
-    if (subcommands.containsKey(subcmd.name)) {
+private fun defaultValue(arg: Arg): Value {
+  arg match {
+  | ValuedArg _ -> arg._default.default(MissingValue())
+  | BoolArg _ -> arg._default.default(BoolValue(false))
```
jberdine, line 182: Just a passing comment, but I don't understand why BoolArg is special in this way. Doesn't it make sense for bool args that default to true to be possible?

jberdine (approved):
LGTM. The CI essentially does not test PRs like this, I guess you did some local testing?

cli/src/parser.sk
beauby: They can indeed default to `true`, if `arg._default` is `Some(true)`. As for special casing the default value in the absence of an explicit default value, it is so that `args.getBool("foo")` returns `false` if `--foo` was not passed (and `foo` was declared as `BoolArg("foo")`.

beauby:
@jberdine Yes, tests are green locally on each commit. We should probably trigger CI when a dependency of `skdb` is modified.

PR #174, opened 2024-03-21
[sql] Add support for omitting columns in INSERT SELECT

### LINKED ISSUE #165: Cannot insert using select if using partial column selection
Given
```
CREATE TABLE pk_single_row_hist (
  id INTEGER PRIMARY KEY,
  t TEXT DEFAULT CURRENT_TIMESTAMP,
  client INTEGER,
  value INTEGER
);
```

When
```
INSERT INTO pk_single_row_hist (client, value) SELECT client, value FROM pk_single_row;
```

Then
```
Error: line 2, character 7:
Incompatible select size
```

jberdine:
Values for columns that are either PRIMARY KEYs, NULLABLE, or have a
DEFAULT may be omitted when inserting rows. This PR adds support for
such omissions when INSERTing the rows returned by a SELECT.

The `computeInsert` function is refactored to expose
`computeInsertRow`, which contains the functionality to insert omitted
defaults, etc. This `computeInsertRow` function is then also called
when evaluating INSERT SELECT, in the mapper over the rows returned by
the query that inserts them into the table. Also, the error checking
code is adjusted accordingly.

closes #165


sql/src/SqlEval.sk
```diff
@@ -149,75 +149,84 @@ fun computeInsert(
...
-      for (userDefinedCol in columnParams) {
-        if (!missingCols.containsKey(userDefinedCol)) {
-          error(pos, "Column not found: " + userDefinedCol);
-        };
-        missingCols.remove(userDefinedCol);
+    computeInsertRow(context, pos, columnParamsOpt, values, table, repeat)
+  })
+}
+
+fun computeInsertRow(
```
jberdine, line 156: Github does not show the diff well, but note that the body of this function is simply moved unchanged. No need to carefully compare.

sql/src/SqlEval.sk
```diff
@@ -591,7 +631,23 @@ class Evaluator{options: Options, user: ?UserFile} {
           sinkDirName,
           (context, _writer, key, valueIter) ~> {
             src = SKStore.Path(selectDir.dirName, key);
-            entries = Vector[(key, (src, src, valueIter.collect(Array)))];
+            rows: Array<SKStore.File> = valueIter
+              .map(row -> {
+                row match {
+                | RowValues(cvalues, repeat) ->
```
jberdine, line 637: I have not found a way to write this without an unsafe down-cast like this. Is there a way?

sql/src/SqlEval.sk
```diff
@@ -591,7 +631,23 @@ class Evaluator{options: Options, user: ?UserFile} {
           sinkDirName,
           (context, _writer, key, valueIter) ~> {
             src = SKStore.Path(selectDir.dirName, key);
-            entries = Vector[(key, (src, src, valueIter.collect(Array)))];
+            rows: Array<SKStore.File> = valueIter
```
jberdine, line 634: Typing fails without this annotation, but I haven't been able to find a better way to write this.

sql/src/SqlEval.sk
gregsexton: Yeah, we have code like that everywhere. SKStore has some type safety but it's a layer on top which isn't always used.

sql/src/SqlEval.sk
gregsexton: Thanks. Very helpful. :)

gregsexton:
Looks great! Thanks for fixing!

I did just create a new issue (#175) if you're on a roll. ;) :D

PR #177, opened 2024-03-25
[skip] Question mark operator.

beauby:
This commit introduces the postfix question mark operator, which semantically desugars as follows:

```
// For a value `x` of type `Result<T, E>`:
x?

// desugars to:
x match {
| Success(v) -> v
| Failure(err) -> return Failure(err)
}
```

```
// For a value `x` of type `Option<T>`:
x?

// desugars to:
x match {
| Some(v) -> v
| None _ -> return None()
}
```

In practice, it is actually desugared to:
```
x.branch() match {
| ControlFlowContinue(v) -> v
| ControlFlowBreak(r) -> return <ReturnType>::from_residual(r)
}
// where `<ReturnType>` is the return type of the current function.
```
which works on any type implementing the following `Try` trait:

```
trait Try<+TOutput, +TResidual> {
  fun branch(): ControlFlow<TOutput, TResidual>;
  static fun from_residual(r: TResidual): this;
}
```

NOTE: This is more of a proof of concept, as I am not too familiar with the typer, or with the `FastOption` implementation.

### DIFF
--- a/prelude/src/core/Option.sk
+++ b/prelude/src/core/Option.sk
@@ -34,6 +34,10 @@ base class Option<+T> uses
   | Some(value: T)
   | None()
 
+  fun liftFailure<U>(): Result<T, None<U>>
+  | Some(val) -> Success(val)
+  | None() -> Failure(None())
+
   // For Some() values, returns a new Option with the result of applying the
   // mapping function to the contained value. Returns None() when this is None().
   fun map<T2>(f: T -> T2): ?T2
--- a/prelude/src/core/Result.sk
+++ b/prelude/src/core/Result.sk
@@ -28,6 +28,10 @@ base class Result<+T, +E> {
   | Failure(error: E)
   | Success(value: T)
 
+  fun liftFailure<U>(): Result<T, Failure<U, E>>
+  | Success(val) -> Success(val)
+  | Failure(err) -> Failure(Failure(err))
+
   // # Creating a Result
 
   // Execute the given function, returning as follows:
--- a/compiler/src/SkipParseTree.sk
+++ b/compiler/src/SkipParseTree.sk
@@ -4564,6 +4564,43 @@ class TypeParametersTree{
   }
 }
 
+class EnsureExpressionTree{
+  operand: ParseTree.ParseTree,
+  operator: ParseTree.ParseTree,
+} extends ParseTree.ParseTree {
+  fun getNamedFields(): List<(String, ParseTree.ParseTree)> {
+    List<(String, ParseTree.ParseTree)>[
+      ("operand", this.operand),
+      ("operator", this.operator),
+    ];
+  }
+
+  fun getKind(): String {
+    "EnsureExpression";
+  }
+
+  fun getChildren(): mutable Iterator<ParseTree.ParseTree> {
+    yield this.operand;
+    yield this.operator;
+    yield break;
+  }
+
+  fun transform(
+    codemod: mutable CodeMod,
+  ): (ParseTree.ParseTree, Vector<Subst>) {
+    tx_operand = codemod.transform(this.operand);
+    tx_operator = codemod.transform(this.operator);
+    (
+      EnsureExpressionTree{
+        range => this.range,
+        operand => tx_operand.i0,
+        operator => tx_operator.i0,
+      },
+      Vector[tx_operand.i1, tx_operator.i1].flatten(),
+    );
+  }
+}
+
 class UnaryExpressionTree{
   operator: ParseTree.ParseTree,
   operand: ParseTree.ParseTree,
--- a/compiler/src/SkipParser.sk
+++ b/compiler/src/SkipParser.sk
@@ -3874,6 +3874,15 @@ mutable class SkipParser{
       this.parsePostfixSuffix(this.parseMemberSelection(value))
     | TokenKind.COLON_COLON() ->
       this.parsePostfixSuffix(this.parseMemberSelection(value))
+    | TokenKind.QUESTION() ->
+      operator = this.tokenResult();
+      this.parsePostfixSuffix(
+        ParseTree.EnsureExpressionTree{
+          range => Parser.createRangeOfTrees(value, operator),
+          operator,
+          operand => value,
+        },
+      )
     | _ -> value
     }
   }
@@ -4925,6 +4934,7 @@ mutable class SkipParser{
     | ParseTree.TemplateLiteralExpressionTree _
     | ParseTree.ErrorTree _
     | ParseTree.EmptyTree _
+    | ParseTree.EnsureExpressionTree _
     | ParseTree.InfiniteLoopExpressionTree _
     | ParseTree.DoLoopExpressionTree _
     | ParseTree.WhileLoopExpressionTree _
--- a/compiler/src/convertTree.sk
+++ b/compiler/src/convertTree.sk
@@ -586,6 +586,8 @@ class Converter{filename: String} {
         operator,
         member,
       )
+    | ParseTree.EnsureExpressionTree{range, operand} ->
+      this.convertEnsureExpression(this.convertRange(range), operand)
     | ParseTree.UnaryExpressionTree{range, operator, operand} ->
       this.convertUnaryExpression(this.convertRange(range), operator, operand)
     | ParseTree.WithExpressionTree{range, value, arguments} ->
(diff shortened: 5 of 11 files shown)


jberdine:
I might get flamed for this, but note that what you are actually doing here is very close to lifting the Option and Result monads, via their standard monad transformers, into the language's implicit monad where bind is `;` and pure is `return`. The implementations of the new trait look like they are mainly the standard monad transformer, see e.g. https://en.wikipedia.org/wiki/Monad_transformer, where `branch` corresponds to the transformer's bind. I expect that if we added standard monad transformers to Option and Result, they could be used instead of the implementations of the Try trait. That wouldn't be any advantage itself, but maybe there are other uses of those monad transformers.

jberdine:
Thinking a bit more, the magic/builtin nature of the language's `;`-`return` monad will prevent being able to use it with a monad transformer defined in the language. So that kills that approach I think.

I think I would call this something like `EarlyReturn` rather than `Try` to avoid confusion with association to exceptions, as this doesn't have anything to do with them.

Am I missing something, or could this be simplified by noting that the `ControlFlow` type is isomorphic to `Result`, and so couldn't the trait be defined as:
```
trait Try<+T, +E> {
  fun toResult(): Result<+T, +E>;
}
```
and the expansion of `x?` be:
```
x.toResult() match {
| Success(v) -> v
| Failure(e) -> return e
}
```
Then when defining the instances, the `toResult` function would not be specialized to just this usage, but could be generally useful.

beauby:
> Thinking a bit more, the magic/builtin nature of the language's `;`-`return` monad will prevent being able to use it with a monad transformer defined in the language. So that kills that approach I think.

I will fully trust you on that.
> 
> I think I would call this something like `EarlyReturn` rather than `Try` to avoid confusion with association to exceptions, as this doesn't have anything to do with them.

Agreed, the name `Try` is confusing. Will update to `EarlyReturn`.
> 
> Am I missing something, or could this be simplified by noting that the `ControlFlow` type is isomorphic to `Result`, and so couldn't the trait be defined as:
> 
> ```
> trait Try<+T, +E> {
>   fun toResult(): Result<+T, +E>;
> }
> ```
> 
> and the expansion of `x?` be:
> 
> ```
> x.toResult() match {
> | Success(v) -> v
> | Failure(e) -> return e
> }
> ```
> 
> Then when defining the instances, the `toResult` function would not be specialized to just this usage, but could be generally useful.

Do you mean having `Result` implement `toResult(): Result<T, Failure<void, E>>` or `toResult(): Result<T, E>`?
In both case, simply writing
```
x.toResult() match {
| Success(v) -> v
| Failure(e) -> return e
}
```
would not work for things such as:
```
fun bar(): Result<String, Error> { ... }

fun foo(): Result<Int, Error> {
  x = bar()?;
  if (x == "hello") {
    Success(1)
  } else {
    Success(2)
  }
}
```
which is expected to be a common use case. The reason is that `bar()?` would return either `Failure<String, Error>` or `Failure<void, Error>`, neither of which are `Failure<Int, Error>`. This is the main reason why I'm doing type erasure and fiddling with the typer in the first place. Or am I missing something?

Edit: 
Sorry, I got confused. The `ControlFlow` type is indeed isomorphic to `Result`, so we could get rid of `ControlFlow`. The problems around return types remain the same though.

jberdine:
I would be a lot more comfortable if we could implement this as just a syntactic expansion, without needing any tricks in the type checker. I played with this a bit more and so far I'm encouraged. With the following in Result.sk:
```
  fun toResult<U>(): Result<T, Failure<U, E>>
  | Success(val) -> Success(val)
  | Failure(err) -> Failure(Failure(err))
```
and the following in Option.sk:
```
  fun toResult<U>(): Result<T, None<U>>
  | Some(val) -> Success(val)
  | None() -> Failure(None())
```
then the following code type-checks:
```
type Error = Float;

fun bar1(): Result<String, Error> {
  if (Random.next() / 2 == 0) {
    Success("baz")
  } else {
    Failure(0.0)
  }
}

fun foo1(): Result<Int, Error> {
  x = bar1().toResult() match {
  | Success(v) -> v
  | Failure(e) -> return e
  };
  if (x == "hello") {
    Success(1)
  } else {
    Success(2)
  }
}

fun bar2(): ?String {
  if (Random.next() / 2 == 0) {
    Some("baz")
  } else {
    None()
  }
}

fun foo2(): ?Int {
  x = bar2().toResult() match {
  | Success(v) -> v
  | Failure(e) -> return e
  };
  if (x == "hello") {
    Some(1)
  } else {
    Some(2)
  }
}
```
This makes me think that we could make
```
x?
```
expand to
```
  x.toResult() match {
  | Success(v) -> v
  | Failure(e) -> return e
  }
```
What I don't think is possible is to define a trait for the `toResult` function. Consider the two instances for Option and Result:
```
  fun toResult<U>(): Result<T, None<U>>
  fun toResult<U>(): Result<T, Failure<U, E>>
```
In order to unify those, we need to be able to abstract over a type constructor, which would be `U -> None<U>` in the Option case and `U -> Failure<U, E>>` in the Result case. We would need a way to specify those when declaring that Option and Result implement the trait. For that we would need higher-kinded polymorphism, functors, etc.

Personally, I think I would prefer to have a simple syntactic expansion that just generates calls to the `toResult` function, and leave the trait implicit. It would mean that we could not use the trait as a constraint on type parameters, and so could not write code that is e.g. parametric over Option and Result, but I don't think that we can really do that without functors anyhow.

beauby:
This indeed works. One drawback with having the construction be a simple syntactic expansion is error messages (the typer will complain about `.toResult()` not being defined or not returning the correct type if `?` is used on types that are neither `Option` nor `Failure`), but I don't think it matters much at this point and we can always change the implementation later.

prelude/src/core/Option.sk
```diff
@@ -34,6 +34,10 @@ base class Option<+T> uses
   | Some(value: T)
   | None()
 
+  fun toResult<U>(): Result<T, None<U>>
+  | None _ -> Failure(None())
```
jberdine: Just curious: is there a reason to prefer `None _` over `None()`?

tiny nit: the other instances have the Success case first

prelude/src/core/Result.sk
```diff
@@ -28,6 +28,10 @@ base class Result<+T, +E> {
   | Failure(error: E)
   | Success(value: T)
 
+  fun toResult<U>(): Result<T, Failure<U, E>>
+  | Success(val) -> Success(val)
+  | Failure(err) -> Failure(Failure(err))
```
jberdine: I hate to go back and forth, but when thinking of the `toResult` name I was expecting that this function would be the identity rather than lifting the `Failure` case. That plus the comment about not having good error messages when implementing `?` as a syntactic expansion makes me think that the `toResult` name isn't great. Considering these functions by themselves, they are essentially lifting the failure case into the "failure monad", so `liftFailure` would make sense. But maybe there is a different name that would make an error message coming from using `e?` when `e` doesn't implement the needed function more understandable. With no keyword in `e?` though, it is hard to guess what name would be a good guide. I guess we could add a special case in the compiler where undefined function error messages originate that checks if the needed function is this one and if so adjusts the message to include something like "e? can only be used if the type of e implements liftFailure<U>(): Result<T, _>, such as Option or Result".

jberdine (approved):
LGTM. Just some dithering about naming inline.

prelude/src/core/Option.sk
beauby: > Just curious: is there a reason to prefer `None _` over `None()`?

No, not really. I just have taken the habit of using `_` instead of `()` when matching types with no parameters.

> tiny nit: the other instances have the Success case first

Will update.

prelude/src/core/Result.sk
beauby: I agree that the `toResult()` name is weird. I'll update the name to `liftFailure()`.

Since we're now expanding the `?` operator when building the AST from the CST, I don't think there is a solid way to customize error messages: if for some reason the user defines a `liftFailure()` method on a type, but calls it on the wrong type, they would get an error about the `?` operator although it has nothing to do with it.

jberdine: Yes, making the error reporting code look for `liftFailure` could lead to mis-identifying something as a mistake related to `?`. But it might be worth it anyhow if the error message was worded carefully to indicate that `?` is only one possible explanation. All of the "did you mean" stuff, in skip and other languages, also falls into that category.

PR #178, opened 2024-03-25
Support for client-specified schemas in mirror/tail operations

bennostein:
This PR adds support for client-specified schemas; instead of enforcing that schemas exactly match between replicas, we allow tailers to specify an expected schema.  That schema is allowed to (a) reorder columns and (b) drop any columns that are nullable or include a default value.  All data-layer communication happens in the given schema, with translations on the server for both the write and read path. 

Still a WIP -- the bulk of the functionality is in place, but needs more thorough testing.

### DIFF
--- a/rfc/003-protocol.org
+++ b/rfc/003-protocol.org
@@ -321,6 +321,8 @@ Where not specified, strings are UTF-8.
 - push promise
   - type = 0x3
   - reserved (24) - padding - must be zeroed
+  - schemas_length (16) - in bytes
+  - schemas (schema_length * 8) - JSON-encoded map from table names to schemas
 
 - schema
   - type = 0x4
--- a/rfc/003-protocol.org
+++ b/rfc/003-protocol.org
@@ -321,7 +321,7 @@ Where not specified, strings are UTF-8.
 - push promise
   - type = 0x3
   - reserved (24) - padding - must be zeroed
-  - schemas_length (16) - in bytes
+  - schemas_length (32) - in bytes
   - schemas (schema_length * 8) - JSON-encoded map from table names to schemas
 
 - schema
--- a/sql/server/core/src/main/kotlin/core/Orchestration.kt
+++ b/sql/server/core/src/main/kotlin/core/Orchestration.kt
@@ -289,7 +289,7 @@ fun encodeProtoMsg(msg: ProtoMessage): ByteBuffer {
       buf.flip()
     }
     is ProtoPushPromise -> {
-      val buf = ByteBuffer.allocate(6 + msg.schemas.length * 4)
+      val buf = ByteBuffer.allocate(8 + msg.schemas.length * 4)
 
       buf.putInt(0x0)
       buf.put(0, 0x03.toByte())
--- a/sql/server/core/src/main/kotlin/core/Orchestration.kt
+++ b/sql/server/core/src/main/kotlin/core/Orchestration.kt
@@ -46,7 +46,7 @@ data class ProtoSchemaQuery(
     val suffix: String? = null,
 ) : ProtoMessage()
 
-class ProtoPushPromise(val schemas: String) : ProtoMessage()
+class ProtoPushPromise(val schemas: ByteArray) : ProtoMessage()
 
 data class ProtoCreateDb(val name: String) : ProtoMessage()
 
@@ -118,7 +118,7 @@ fun decodeProtoMsg(data: ByteBuffer): ProtoMessage {
       val schemasLength = data.getShort()
       val schemasBytes = ByteArray(schemasLength.toInt())
       data.get(schemasBytes)
-      ProtoPushPromise(String(schemasBytes, StandardCharsets.UTF_8))
+      ProtoPushPromise(schemasBytes)
     }
     4u -> {
       val scope = SchemaScope.values()[(data.get().toUInt() and 0x0Fu).toInt()]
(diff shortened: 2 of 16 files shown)


gregsexton:
I'll review this once you've rebased. LMK if you run in to issues.

sql/server/dev/src/main/kotlin/Service.kt
```diff
@@ -188,7 +188,7 @@ class RequestHandler(
       }
       is ProtoRequestTail -> {
         if (!skdb.canMirror(request.table, request.expectedSchema)) {
-          stream.error(2003u, "Schema mismatch")
+          stream.error(2003u, "Invalid schema for ${request.table}")
```
gregsexton, line 191: [major] Just a reminder, any changes you make to Service.kt probably also need to be made to App.kt in the cloud repo.

sql/src/SqlTailer.sk
```diff
@@ -130,20 +168,31 @@ fun tailSub(
       print_error("Error: session not found");
       skipExit(2)
     | Some(sub) ->
+      dirDescrs = getDirDescrs(gContext);
       sub with {
         dirSubs => sub.dirSubs.map(dirSub -> {
-          tableName = P.Name::create(
-            dirSub.dirName.toString().stripPrefix("/").stripSuffix("/"),
-          );
+          table = dirSub.dirName.toString().stripPrefix("/").stripSuffix("/");
```
gregsexton: I just created a helper for this in SqlTables.sk: getSqlTableName

rfc/003-protocol.org
```diff
@@ -321,6 +321,8 @@ Where not specified, strings are UTF-8.
 - push promise
   - type = 0x3
   - reserved (24) - padding - must be zeroed
+  - schemas_length (16) - in bytes
```
gregsexton: 16 bits might not be enough? Back of the envelope, I got ~3-4k columns. For a large app with many tables, I think you could run in to this limit. Let's just bump to 32? It's not a frequently sent message.

sql/server/core/src/main/kotlin/core/Orchestration.kt
```diff
@@ -285,9 +289,24 @@ fun encodeProtoMsg(msg: ProtoMessage): ByteBuffer {
...
     }
     is ProtoPushPromise -> {
-      val buf = ByteBuffer.allocate(4)
+      val buf = ByteBuffer.allocate(6 + msg.schemas.length * 4)
+
       buf.putInt(0x0)
       buf.put(0, 0x03.toByte())
+
+      val pos = buf.position()
+      buf.putInt(0x0)
```
gregsexton: putShort?

sql/server/core/src/main/kotlin/core/Skdb.kt
```diff
@@ -132,18 +137,33 @@ class Skdb(val name: String, private val dbPath: String) {
...
+            replicationId,
+            "--expect-schemas")
 
     // TODO: for hacky debug
     pb.redirectError(ProcessBuilder.Redirect.INHERIT)
 
     val proc = pb.start()
 
+    val stdin = proc.outputStream.buffered()
+    stdin.write((schemas + "\n").toByteArray(StandardCharsets.UTF_8))
```
gregsexton: [minor] More efficient to call write twice (as it's buffered) than to concat the string then call it once.

gregsexton: But it would be even better if we could just pass the utf8 bytes straight through - no need to decode and re-encode them. Do we do that elsewhere for e.g. tail spec?

sql/src/Main.sk
```diff
@@ -54,6 +54,47 @@ fun queryParams(options: SKDB.Options): Map<String, P.Value> {
   }
 }
 
+// Read a single line from stdin, decode it to a JSON Object, expected
+// to be a map from table names to corresponding schemas (both strings)
+fun readSchemas(): Map<String, Array<P.ColumnDefinition>> {
```
gregsexton: For consistency, might be worth moving this to the top of SqlEval, where we have other similar functions.

sql/src/SqlTables.sk
```diff
@@ -616,10 +616,19 @@ fun getIndexByName(
...
 /*****************************************************************************/
 
+fun getDirDescrs(
+  origContext: readonly SKStore.Context,
+): SortedMap<P.Name, SKDB.DirDescr> {
+  context = SKStore.Context::fromSaved(origContext.clone());
+  getTableDir(context)
+    .items(context)
+    .flatMap(x -> x.i1)
+    .reduce((acc, curr) -> acc.set(curr.name, curr), SortedMap::create())
```
gregsexton, line 627: [minor] Can you `.collect(SortedMap)`?

sql/src/SqlCsv.sk
```diff
@@ -355,6 +417,12 @@ fun applyDiffStrategy(
       tableName = table.name.origName;
       authorCol = SKDB.getAuthorIdx(table);
       accessCol = SKDB.getAccessIdx(table);
+      rows = sourceSchema match {
```
gregsexton, line 431: Should we just pass this in to applyDiffStrategy? Stops the interface growing and helps keep the complexity of this down.

gregsexton (changes requested):
Looks great! I left you a bunch of comments, but I think they're all really minor.

Couple of requests:
- Could you add some testing of this to the integration tests (apitests.ts) and also the soak? I think it looks solid, but this would add a lot of confidence.
- Can we bump the min version supported in the server? Between this and my changes in #182 we've made a lot of backwards breaking changes. I missed doing this and just thought of it. Then we'll need to cut a release with this and 182 in and go around updating everything. I'm happy to take care of that.

bennostein:
Awesome, thanks for the comments!  Will go through and address those before landing anything, they all sound good to me.

Re: tests, I'm in the process of adding integration tests, and will add to the soak too.

sql/src/SqlCsv.sk
bennostein: I think not -- we need a (mutable) context to do this translation (to compute default values), so pushing it down into applyDiffStrategy makes sure that context transformations are produced/applied as expected.

Alternatively, we could use the mutable context from `replayDiff` to translate these rows, but I was worried that would interfere with the context juggling happening here.

sql/src/SqlTables.sk
bennostein: No FromIterator implementation for SortedMap, no.  Could add that separately but leaving as is for now.

gregsexton (approved):
LGTM! Thanks in particular for making the change to pass the byte array through. String decoding and re-encoding can be surprisingly expensive.

PR #181, opened 2024-03-26
[runtime:node] Use path library.

skiplabsdaniel:


prelude/ts/src/sk_node.ts
```diff
@@ -20,7 +21,9 @@ class WrkImpl implements Wrk {
     url: URL,
     options: WorkerOptions | undefined,
   ): Wrk {
-    const filename = "." + url.pathname.substring(process.cwd().length);
+    const cwd = process.cwd();
+    const child = url.pathname.startsWith(cwd + path.sep);
+    const filename = child ? "." + url.pathname.substring(cwd.length) : url;
```
mbouaziz: Another version:
```js
const filename = path.relative(process.cwd(), url.pathname);
```
I don't know which is better :shrug:

prelude/ts/src/sk_node.ts
```diff
@@ -61,18 +64,21 @@ class Env implements Environment {
...
-      path = "." + url.pathname.substring(process.cwd().length);
+      const child = url.pathname.startsWith(cwd + path.sep);
+      filename = child ? "." + url.pathname.substring(cwd.length) : url.pathname;
     // @ts-ignore
     } else if (url && url.default) {
-      // @ts-ignore
-      path = "." + url.default.substring(process.cwd().length);
+      const aurl = url as any;
+      const child = aurl.default.startsWith(cwd + path.sep);
+      filename = child ? "." + aurl.default.substring(cwd.length): aurl.default;
```
mbouaziz:
```suggestion
      const default = url.default as string;
      const child = default.startsWith(cwd + path.sep);
      filename = child ? "." + default.substring(cwd.length): default;
```
?

prelude/ts/src/sk_node.ts
skiplabsdaniel: I hesitated to do it, I asked myself the same question. It's true that it's more concise.

skiplabsdaniel: But the if it's relative path it must start with "./"

mbouaziz (approved):

PR #182, opened 2024-03-27
Remove resets from the replication protocol

gregsexton:
tl;dr: collapse the reset mechanism in the protocol to a cold start.
This greatly reduces complexity in the protocol and removes potential
for large scale data loss.

Due to directory purging, we sometimes cannot produce a diff for a
given client's clock. In these situations the protocol sends a
'reset'. This is a full dump of the table with a marker to tell a
client to apply it as an overwrite. (There's logic to handle
concurrent writes.)

The purging/GC behaviour is essential but this way of handling the
situation for replication is undesirable:

1. If the logic has bugs, we're dealing with entire table re-writes,
   and so it's likely to go catastrophically wrong and lead to large
   scale data loss. Particularly on the server, which is handling the
   data from multiple clients, and applying mass deletes concurrently
   for partial views due to privacy. This is very difficult to get
   right and makes changing the protocol dangerous and difficult.

2. Clients will keep up with the server most of the time and so resets
   are a rare case. This makes resets an infrequently exercised path
   that is not tested in regular day to day operation. With the above
   point, they're very dangerous.

3. They double the complexity of the entire replication algorithm as
   they're a different case to handle orthogonal to everything else.

Instead, this diff eliminates push-based resets and replaces them with
a pull based mechanism that (to distinguish it) we call 'rebuilds'.

If we cannot create a diff, due to GC/purge, then we send a message
that simply says 'rebuild'. The recipient then must clear out their
data and reconnect from 0. This collapses this case to a regular cold
start which is a heavily exercised and tested path. It also benefits
the client as they'll receive optimisations that were never
implemented for resets, such as chunking.

Receiving a rebuild works great for clients. The server, however, has
rebuild processing disabled as a) it cannot reconnect to a client
and rebuild and b) we DO NOT WANT to handle mass data rebuilding on
the server, this is where it is most dangerous.

To make this work, we ensure that clients never send a rebuild
message. This is a scalable proposition for clients. We track where
the client has been ack'd up to and ensure we never purge past this
point. When calculating a diff we always have the data needed to
produce it without rebuilds. The server can't do this as it needs to
scale to many clients that can lag arbitrarily.


gregsexton:
This was quite the slog. I was surprised by how much I needed to track down, debug, and rework. So sorry it's a large PR: it needs to be merged atomically.

The main change is described above. To make it work and pass tests/soak I had to:

- make the soak test much more robust in its expectation checking
- rework chunking in the tailer
- change filtering for tail since 0 to not exclude user's changes
- create groups atomically under a single txn. This required special casing privacy rules specifically for groups to break the cyclic dependency. It also meant moving privacy checks in to the lock in write-csv and checking them incrementally.
- relax when we issue rebuilds due to privacy changes. Now we don't rebuild if the change was made by this user. This is more efficient and also leads to a better UX for group creators - see the commit for details.
- several more minor changes (I wrote detailed descriptions in all of the commits)

gregsexton:
Lastly, here's a good proxy for why I'm doing this: https://github.com/SkipLabs/skdb/pull/182/commits/13c1352d35358e388fefcad253259f532c849cbb I was able to remove ~1k lines of tests as we remove the consideration entirely.

gregsexton:
I'm confused. The mem test works for me locally every time. If it runs out of memory it doesn't timeout, it throws an exception and fails, so I think this is actually due to a perf regression. But I just measured and I actually see a slight perf improvement with my latest changes.

I'm going to try re-running to see if this is reproducible or whether I just got really unlucky.

jberdine:
Have you tried locally when limiting docker to 4 CPUs and 8GB RAM? Sometimes funny perf things happen as you approach memout...

gregsexton:
Good idea!

gregsexton:
Benchmarking locally, there is no regression in memory or cpu util.

SSHing in to CircleCI, I see that the box is occasionally highly saturated (run length >> num cpu cores). Running the test suite in a loop, the threads and mem regression tests will occasionally timeout: they're longer running and are presumably being starved CPU time.

I bumped the timeout for the flaky tests. Let's see if it passes now...

sql/ts/src/skdb_group.ts
```diff
@@ -50,54 +50,31 @@ export class SKDBGroupImpl implements SKDBGroup {
...
+      .add(
+        "INSERT INTO skdb_groups VALUES (@adminGroupID, @userID, @ownerGroupID, @adminGroupID);",
+      )
+      .add(
+        "INSERT INTO skdb_groups VALUES (@groupID, @userID, @adminGroupID, 'read-write')",
+      )
+      .add(
+        "INSERT INTO skdb_groups VALUES (@ownerGroupID, @userID, @ownerGroupID, @ownerGroupID);",
+      )
+      .commit({ groupID, adminGroupID, ownerGroupID, userID });
```
bennostein, line 77: This is great -- it was always a little nerve-wracking to have this happen outside of a transaction to satisfy privacy constraints as we put together the admin/owner/primary group.

sql/src/SqlPrivacy.sk
```diff
@@ -549,13 +553,13 @@ fun checkUserCanWriteRow(
...
       groupIndex = skdbGroups.getIndex(context);
-      adminID = groupIndex.get(skdbGroups.makeKey(groupID)) match {
-      | Some(r) -> r.getString(adminIDColName)
-      | None() -> throw SqlError(0, "malformed predefined table skdb_groups")
+      groupIndex.get(skdbGroups.makeKey(groupID)) match {
+      | Some(r) ->
+        adminID = r.getString(adminIDColName);
+        (row.repeat == 0 && access.canDelete(context, adminID)) ||
+          access.canInsert(context, adminID)
+      | None() -> return AROK()
```
bennostein, line 572: I see know why this makes sense -- disregard my other comment on #185 !

rfc/003-protocol.org
```diff
@@ -383,32 +383,77 @@ There are plans to re-design this protocol and use a binary
...
+  starting.
+- this payload takes you up to date with the sender at tick 197. A
+  should update its watermarks appropriately.
+
+
+Response from A:
+#+BEGIN_SRC
+^table1__skdb_mirror_feedback
+:197
+^table2
```
bennostein: Should this be `table2__skdb_mirror_feedback`?

sql/src/Main.sk
```diff
@@ -119,9 +119,10 @@ untracked fun main(): void {
         .arg(Cli.StringArg("table-suffix").about("Optional suffix")),
     )
     .subcommand(
-      Cli.Command("set-user")
-        .about("Sets the user of the database")
-        .arg(Cli.StringArg("userID").positional().required().about("userID")),
+      Cli.Command("connected-as")
+        .about("Informs the database that it is a replicating peer")
+        .arg(Cli.StringArg("userId").required().about("user ID"))
```
bennostein: Could we give a few words more detail in this (and the following) "about" string?  As is these are a bit redundant with the arg name itself.

sql/src/Main.sk
```diff
@@ -203,7 +204,12 @@ untracked fun main(): void {
             "Globally unique identity of this write stream",
           ),
         )
-        .arg(Cli.StringArg("user").about("Name of the user")),
+        .arg(Cli.StringArg("user").about("Name of the user"))
+        .arg(
+          Cli.BoolArg("enable-rebuilds").about(
+            "Allow write-csv to apply rebuilds. Used wrongly, can lead to data loss.",
```
bennostein: This is a bit ominous/spooky but doesn't say how to avoid using it wrongly.  Would it be right to say something to the effect of "don't enable this server-side"?

bennostein (approved):
Looks great.  A few minor comments, going to start rebasing on top of this now to test.

rfc/003-protocol.org
gregsexton: Yes, great catch. Thanks.

PR #183, opened 2024-03-27
[ci] Add check that code is formatted

jberdine:
This PR makes 3 changes related to code formatting:

1. Optimizes `make fmt`

Now there are separate `fmt-sk`, `fmt-c`, and `fmt-js` targets that
run `skfmt`, `clang-format`, and `prettier`. The `skfmt` and
`clang-format` invocations are parallelized, but `prettier` is not
since its startup time is so high that running in parallel is even
slower. Also note that `make -j fmt` runs all 3 targets in parallel.
This formats all the code in the repo.

2. Addes a git pre-commit hook to check formatting

Install locally by running `make setup-git-hooks`. At that point,
running `git commit` will invoke the pre-commit script to check that
the staged changes are well-formatted. If so, it silently succeeds and
`git commit` proceeds as before. If not, it reports the files that
changed and the diff between the staged code and the formatted
code. At this point `make -sj fmt` will format the files in the
working directory, and the changes can be staged and the commit
process continued.

Note that to override the check and commit even if it fails, use `git
commit --no-verify`.

3. Adds a CI job that checks whether the code is well-formatted.


.circleci/generate_config.sh
```diff
@@ -8,15 +8,21 @@ skfs=$?
...
 skdb=$?
 
-if (($skc == 0 && $skfs == 0 && $skdb == 0))
-then
-    circleci-agent step halt
-fi
+# if (($skc == 0 && $skfs == 0 && $skdb == 0))
+# then
+#     circleci-agent step halt
+# fi
```
beauby: The reason for this was that CircleCI would choke on empty `workflows:` sections. Now that `check-format` is always present, it can safely be removed.

skdate/extern/src/date.cpp
```diff
@@ -1,118 +1,88 @@
...
-    }
-    return sk_string_create(locale, strlen(locale));
+/**
+ * Retrieves the local version of variables required for formatting using format
+ * specifier. A => Full weekday name. a => Abbreviated weekday name. B => Full
+ * month name. b => Abbreviated month name. c => Date/Time format string of the
+ * locale. x => Date format string of the locale. X => Time format string of the
+ * locale. r => AM/PM Time format string of the locale. p => AM or PM locale
+ * string.
+ */
```
mbouaziz: :crying_cat_face: 

Use `-` to itemize?

mbouaziz (approved):

gregsexton:
We should make formatting a git hook or something. It's going to be super annoying to come back to a PR and realise the CI failed because you forgot to run make fmt.

jberdine:
> We should make formatting a git hook or something. It's going to be super annoying to come back to a PR and realise the CI failed because you forgot to run make fmt.

(Just in case it isn't clear, note that if a PR fails the format check, it does not prevent the other ci checks from running.)

I looked a bit and don't see a way to do this sort of thing server-side, unless going the self-hosted git server route (perhaps via github enterprise).

What do people think about needing everyone to do some per-checkout config to do the autoformat as a local pre-commit hook? I can make a faster version that only formats files changed in a commit.

gregsexton:
> (Just in case it isn't clear, note that if a PR fails the format check, it does not prevent the other ci checks from running.)

Oh, ok, cool, I just assumed it would prevent the others running, sorry. Still be very good to have a local hook to save some churn.

This[1] seems to suggest that hooks can't be enforced. But that's ok I think. We can have a hooks dir and you can opt in if you find them useful - I guess the goal is to make it so you would want to.

[1]: https://stackoverflow.com/questions/427207/can-git-hook-scripts-be-managed-along-with-the-repository

.circleci/base.yml
```diff
@@ -10,6 +10,19 @@ commands:
...
+    docker:
+      - image: skiplabs/skdb-base:latest
+    steps:
+      - checkout
+      - run:
+          name: Check code is formatted
+          command: |
+            apt-get install -q -y clang-format-15
+            update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-15 100
+            npm install -g prettier
```
jberdine: Note these lines are `[DO NOT COMMIT] temporary per-job setup of auto-formatters` and won't be landed.

skdate/extern/src/date.cpp
```diff
@@ -1,118 +1,93 @@
...
+ * specifier.
+ * - A => Full weekday name.
+ * - a => Abbreviated weekday name.
+ * - B => Full month name.
+ * - b => Abbreviated month name.
+ * - c => Date/Time format string of the locale.
+ * - x => Date format string of the locale.
+ * - X => Time format string of the locale.
+ * - r => AM/PM Time format string of the locale.
+ * - p => AM or PM locale string.
```
jberdine, line 29: Yes, much better with `-` now.

jberdine:
I think that everything has been addressed now. It wouldn't be a bad idea for someone to glance at the shell hacking in Makefile and check_format.sh.

mbouaziz:
:heart: 

It'd be great to mention the hook in some sort of skdb developers doc, e.g. a `CONTRIBUTING` or `HACKING` file?

jberdine:
> It'd be great to mention the hook in some sort of skdb developers doc, e.g. a `CONTRIBUTING` or `HACKING` file?

There is bin/README.md which currently has the release process docs, and the hook scripts do live in bin, so how about there?

mbouaziz:
That's better than nothing but I think we should move the file somewhere else or at least mention the file from the root `README`

jberdine:
I will open another PR with a proposal for dev docs.

PR #184, opened 2024-03-28
Give the ability the copy to parent obstack without deleting + Some cleaning

skiplabsdaniel:


compiler/runtime/obstack.c
```diff
@@ -37,20 +37,31 @@
...
+
+unsigned char* decr_heap_end(size_t size);
+void reset_heap_end();
+
+#else
+
 __thread struct sk_obstack* page = NULL;
 __thread char* head = NULL;
 __thread char* end = NULL;
+__thread sk_cell_t pages[NUMBER_OF_PAGES];
```
gregsexton: We should probably initialise this - dangerous otherwise. Just zero it out with`sk_cell_t pages[N] = { 0 };`

compiler/runtime/obstack.c
```diff
@@ -37,20 +37,31 @@
 // head: The current position in the page.
 // end: The end of the page.
 
+#define NUMBER_OF_PAGES 20
```
gregsexton: Can I just check: what's motivating having a fixed buffer of pages? Seems like the logic gets more complex to reduce just a few allocations? Are you sure it's worth it? Plus we end up allocating a lot of heap for things that don't need it: we tend to have quite a few short lived processes. I know the mem is handed out lazily, but still.

sql/ts/tests/node.play.ts
```diff
@@ -13,19 +16,21 @@ function run(t, asWorker: boolean) {
 tests(false).forEach((t) => run(t, false));
 tests(true).forEach((t) => run(t, true));
 
+
+const N = 765000; // we should be able to process this many rows without running out of address space
```
gregsexton, line 17: Nice! That's quite a bit of fragmentation saved. :)

compiler/runtime/memory.c
```diff
@@ -28,24 +34,40 @@ size_t sk_pow2_size(size_t size) {
...
-void sk_add_ftable(void* ptr, size_t size) {
-  size_t slot = sk_bit_size(size);
-  *(void**)ptr = sk_ftable[slot];
-  sk_ftable[slot] = ptr;
+size_t sk_obj_size(size_t size, sk_size_info_t* size_info) {
+  size_t pow2_size = sk_pow2_size(size);
+  size_t bit_size = sk_bit_size(size);
+  if (pow2_size > size && bit_size > 8) {
+    size_t main_size = 1 << (bit_size - 1);
+    size_t reminder_size = sk_pow2_size(size - main_size);
```
gregsexton: [nit] s/reminder/remainder

sql/ts/tests/node.play.ts
```diff
@@ -1,9 +1,12 @@
-import { test, type Page, expect } from "@playwright/test";
+import { test, expect } from "@playwright/test";
 import { tests } from "./tests";
-import { createSkdb } from "skdb";
+import { createSkdb, type SKDB } from "skdb";
 
 function run(t, asWorker: boolean) {
   test(t.name, async () => {
+    if (t.timeout) {
```
gregsexton: I took care of this and another test in #182. I spent time ssh'd in to circleci understanding resource usage. I'll land it in a minute.

gregsexton (approved):
This is great.

compiler/runtime/memory.c
```diff
@@ -28,24 +34,40 @@ size_t sk_pow2_size(size_t size) {
...
+  }
+  size_info->main = bit_size;
+  size_info->remainder = 0;
+  return  pow2_size;
 }
 
-void* sk_get_ftable(size_t size) {
-  size_t slot = sk_bit_size(size);
-  void** ptr = sk_ftable[slot];
+void sk_add_ftable(void* ptr, sk_size_info_t* si) {
```
gregsexton: This is a bit of a micro optimisation, but I think you're better off passing si by value here (and in other places such as sk_get_ftable where it's only read). The compiler will just utilise registers for the two size_t values, which si will be loaded in to anyway, and save a potential tlb or l1 miss. But I think it opens up inlining too.

compiler/runtime/obstack.c
skiplabsdaniel: It was just a test, and indeed, it's certainly not a good idea.

compiler/runtime/memory.c
gregsexton: >  But I think it opens up inlining too.

This was based on past experience, not anything the compiler must take care about, such as aliasing. I just checked out of curiosity. It will help gdb to inline. Clang 15 figures it out regardless.

compiler/runtime/obstack.c
jberdine: Not relevant, but to check my understanding: Globals in C are initialized to the 0 bit pattern by default, `__thread` doesn't change that, right?

gregsexton: Ah, you're right! I'm so paranoid about initialising everything that I never learned this.

PR #185, opened 2024-03-28
Update replication simulator for replacing resets with rebuilds

gregsexton:
The rebuild mechanism needs to be modelled. Here we enrich the
mechanism a little to allow this and review tests that no longer make
sense.

Depends on #182.

### DIFF
--- a/sql/test/replication/model.py
+++ b/sql/test/replication/model.py
@@ -351,7 +351,7 @@ class SkdbPeer:
         spec = { table: { "since": since } }
         payload, log, exitcode = await tail(db, session, peerId, spec)
         schedule.debug(log.decode().rstrip())
-        self.checkTailOutputExpectations(payload.decode())
+        self.checkTailOutputExpectations(payload.decode(), since)
         stream.send(schedule, payload)
         schedule.storeScheduleLocal(sinceKey, getOurLastCheckpoint(since, payload.decode()))
 
@@ -360,7 +360,7 @@ class SkdbPeer:
         if exitcode > 0:
           payload, log, exitcode = await tail(db, session, peerId, { table: { "since": 0 } })
           schedule.debug(log.decode().rstrip())
-          self.checkTailOutputExpectations(payload.decode())
+          self.checkTailOutputExpectations(payload.decode(), since)
           stream.send(schedule, payload)
           schedule.storeScheduleLocal(sinceKey, getOurLastCheckpoint(since, payload.decode()))
 
@@ -408,8 +408,9 @@ class Server(SkdbPeer):
   def startWriteCsv(self, *args):
     return startStreamingWriteCsv(*args, enableRebuilds=False)
 
-  def checkTailOutputExpectations(self, output):
-    pass
+  def checkTailOutputExpectations(self, output, since):
+    if since == 0 and "rebuild" in output:
+      raise AssertionError(f"Found rebuild when tailing from 0")
 
 class Client(SkdbPeer):
   def initTask(self) -> Task:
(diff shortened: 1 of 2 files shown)


gregsexton:
I was able to remove ~half of the tests(!!) due to replacing resets with rebuilds. This is quite a win as the simulation now runs in only 30 mins.

sql/server/dev/src/main/kotlin/Service.kt
```diff
@@ -292,10 +292,10 @@ fun connectionHandler(
                                 try {
                                   handler.set(handler.get().handleMessage(data, stream))
                                 } catch (ex: RevealableException) {
-                                  System.err.println("Exception occurred: ${ex}")
+                                  ex.printStackTrace(System.err)
```
bennostein: Beautiful, this will be very helpful

sql/src/SqlEval.sk
```diff
@@ -1683,6 +1683,21 @@ fun evalWithContext(
...
 }
 
+fun sqlPurgeFrom(
+  replicationId: Int,
+): (mutable SKStore.Context, SKStore.Dir) ~> SKStore.Tick {
+  (ctx, dir) ~> {
+    table = getSqlTableName(dir.getDirName());
+    ackTable = feedbackTable(table);
+    watermark = getWatermark(ctx, ackTable, Some(replicationId));
+    default = SKStore.Tick(ctx.tick.value - 1000);
```
bennostein: Is there reasoning for this choice of constant or just a best-guess for a reasonable amount to purge?

Also, I _think_ that everything should be fine with negative ticks, but am not 100% confident on that -- are you? If not, maybe we could take the max with 0 here

sql/src/SqlPrivacy.sk
```diff
@@ -549,13 +564,13 @@ fun checkUserCanWriteRow(
...
       groupIndex = skdbGroups.getIndex(context);
-      adminID = groupIndex.get(skdbGroups.makeKey(groupID)) match {
-      | Some(r) -> r.getString(adminIDColName)
-      | None() -> throw SqlError(0, "malformed predefined table skdb_groups")
+      groupIndex.get(skdbGroups.makeKey(groupID)) match {
+      | Some(r) ->
+        adminID = r.getString(adminIDColName);
+        (row.repeat == 0 && access.canDelete(context, adminID)) ||
+          access.canInsert(context, adminID)
+      | None() -> return AROK()
```
bennostein: Hmm, this should never happen if we've properly initialized privacy tables, but it makes me a little uncomfortable for the default behaviour to be permissive instead of erroring out if we reach this case.

sql/test/replication/model.py
```diff
@@ -369,15 +387,18 @@ async def stop(schedule):
...
+          proc.stdin.write(payload)
+          await proc.stdin.drain()
 
-        expectAck = payload.decode().startswith("^")
-        while expectAck:
-          ack = await proc.stdout.readline()
-          if ack.decode().startswith(":"):
-            return
+          payload = payload.decode()
+          expectAck = payload.startswith("^")  and not "!rebuild" in payload
```
bennostein, line 395: (Just a comment, I know this is reasonable/idiomatic python) but `and not "!rebuild" in payload` syntax is unhinged

bennostein (approved):
Looks good to me, just a few minor comments -- awesome that we can delete all of those tests!  I'll try to review #182 as well this afternoon.

bennostein:
Actually, just started looking at #182 and there seems to be a bit of overlap between there and here.  Am I missing some context on how they relate to each other, or should I just review independently?

gregsexton:
One is just stacked on the other. Sorry, I try to avoid this, but couldn't in this case. So the prefix of this one's commits is #182.

sql/src/SqlEval.sk
gregsexton: This has been the logic forever, so I'm pretty confident. Julien picked 1k out of thin air to be honest. We need to do some measurement at some point.

Note this just sets the tombLimit for a dir. We don't end up with negative ticks in the data. We also have a check during purge that the limit is monotonic.

sql/src/SqlPrivacy.sk
gregsexton: It happens now because I made group creation atomic. I had to break the cyclic dependencies. So I reasoned it's ok to create permissions for groups that don't exist yet.

gregsexton: (Because anyone can create a group and you obviously won't create a group that already has permissions you don't want and can't change.)

PR #186, opened 2024-03-28
[stdlib] Make use of question mark operator in IO.

beauby:


jberdine (approved):
`?` does indeed help readability :-)

gregsexton:
+1. I love it.

PR #189, opened 2024-04-04
Fix: dumping indices broken if we make use of an index first

gregsexton:
Looks like a simple copy paste error that has been around forever. Due
to getIndexByColNbr and getIndexByName both using the same dir name,
there's a race to see who gets there first. If getIndexByColNbr wins
then we don't dump indices, otherwise we (probably, I didn't test)
don't use indices.


beauby (approved):
LGTM.

PR #190, opened 2024-04-04
Force a rebuild when a client reconnects to a new server

gregsexton:
Why? Because a new server loses watermarks and has a different tick.
This breaks all the metadata that replication relies on. The rebuild
mechanism is fundamentally: reset your view of the server's clock and
resync your data to this. So it is a perfect fit to solve this
problem.

I disregarded trying to port the watermark data and initialise a new
server with the same tick as the old. We would also have to dump and
import all the Source data for each and every row. This is a much
bigger lift and all becomes irrelevant with upcoming vector clocks.

### DIFF
--- a/sql/src/SqlCsv.sk
+++ b/sql/src/SqlCsv.sk
@@ -715,7 +715,7 @@ fun replayDiff(
   identity: Int,
   rebuildsEnabled: Bool,
 ): SKStore.ContextOp {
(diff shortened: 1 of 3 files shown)


gregsexton:
CI is green now. I'm pretty sure this PR isn't responsible for any errors - so shipping. But have repro'd the flaky wasm test and will see if I can fix it.

PR #192, opened 2024-04-08
Clean prelude/libbacktrace more thoroughly between cross-compiles

bennostein:


bennostein:
This may be needlessly aggressive but it fixes the cross-compile, and nuking prelude build artifacts when you nuke `./compiler` sounds right to me.

compiler/Makefile
```diff
@@ -96,7 +96,8 @@ install: stage$(STAGE)
 
 .PHONY: clean
 clean:
-	rm -Rf stage* bootstrap/*.ll
+	OUT_DIR=$$(realpath ./bootstrap) $(MAKE) -C ../prelude clean
+	rm -Rf stage* bootstrap/*.ll bootstrap/libbacktrace.a
```
jberdine, line 100: Previously we also had here:
```
$(MAKE) -C runtime clean
```
and the definition of that rule was:
```
clean:
	rm -f *.o *.bc magic.c libskip_runtime64.a
	(test ! -f libbacktrace/Makefile && echo "Warning: libbacktrace submodule not initialized") || $(MAKE) -C libbacktrace clean
```
I'm not sure how much cleaning prelude does.

bennostein: Notably includes the following:

https://github.com/SkipLabs/skdb/blob/534d3626ecaf37a1d59aac90763b235ca8823805/prelude/Makefile#L82

PR #196, opened 2024-04-09
Fix flaky test undefined property TypeError

gregsexton:
```
      376 |       root,
      377 |       "select y from test_pk_string where x = 0 and y = @str",
    > 378 |       (tail) => tail[0].y == str,
          |                         ^
      379 |       { str },
      380 |       true,
      381 |     );
```

We would get this error very occassionally. The waitSynch call is
testing to see if the data has arrived yet, but if it hasn't the test
fails: `tail[0]` returns undefined, accessing the property throws a
TypeError.

Fixed this and other occurrences of the problem.


bennostein:
LGTM!

PR #199, opened 2024-04-10
Ensure multiple `select id()` produce differing results

gregsexton:
The gist is we need to ensure re-evaluation as the id function is
non-deterministic.

The fix is for when there is no from clause as this seems to be
somewhat special cased. We already re-evaluate when there is a from
clause, both when from is a query or a table.

### DIFF
--- a/sql/src/SqlSelect.sk
+++ b/sql/src/SqlSelect.sk
@@ -694,27 +694,31 @@ class SelectEvaluator{
   }
 
   fun evalEmptyFrom(context: mutable SKStore.Context): void {
-    dirName = this.selectName2;
-    if (context.unsafeMaybeGetEagerDir(dirName) is None()) {
-      evaluator = ExprEvaluator(Array[], this.select.from, this.up);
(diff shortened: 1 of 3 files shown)


gregsexton:
Something that was surfaced to me during this is that we persist the result of every unique query evaluation without a from clause. This must be happening and I would bet that it's because the directory we create in `evalEmptyFrom` is an input directory and so gets merged in to the root context on import/export.

If people agree this is an issue I'll file a task.

bennostein:
Looks solid to me.  If we add some SQL constant folding machinery it could be useful here to distinguish between cacheable and non-cacheable results but that'd just be an optimization.

As for persisting every query evaluation without a from clause: I'd agree that this is an issue in that it's probably unnecessary to cache all of that.  But, it's probably a negligible overhead the vast majority of the time, so not the highest-pri fix if I understand correctly.

gregsexton:
I'm going to ship to unblock a quick fix, but very happy to revisit this and address any concerns.

PR #200, opened 2024-04-10
[skbuild] Fix wasm build with c files

skiplabsdaniel:
Add some default flags


beauby (approved):
LGTM. Maybe add skbuild to the list of directories that trigger CI for skdb (in .circleci/)?

PR #201, opened 2024-04-10
[skmonitor] Add monitoring API

skiplabsdaniel:


prelude/runtime/runtime64_specific.cpp
```diff
@@ -448,11 +457,15 @@ int32_t SKIP_notify(char* filename_obj, int32_t tick) {
   return 0;
 }
 
+bool random_initialized = false;
```
jberdine: Maybe make this a static local of SKIP_random_init

gregsexton:
Bear with me here, I do plan to look at this.

prelude/runtime/runtime64_specific.cpp
```diff
@@ -306,6 +310,11 @@ int main(int pargc, char** pargv) {
   SKIP_initializeSkip();
   sk_persist_consts();
   SKIP_destroy_Obstack(saved);
+  if (psession->low == 0 || psession->high == 0) {
```
gregsexton: && ?

prelude/runtime/obstack.c
```diff
@@ -75,12 +93,29 @@ typedef struct sk_obstack {
...
+  sk_saved_obstack_t obstack;
+  sk_saved_obstack_t switched;
+} sk_trace_t;
+
 #ifdef SKIP32
 sk_saved_obstack_t init_saved = {NULL, NULL, NULL};
+sk_trace_t trace = {NULL, {NULL, NULL, NULL}, {NULL, NULL, NULL}};
 #else
 __thread sk_saved_obstack_t init_saved = {NULL, NULL, NULL};
+__thread sk_trace_t trace = {NULL, {NULL, NULL, NULL}, {NULL, NULL, NULL}};
```
gregsexton: You can just `trace = {}` to zero init all at once.

prelude/runtime/obstack.c
```diff
@@ -201,6 +241,65 @@ sk_saved_obstack_t* sk_saved_obstack(sk_obstack_t* page) {
...
+  trace_page->saved.head = NULL;
+  trace_page->saved.end = NULL;
+  trace_page->saved.page = NULL;
+  trace.main = trace_page;
+  trace.obstack.page = trace_page;
+  trace.obstack.end = (char*)trace_page + block_size;
+  trace.obstack.head = trace_page->user_data;
+}
+
+char* SKIP_duplicate(char* skstr) {
```
gregsexton: Is this SKIP_copy_string with a null large_page param?

sql/src/Monitor.sk
```diff
@@ -0,0 +1,123 @@
...
+}
+
+fun read_to_end(): String {
+  text = .read_to_end();
+  SKDBMonitor.addBytesRead(byteSize(text));
+  text
+}
+
+fun read_line(): ?String {
+  .read_line().map(line -> {
```
gregsexton, line 179: Will this end up copying?

sql/src/Monitor.sk
```diff
@@ -0,0 +1,123 @@
...
+  SKDB.print_error(error)
+}
+
+fun print_raw(str: String): void {
+  SKDB.print_raw(str)
+}
+
+fun print_string<T: readonly Show>(value: T): void {
+  SKDB.print_string(value)
+}
```
gregsexton, line 207: Did you intend to capture stuff here?

sql/src/SqlCsv.sk
```diff
@@ -807,6 +807,7 @@ mutable class DiffParsingState(
 )
 
 fun nack(writer: mutable Debug.BufferedWriter, diff: FinalTableDiff): void {
+  SKDBMonitor.addReplicationWrites(1 + diff.lines.size());
```
gregsexton: It's a little confusing to call these writes - although I see why you did. These are 'rejected rows'. And I would rename reads to be 'processed rows' perhaps.

sql/src/SqlCsv.sk
```diff
@@ -824,6 +825,7 @@ fun replayDiff(
 ): SKStore.ContextOp {
   SKDB.getReplicationState(context, Some(identity)) match {
   | Some(SKDB.RSError()) ->
+    SKDBMonitor.addReplicationWrites(1);
```
gregsexton: It would be really valuable to break this out and measure explicitly # of reboots. This should be a rare event that we want to minimise.

sql/src/SqlCsv.sk
```diff
@@ -844,6 +846,7 @@ fun replayDiff(
         None()
 
       | None() ->
+        SKDBMonitor.addReplicationReads(1);
```
gregsexton: This is not a row to be processed but a table. Will inflate the metric if we're interested in rows. Be nice to capture as a different one so we can see how many different tables are touched in a single txn.

sql/src/SqlCsv.sk
```diff
@@ -856,6 +859,7 @@ fun replayDiff(
       | Some(
         DiffParsingState(tableDiffs, currentDiff),
       ) if (String.getByte(line, 0) == 94) -> // startsWith("^")
+        SKDBMonitor.addReplicationReads(1);
```
gregsexton: Switching to a different table.

sql/src/SqlCsv.sk
```diff
@@ -869,6 +873,7 @@ fun replayDiff(
       | Some(
         DiffParsingState(tableDiffs, currentDiff),
       ) if (String.getByte(line, 0) == 58) -> // startsWith(":")
+        SKDBMonitor.addReplicationReads(1);
```
gregsexton: Processing the whole lot. Could record this as 'handling a batch'.

sql/src/SqlCsv.sk
```diff
@@ -885,14 +890,15 @@ fun replayDiff(
       | current @ Some(DiffParsingState(_, currentDiff)) if (
         line == "!rebuild"
       ) ->
+        SKDBMonitor.addReplicationReads(1);
```
gregsexton: Worth capturing rebuilds as a separate metric. Obviously we want to minimise these.

sql/src/SqlCsv.sk
```diff
@@ -885,14 +890,15 @@ fun replayDiff(
...
         if (!rebuildsEnabled) {
           SKDB.setReplicationState(context, identity, SKDB.RSError());
           invariant_violation("Received rebuild when processing disabled.")
         };
         currentDiff.!rebuild = true;
         current
-
       | current @ Some(DiffParsingState(_, currentDiff)) ->
+        SKDBMonitor.addReplicationReads(1);
```
gregsexton: This is a row update.

sql/src/Main.sk
```diff
@@ -653,6 +681,7 @@ fun execReplicationId(args: Cli.ParseResults, options: SKDB.Options): void {
 }
 
 fun execWriteCsv(args: Cli.ParseResults, options: SKDB.Options): void {
+  SKDBMonitor.setName("skdb.write.csv");
   ensureContext(args);
```
gregsexton, line 706: Do we set the user in the monitoring for write csv?

skmonitor/src/Id.sk
```diff
@@ -0,0 +1,148 @@
+module SKMonitor;
+
+const kINVALID_TRACE_ID: String = "00000000000000000000000000000000";
+const kINVALID_SPAN_ID: String = "0000000000000000";
+
+base class Id private () {
```
gregsexton: Is it possible to use any of the KSUID stuff that Josh built? This would give us unique ids that sort by time - pretty useful for logging.

skmonitor/src/Monitor.sk
```diff
@@ -0,0 +1,398 @@
+module SKMonitor;
+
+base class Monitoring() {
+  children =
+  | MOn
+  | MMemory
+  | MNo
```
gregsexton: MOff? I was initially confused.

skmonitor/extern/native.c
```diff
@@ -0,0 +1,58 @@
...
+  if (fifo_write < 0) {
+    return 0;
+  }
+  char* b = trace;
+  const char* s;
+  for (s = trace; *s; ++s)
+    ;
+  size_t n = (s - trace);
+  size_t r = 0;
+  sk_global_lock();
```
gregsexton: Why do we need to do this? Do different processes write to the same fifo?

skmonitor/src/Monitor.sk
```diff
@@ -0,0 +1,398 @@
...
+    None()
+  };
+  span = mutable Span(
+    now(),
+    traceId,
+    SpanId::fromInt(Random.next()),
+    optParent,
+  );
+  set_span(Some(span));
+  SKStore.restoreFromTrace();
```
gregsexton: Should we put restore in a finally block? Same for elsewhere obviously.

skmonitor/src/Monitor.sk
```diff
@@ -0,0 +1,398 @@
...
+}
+
+fun fatal(msg: String, meta: Logger.Metadata): void {
+  log(Fatal(), msg, meta)
+}
+
+fun flush(span: readonly Span): void {
+  trace = span.toJSON().toString();
+  get_writer() match {
+  | Some(writer) -> writer(`${trace}\n`)
```
gregsexton: This will copy the string right? Can we call writer twice? There might be other opportunities like this.

skmonitor/src/Monitor.sk
```diff
@@ -0,0 +1,398 @@
...
+      span.!optEnd = Some(now());
+      if (span.optParent.isNone()) {
+        memory = memory();
+        span.setIntAttribute("skruntime.memory.persistent", memory.persistent);
+        span.setIntAttribute("skruntime.memory.freetable", memory.freetable);
+        span.setIntAttribute("skruntime.obstack.pick", memory.obstack_pick);
+      };
+      impact = get_trace_size(span.optParent.map(p -> p.i1));
+      span.setIntAttribute("skmonitor.memory", impact);
+      flush(span);
```
gregsexton: So this is when we write information out, at the grain of a span. Totally reasonable. But, do I understand correctly: we learn about events when the span ends, and we'll get a parent id, but we won't learn any info about that parent until its span ends? For a long lived process like write-csv if the parent span lasts the whole time of the process and contains e.g. the user we won't be able to figure out who's affected until that process ends?

This all seems reasonable, I just wanted to check, and to point out we need to be careful about what info we store where and when we'll get it.

skmonitor/extern/native.c
```diff
@@ -0,0 +1,58 @@
...
+    return 1;
+  }
+  if (fifo != skfifo) free(fifo);
+  return 0;
+}
+
+__uint8_t SKIP_SKMonitor_appendToFifo(char* skfifo, char* sktrace) {
+  char* trace = sk2c_string(sktrace);
+  char* fifo = sk2c_string(skfifo);
+  int fifo_write = open(fifo, O_WRONLY | O_APPEND | O_NONBLOCK);
```
gregsexton: What are the semantics of nonblock with fifos? I think this just means you can write successfully if there's no reader. We'll still need to be careful that the reader can keep up otherwise we'll block the skdb processes, right? Or do we lose log events if the reader can't keep up and the OS pipe buffer fills?

gregsexton (commented):
This is great. I really like it.

Big +1 for using open telemetry - the data model (tree of spans, etc) looks solid to me, and we'll be able to utilise all the tools of that ecosystem.

Also, thanks for capturing all the memory info. That will be priceless.

What are your thoughts on the client? Due to the nature of the product, a lot of action is happening there. Should we log at all? Should we just expose to the user or should we try and capture this info somehow? Maybe make it opt in or out?

gregsexton:
And just one other thing: we don't have any logging in the production side. So on the server it's SqlTailer.sk and on the client it's the combination of update and diff. Without this we can only see what was processed, but not what was sent.

rfc/007-monitoring.org
```diff
@@ -0,0 +1,231 @@
...
+** Writer
+
+By default the monitoring trace is written to stderr, a SK_MONITORING_WRITER environment variable can be defined with following value:
+
+- 'fifo' to write the trace into '/tmp/skmonitor.trace' fifo.
+- 'file' to write the trace into 'skmonitor.log' file.
+- 'fifo:<name>' to write the trace into '<name>' fifo.
+- 'file:<name>' to write the trace into '<name>' file.
+
+** Collector (TO BE DONE)
```
gregsexton: If we're producing to the open telemetry spec, there must be stuff out there that can do this for us?

prelude/runtime/obstack.c
skiplabsdaniel: Yes

rfc/007-monitoring.org
skiplabsdaniel: OpenTelemetry provides tools for data collection and processing. However, to be able to extract metrics from the attributes of the spans, it is necessary to create a custom collector.
With this in mind, to limit the amount of trace data generated on skdb side, I opted for a format similar to but lighter than OpenTelemetry one.

skmonitor/extern/native.c
skiplabsdaniel: In order to avoid creating a fifo management mechanism, it's true that in the first version, concurrent processes will probably write to the same fifo.

skmonitor/src/Monitor.sk
skiplabsdaniel: I don't think so, there should never have exceptions during trace calls.

sql/src/Monitor.sk
skiplabsdaniel: We need to manage the skipExit for clean exit and readline to get the number of read bytes.
The rest is to be belt and braces.

skmonitor/extern/native.c
skiplabsdaniel: Good point.
It seems that in non-blocking mode, however, write() will write as much data into the file as it can, and then return. So we lose log events if the reader can't keep up and we can have some partial log events.

gregsexton: That works for me. I'd rather lose log events than stall.

gregsexton: e.g. imagine a load shedding situation producing a huge volume of error logs. We want to fail fast here and not stall because of the reader. Would be very bad for resiliency.

skmonitor/src/Monitor.sk
skiplabsdaniel: You understand correctly. I had the same thought when I looked at OpenTelemetry.
Like in OpenTelemetry Protocol File Exporter.
- There is no guarantee that the data is ordered.
- There is no guarantee in particular that timestamps will be monotonically increasing.
Some data could be added the children to identify more precisely the parent.

If we need to have some kind validation on span tree, it could be done in the "Collector"

skmonitor/extern/native.c
gregsexton: I'm ok with this for first version, but I feel strongly that we should prioritise avoiding the lock. Logging can be very high throughput and this would then introduce a major scaling bottleneck.

sql/src/Monitor.sk
skiplabsdaniel: I don't think so.

sql/src/Main.sk
skiplabsdaniel: Now, no.
This is one of the questions I ask myself in terms of safety, but I don't really have an answer.
You certainly have more experience than I do on this subject.

skmonitor/src/Id.sk
skiplabsdaniel: That's what I wanted to do.
Unfortunately, KSUID does not follow the TraceId or SpanId specifications.

sql/src/Main.sk
gregsexton: I don't think it hurts to log who the user is. Obviously logging the data is much riskier and we shouldn't do that. But knowing the user could really help with figuring out errors. And in terms of safety we probably really want to know who is interacting with databases.

prelude/runtime/obstack.c
```diff
@@ -75,12 +93,29 @@ typedef struct sk_obstack {
...
+
 #ifdef SKIP32
 sk_saved_obstack_t init_saved = {NULL, NULL, NULL};
+sk_trace_t trace = {};
 #else
 __thread sk_saved_obstack_t init_saved = {NULL, NULL, NULL};
+__thread sk_trace_t trace = {};
 #endif
 
+uint64_t SKIP_obstack_peack() {
```
jberdine: I guess s/peak/peack/

jberdine (commented):
Looks good! Just a couple tiny comments

skiplabsdaniel:
On the client side, I think we need to allow it with options for sending data to ours or/and app developer monitoring servers.
We can add the possibility to register the log into IndexedDB if the final user don't allow data sending.

skiplabsdaniel:
Redesign underway, new Pr to follow

PR #203, opened 2024-04-11
[skc] Fix `Could not open file` in `catchErrors()`.

### LINKED ISSUE #198: [skargo] Syntax errors produce misleading error messages
When the compiler frontend produces errors, `skargo build` is warning of "No such file or directory" instead of reporting the `skc` error messages.

e.g. if I introduce a syntax error

```
diff --git a/sql/src/Main.sk b/sql/src/Main.sk
index e87e010c..77ef4d71 100644
--- a/sql/src/Main.sk
+++ b/sql/src/Main.sk
@@ -763,3 +763,4 @@ fun execToggleView(args: Cli.ParseResults, _options: SKDB.Options): void {
 }
 
 module end;
+syntax error
```

and run `skc --check sql/src/Main.sk` then I get the expected

```
File "sql/src/Main.sk", line 766, characters 1-6:
Expected declaration.
764 |
765 | module end;
766 | syntax error
    | ^^^^^^
```

but if I run `(cd sql && skargo build --release)` then I get the following

```
    Fresh std v0.1.0 [build-script-std] (/skdb/prelude)
    Fresh std v0.1.0 [lib] (/skdb/prelude)
    Fresh skbuild v0.1.0 [lib] (/skdb/skbuild)
    Fresh skdate v0.1.0 [build-script-skdate] (/skdb/skdate)
    Fresh skdate v0.1.0 [lib] (/skdb/skdate)
    Fresh cli v0.1.0 [lib] (/skdb/cli)
    Fresh skjson v0.1.0 [lib] (/skdb/skjson)
    Fresh sqlparser v0.1.0 [lib] (/skdb/sqlparser)
    Compiling skdb v0.1.0 [lib] (/skdb/sql)
error command exited with non-zero status

Caused by:
  process did not exit successfully: `skc` (exit status: 45)
  --- stdout

  --- stderr
  ERROR (open failed): No such file or directory
  Could not open file: skdb:/skdb/sql/src/Main.sk
```

Can of course be worked around by running `skc --check` when needed, but curious if others are seeing the same behaviour.

beauby:
Fixes #198.

### DIFF
--- a/compiler/src/skipError.sk
+++ b/compiler/src/skipError.sk
@@ -91,8 +91,12 @@ fun throwFlattenedErrors(errors: Vector<Vector<Error>>): void {
 
 fun printErrorsAndExit<T>(
   errors: Array<SkipErrorException>,
-  get_file: String -> (String, String) = filename ->
-    (filename, FileSystem.readTextFile(filename)),
+  get_file: String -> (String, String) = filename -> {
+    if (filename.contains(":")) {
+      !filename = filename.splitFirst(":").i1
+    };
+    (filename, FileSystem.readTextFile(filename))
+  },
 ): T {
   SkipError.printErrors(
     errors.map(x -> x.errors).flatten().collect(Vector),


compiler/src/skipError.sk
```diff
@@ -91,8 +91,12 @@ fun throwFlattenedErrors(errors: Vector<Vector<Error>>): void {
 
 fun printErrorsAndExit<T>(
   errors: Array<SkipErrorException>,
-  get_file: String -> (String, String) = filename ->
-    (filename, FileSystem.readTextFile(filename)),
+  get_file: String -> (String, String) = filename -> {
+    if (filename.contains(":")) {
+      !filename = filename.splitFirst(":")[1]
```
bennostein: should be `.i1` rather than `[1]` I think

compiler/src/skipError.sk
```diff
@@ -91,8 +91,12 @@ fun throwFlattenedErrors(errors: Vector<Vector<Error>>): void {
 
 fun printErrorsAndExit<T>(
   errors: Array<SkipErrorException>,
-  get_file: String -> (String, String) = filename ->
-    (filename, FileSystem.readTextFile(filename)),
+  get_file: String -> (String, String) = filename -> {
+    if (filename.contains(":")) {
```
bennostein, line 95: It's not common or a good idea, but `:` is legal in unix file/directory names... this could lead to some weird behaviours if we encounter that.  Could we match something a little stricter for this case?

beauby: My bad, it's up to date now.

compiler/src/skipError.sk
beauby: My bad, it's up to date now.

compiler/src/skipError.sk
beauby: Yes, the current encoding of `(?lib, path)` as `lib:path` is a bit fragile (though it shouldn't be an issue as long as we don't use `:` in file names). Doing this correctly is on my TODO list but it requires quite a few changes in the compiler.

bennostein (approved):
Confirmed all good now with this fix -- thanks!

PR #206, opened 2024-04-15
Reconnect sktest on wasm

skiplabsdaniel:


gregsexton:
What's this change about?

skiplabsdaniel:
It's to allow to make the equivalent of `skargo test` in wasm through `sknpm`.
It's used in skui to run all the font tests.

PR #208, opened 2024-04-17
[test] Fix incorrect size of SKIP_utf8_test_string

jberdine:
This bug is only triggered if string.c is compiled with -O0.

### DIFF
--- a/prelude/runtime/string.c
+++ b/prelude/runtime/string.c
@@ -359,8 +359,7 @@ static unsigned int test_utf8_data[] = {
     157, 153, 137, 224, 167, 166, 206, 161, 240, 157, 151, 164, 201,
     140, 240, 157, 147, 162, 200, 154, 208, 166, 240, 157, 146, 177};
 
-static unsigned char
-    string_utf8_buffer[sizeof(test_utf8_data) / sizeof(int) + 1];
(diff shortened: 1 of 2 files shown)


jberdine:
That worrying feeling when you see a difference in test results when changing optimization levels, only to eventually find that it is just a bug in the test code itself...

gregsexton:
Do you know what's happening? I'm guessing -O3 changes the alignment so a 0 is copied over?

jberdine:
I don't know exactly, but it looked something along those lines. The code on main reads past the end of `test_utf8_data` when populating `string_utf8_buffer`. When compiled with -O0 I was seeing that extra byte hold value 255, which breaks the utf8 decoding, while with higher optimization more values are statically resolved and laid out differently, maybe also due to alignment, but that extra byte was 0 and just got dropped.

But the expected results in the tests were just wrong. The byte length is easy as the number of integers in the static initializer of the array can be counted 5 * 13 = 65, not 66, and I did decode them as utf8 to ~𝘈Ḇ𝖢𝕯٤ḞԍНǏ𝙅ƘԸⲘ𝙉০Ρ𝗤Ɍ𝓢ȚЦ𝒱 which does indeed have length 23 not 24.

PR #209, opened 2024-04-17
Rework can-mirror to locally check schemas and fail fast

bennostein:
`can-mirror` now explicitly compares a remote table/schema against the provided schema, instead of relying on the existing local table's schema.

The previous behavior delayed/obscured failure in some cases, causing the client to hang since failures were happening on the remote.  Now, there should be a local error if a user attempts to mirror a table with an incorrect or incompatible schema.

### DIFF
--- a/sql/src/Main.sk
+++ b/sql/src/Main.sk
@@ -281,8 +281,15 @@ untracked fun main(): void {
     )
     .subcommand(
       Cli.Command("can-mirror")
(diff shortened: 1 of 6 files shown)


gregsexton:
LGTM! Thanks for working on this and fixing so quickly. Would it be possible to paste in that test I gave you so that we don't accidentally regress?

PR #210, opened 2024-04-18
[runtime] Use `mmap()` even in non-persistent mode.

beauby:
This PR homogenizes the memory allocators between persistent and non-persistent modes.

### DIFF
--- a/prelude/runtime/palloc.c
+++ b/prelude/runtime/palloc.c
@@ -26,8 +26,6 @@
 
 void*** pconsts = NULL;
 
-void* program_break = NULL;
-
 /*****************************************************************************/
 /* Database capacity. */
 /*****************************************************************************/
@@ -66,7 +64,6 @@ typedef struct ginfo {
   char* head;
   char* end;
   char* fileName;
-  char* break_ptr;
   size_t total_palloc_size;
 } ginfo_t;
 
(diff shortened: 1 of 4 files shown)


gregsexton (approved):
LGTM

prelude/runtime/palloc.c
```diff
@@ -496,22 +497,23 @@ void sk_create_mapping(char* fileName, char* static_limit, size_t icapacity) {
     exit(ERROR_MAPPING_MEMORY);
   }
 
-  memcpy(persistent_fileName, fileName, fileName_length);
+  if (fileName != NULL) {
+    memcpy(persistent_fileName, fileName, fileName_length);
+  }
```
jberdine: Probably safer to init persistent_fileName to the empty string in the else case, even though it is not meaningful, it will stop string functions from walking memory if called without checking then enclosing ginfo->fileName for NULL.

jberdine (approved):

prelude/runtime/palloc.c
beauby: I agree on the principle but at the moment we use `ginfo->fileName` being `NULL` as a proxy for checking whether we are running in persistent mode.

jberdine: I didn't mean to change this diff's way of setting ginfo->fileName, just to change the hopefully-dead value in persistent_fileName when ginfo->fileName is NULL from something uninit to 0.

prelude/runtime/palloc.c
```diff
@@ -690,14 +696,6 @@ void SKIP_print_persistent_size() {
 }
 
 void* sk_palloc(size_t size) {
-  if (ginfo->fileName == NULL) {
```
mbouaziz, line 693: Will this still work with `__APPLE__` defined?

beauby: Good question. OSX support is currently broken for other reasons (and has not been an explicit goal so far). I will try to revive it once current priorities are settled.

PR #211, opened 2024-04-18
Implement `CHECK` constraints

gregsexton:
See RFC 6 for detail on what is being implemented here.

The gist is we have the possibility of an expression that can induce a
side effect - cause failure - and this can be used as a top-level
parameter in a SELECT statement.

Combined with reactive views, this allows efficiently enforcing data
invariants during any modification.


sql/src/SqlCsv.sk
```diff
@@ -798,13 +798,13 @@ fun buildDiffChain(
         }
       }
     } catch {
-    | PrivacyWriteViolation _ ->
+    | exn ->
```
bennostein, line 801: Could we keep this specifying exception types and just include SqlError here? (or even introduce something more specific like `SqlConstraintViolation`)

I worry that issuing nacks for unrelated server faults could cause confusion/issues later on...

sql/src/SqlCompile.sk
```diff
@@ -567,15 +569,31 @@ mutable class Compiler{
...
   }
 
+  mutable fun compileSideEffectParams(
+    context: mutable SKStore.Context,
+    selectParams: Array<P.SelectResult>,
+  ): Array<CGExpr> {
+    compiledExprs = mutable Vector[];
+    for (_ => param in selectParams) {
+      param match {
+      | P.SelectExpr(check @ P.Check(_), alias) ->
```
bennostein: I might be missing something, but shouldn't we either (a) recurse into other expression forms' subexpressions here or (b) forbid Check expressions anywhere except the top level of a select param?

As is, you could write `SELECT 1 + CHECK(0);` or similar and not pick up the side effect as it's hidden in the operand of a BinOp node.

bennostein (approved):
A couple small comments but looks great overall.  Nice test coverage too!

sql/src/SqlCsv.sk
gregsexton: Will change this. When I wrote this I was of the opinion that we should issue nacks no matter what, but I've changed my mind and agree with you.

gregsexton: TBH, not sure what I was thinking :D

sql/src/SqlCompile.sk
gregsexton: I'll check this (no pun intended). I'm pretty sure `1 + CHECK(0)` will fail to evaluate. This is why I introduced ASuccess() and have the throw clauses everywhere.

bennostein: I checked it on your branch while reviewing, and get the following output:
```
skdb> select 1 + check(0);
select 1 + check(0);
^
|
 ----- ERROR
Error: line 1, character 0:
Failed constraint: 0
```

gregsexton: That's disappointing. `select 1 + check(1)` doesn't fail either - I was expecting that to produce ASuccess() which BinOp would hopefully choke on. I'll dig in and see what's happening.

gregsexton: Fixed. This was really dumb. I introduced ASuccess for exactly this reason: to deal with CHECK expressions as subexpressions. But I didn't fail on the ASuccess() branch. 🤦‍♂️

bennostein: Out of curiosity: why make this a runtime error instead of a static compile-time check?  Since it's a purely syntactic property, we ought to be able to just inspect the AST upfront instead of baking it into the semantics like this.

The current setup produces some (to me) unintuitive behaviour discrepancy between passing and failing checks in subexpressions.

i.e.
```
skdb> select 1 + check(1);
select 1 + check(1);
^
|
 ----- ERROR
Error: line 1, character 0:
SELECT error: parameter expression contains a side effect
skdb> select 1 + check(0);
select 1 + check(0);
^
|
 ----- ERROR
Error: line 1, character 0:
Failed constraint: 0
```

(sorry not to elaborate on/express this concern more clearly in my initial comments... feel free to ignore if it's a pain to change the dynamic checks out for static ones)

gregsexton: > Out of curiosity: why make this a runtime error instead of a static compile-time check?

Good question. To me it's all runtime as the compilation happens at runtime. So I didn't give it much thought. My intuition said handle at eval as that's a deeper more fundamental layer and so will stand up to code change better, rather than putting in a frontend check and assuming afterward all is good and validated.

But...

> The current setup produces some (to me) unintuitive behaviour discrepancy between passing and failing checks in subexpressions.

This is legit.

I'll take a look at introducing compile time checking.

gregsexton: Yep, np, here we go:

```
# ./build/skdb <<< "select 1 + check(1);"
select 1 + check(1);
^
|
 ----- ERROR
Error: line 1, character 0:
CHECK expressions cannot be nested
1 # ./build/skdb <<< "select 1 + check(0);"
select 1 + check(0);
^
|
 ----- ERROR
Error: line 1, character 0:
CHECK expressions cannot be nested
1 # ./build/skdb <<< "select check(0);"
select check(0);
^
|
 ----- ERROR
Error: line 1, character 0:
Failed constraint: 0
1 # ./build/skdb <<< "select check(1);"
# 
```

I'll push the commit in a sec. Best of both worlds, we have the compile time check and this is also handled at eval time just in case.

PR #212, opened 2024-04-22
[runtime][build] Obey specified OLEVEL when PROFILE missing

jberdine:
and remove duplicate `-g3` flags.


jberdine:
I think this preserves current behavior when building with skargo, but enables building the runtime directly with e.g. `OLEVEL=-O0 make ...` to build the runtime with optimizations disabled for debugging.

PR #213, opened 2024-04-22
[runtime] Fix string address computation in SKIP_posix_spawnp

jberdine:


jberdine:
I don't understand why we haven't hit this before, but when combined with my runtime changes, this leads to calling free with a pointer into the binary, eventually causing `munmap_chunk(): invalid pointer`. I guess that with the changes to support sklibs, there are some string args passed to spawn that are static strings now.

gregsexton:
LGTM.

gregsexton:
I remember I ran this code with valgrind memcheck and it didn't complain, which it definitely would have if we were trying to free bad addresses :D Could it be that we only ever used this with one arg in the array and so the logic worked?

jberdine:
I don't think that we have only used it with one arg, the diff for sklib support around the call that crashes is:
```
     runShell(
       Array[
         "clang++",
         "-no-pie",
         `-O${config.optLevel}`,
         "-mllvm",
         "-inline-threshold=0",
-        llFile,
         "-o",
         config.output,
-      ].concat(config.links),
+        llFile,
+      ].concat(linker_args).concat(static_libs),
       config.verbose,
     );
```
and the arg index that trips the crash is 8, so there have always been multiple args, and static strings too. The only theory I have is that the allocation pattern changed so that `free` decided to unmap instead of linking a block into a free list or something similar. I don't have a good intuition for how this sort of thing is detected by valgrind etc.

gregsexton:
Ok, interesting. Looking at the logic again, I guess we weren't calling free rather than calling it with a bad address. I would expect valgrind or asan to complain about this, so not sure what happened there!

> I don't have a good intuition for how this sort of thing is detected by valgrind etc.

In short: malloc and free are intercepted so that all dynamic memory can be tracked and the runtime can error if you're trying to do bad things. There's usually a region of address space allocated that 'shadows' all the allocations with tracking metadata. I use asan more than valgrind these days. This adds all sorts of instrumentation at compile time and can catch more, is faster, etc. There's a bit more to this with tracking reads/writes on the stack, static vars, etc. You might like this paper: https://storage.googleapis.com/gweb-research2023-media/pubtools/pdf/37752.pdf

jberdine:
> I guess we weren't calling free rather than calling it with a bad address

I've been thinking that it could be due to previously getting lucky that all static strings encountered here had lengths that were not multiples of 8, and so had implicit nul termination, and hence were not re-allocated. I don't know why the sklib support changes and my string changes are each ok separately but not both together.

PR #214, opened 2024-04-25
[sql] Add support for "add column" schema changes

bennostein:
This PR implements SQL statements of the form
```sql
ALTER TABLE t ADD COLUMN <column>
```
under the restriction that `<column>` is either nullable or provides a default.

That restriction ensures that schema changes are backwards compatible and the orchestration layer can still communicate with clients on old schemas.

Views/joins/aggregates/etc. on the table will update as expected.

The schema migration happens in place and under a lock, and I haven't done any perf tests adding columns to large tables.  But, schema migration is a relatively infrequent operation and this has been designed to ensure that subsequent operations on the table pay no extra cost based on the table's schema history.


gregsexton:
Sorry, just seen this PR. Writing docs is exhausting and I need a break! Will look at this very soon.

bennostein:
Marking as ready for review again now; fixed the issue I found while working on extract-column and improved the test coverage to cover it with (1) views over views over altered tables and (2) multiple alter-tables in sequence.

Merge looks trivial, so I'll hold off on that until after review/edits.

sql/src/SqlTables.sk
```diff
@@ -357,6 +357,176 @@ fun getLatestIndexDescr(
...
+
+fun alterTableAddColumn(
+  context: mutable SKStore.Context,
+  pos: Int,
+  name: P.Name,
+  newCol: P.ColumnDefinition,
+  options: Options,
+): void {
+  if (
+    name == SKDBUsersByName::name ||
```
gregsexton: We should put these in a list somewhere

sql/src/SqlTables.sk
```diff
@@ -357,6 +357,176 @@ fun getLatestIndexDescr(
...
+      newKey = RowKey(newVals, newKinds);
+      newSource = SKStore.Path::create(dirName, newKey);
+
+      // Since schema/kinds information is encoded in SKDB.RowKeys, we have to both null
+      // out the old row(s) and write the new row. (instead of just overwriting the old key)
+      for (x in dir.unsafeGetAllDataIter(key)) {
+        (_tick, source, writer, _row) = x;
+        yield (key, (source, writer, Array[]))
+      };
+      yield (newKey, (newSource, newSource, Array[newVals]))
```
gregsexton, line 489: I'm scared of yield. I've seen some weird compiler issues with it and had to remove it from somewhere once (I forget now). Just pointing this out in case you see any super strange behaviour.

sql/src/SqlTables.sk
```diff
@@ -357,6 +357,176 @@ fun getLatestIndexDescr(
...
+      > = (row) -> {
+        key = row.i0;
+        for (datum in table.unsafeGetAllDataIter(key)) {
+          (_tick, source, writer, _value) = datum;
+          yield (key, (source, writer, Array[]))
+        }
+      };
+      table.writeArraySourceMany(
+        context,
+        table.unsafeGetFileIter().flatMap(clearData),
```
gregsexton, line 405: Why do we need to clear the data first? Can we not just build a new dir from the evalSelect and replace the current? I guess there's some hassle with setting up the downstream views?

gregsexton (commented):
Looks solid to me.

I'm curious about the eager approach to rebulding everything. Did you look much in to what it would take to do this incrementally? Could even be lazy about it and do a 'read repair' until a row is changed or purge kicks in and at this point write in the nulls.

I guess this gets pretty tricky for downstream views that need their schema changed.

sql/src/SqlTables.sk
bennostein: Yes, it's to do with the downstream dependencies.  I originally built a new dir and tried to rewire things, but it was pretty hairy and Julien suggested doing it this way instead.

sql/src/SqlTables.sk
bennostein: Oh, good to know - will keep in mind!  My thinking was that composing iterators would be more lightweight (both in complexity and in actual allocations etc.) but can revisit if we see weirdness

bennostein:
> I'm curious about the eager approach to rebulding everything. Did you look much in to what it would take to do this incrementally? Could even be lazy about it and do a 'read repair' until a row is changed or purge kicks in and at this point write in the nulls.

I looked into it a bit, leaving the old data in place and setting up an incremental computation from it to a new directory.  I decided to go with this eager approach when thinking about adding multiple columns over time, and wanting to avoid paying maintenance cost or growing the incremental computation graph based on that history -- instead, the DB knows only about the current schema and other operations all work as before.
 
I'm thinking that extract-column schema migrations will need to maintain reactive views in order to serve clients on the pre-extraction schema, though.  Still working out the details there. 

> I guess this gets pretty tricky for downstream views that need their schema changed.

And this too :)

sql/src/SqlTables.sk
```diff
@@ -357,6 +357,85 @@ fun getLatestIndexDescr(
   }
 }
 
+fun alterTableAddColumn(
```
pikatchu, line 423: Looks good to me!

pikatchu (approved):

PR #217, opened 2024-04-29
[skdb] Compress representation of IndexProjKey and IndexProjection

jberdine:
The `kinds` of `IndexProjKey` and `IndexProjection` includes direction
and type information that can be removed. The direction is always
`IASC`, and so the comparison function can be specialized and the
direction trimmed from the representation. The type information is
unused.


bennostein:
Looks good to me! But, should probably also rename the field from `kinds` to `indices` or `columns` if we're not storing any "kinds" info there anymore.

bennostein (approved):

PR #219, opened 2024-04-30
[compiler] Missed simplifications of StringSwitch

jberdine:
When all string representations except "LongString" were removed, some
simplifications to the code generated for StringSwitch instructions
were missed.


gregsexton:
LGTM.

PR #222, opened 2024-05-01
[sql] Extend privacy model with explicit group ownership

bennostein:
I was thinking through our group owner/admin model (looking at Greg's draft docs and some related conversations) and realized that the previous setup didn't really match up with my intuitions around application-level group ownership and administration.

Very open to discussion on this and aware that others might have different expectations or opinions on how these things should work!

## Background

Before this PR, an SKDB group specified a corresponding admin group and skdb_access; when a user created a group using the TS API, we set up three groups as follows

```
GROUP | ADMIN_GROUP | SKDB_ACCESS
---------------------------------
base  | admin       | read-write
admin | owner       | admin
owner | owner       | owner
```

That is, we have a globally-visible `base` group administered by an `admin` group, which is visible only to `admins` and itself administered by an `owner` group, which is administered by and visible only to itself.

This works, but doesn't line up perfectly with typical notions of admins/owners, namely in that:
* we can't enforce the property that only _owners_ can delete any of the three groups 
* admins are unable to add/remove other admins

## This PR

I think the simplest solution is to bake "ownership" into the model rather than encoding it with multiple "admin" relationships.

This PR does just that, adding an `owner` column to records in `skdb_groups` and enforcing that (1) owners have full admin privilege, regardless of their privileges in the admin group and (2) only owners can delete groups.

Now, when a user creates a group using the TS API, we set up the three groups as follows

```
GROUP | ADMIN_GROUP | OWNER_GROUP | SKDB_ACCESS
-----------------------------------------------
base  | admin       | owner       | read-write
admin | admin       | owner       | admin
owner | owner       | owner       | owner
```

Note that admins can now add/remove other admins.

### DIFF
--- a/sql/privacy/init.sql
+++ b/sql/privacy/init.sql
@@ -21,13 +21,15 @@ CREATE TABLE skdb_user_permissions(
 
 -- groupID: which group are we talking about?
 -- skdb_author: who is the creator of the group?
--- adminID: who should be able to be admin of the group?
--- skdb_access: who should be able to see/modify the administrators
+-- adminID: who should be able to change permissions for the group
+-- ownerID: who should be able to delete the group and override admins
+-- skdb_access: who should be able to see the group
 
 CREATE TABLE skdb_groups(
   groupID TEXT PRIMARY KEY,
   skdb_author TEXT,
   adminID TEXT NOT NULL,
+  ownerID TEXT NOT NULL,
   skdb_access TEXT NOT NULL
 );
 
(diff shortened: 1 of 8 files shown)


gregsexton (approved):
LGTM. I think this is a clearer model.

Landing this and rolling it out is going to be a PITA though, it's a pretty significant backward breaking change and involves the data itself rather than just protocol.

I think we'll just need to stop the world, migrate all the databases to the new model, and spin up a server with the new version. We'll also need to upgrade the client we use everywhere: console, intern, authd, ctrld, etc.

Can you bump the min required JS version in the server after this lands so that we reject old clients with an appropriate error?

Also can you update the www docs? I actually decided not to document the model and how it works because I knew you were going to make this change. Would be great if we could capture that too.

bennostein:
Ahh, yes, I hadn't thought through the deployment headache here.  I can release updated node and docker artifacts once this lands, and will also document the model in the www docs.

As for the min version in the server, I'm not sure exactly what you're referring to -- I know we have the kotlin client version there (currently kt-0.0.3) but don't see any logic requiring a min JS version.  Where should I be looking?

gregsexton:
Here: https://github.com/SkipLabs/cloud/blob/05ef10d3eeb6185c1acae00fe5c118dcaca7bee7/skgwd/src/main/kotlin/App.kt#L569

> I can release updated node and docker artifacts once this lands

Would you be up for doing server releases together over VC? I'm quite conscious that I'm the only person that knows how to do this stuff.

bennostein:
Thanks, and yes of course!  I'll ping you on discord

PR #224, opened 2024-05-02
Change representation of strings so that all are NUL-terminated

jberdine:
Currently some but not all strings are NUL-terminated, and the code
that calls POSIX/C functions needs to check for the terminator and
reallocate the string if it is missing. This PR changes the
representation so that all strings are NUL-terminated and the
reallocation is unnecessary and removed.

This change has the effect that the `String.byteSize` function returns
the logical "length" of the string in bytes, excluding the
NUL-terminator. So uses of this size that are calculating the number
of bytes the memory representation occupies need to be adjusted
by 1. These changes are, as far as I know, isolated to a few places in
the runtime.


compiler/src/AsmOutput.sk
```diff
@@ -2012,16 +2012,11 @@ mutable class .AsmOutput{
...
+  // Static strings always take a multiple of 8 bytes and have a trailing '\0'
   private readonly fun paddedStringSize(s: UTF8String): Int {
     size = s.utf8.size();
 
-    // WARNING! This must equal CSTR_BUFFER_SIZE in the runtime!
-    cstr_buffer_size = (80 - 8 + 1);
-
-    // Guarantee zero-padding for long enough strings.
-    if (size >= cstr_buffer_size && s.utf8[size - 1] != 0) {
+    if (size == 0 || s.utf8.last() != 0) {
```
mbouaziz: Why don't we just remove the condition so it's easy to know how much padding is added?

prelude/runtime/string.c
```diff
@@ -37,15 +37,20 @@ SkipInt SKIP_String_hash(char* obj) {
...
 
-// Allocates a string (with no data nor hash).
+// Allocates and nul-terminates a string (with no data nor hash).
 char* sk_string_alloc(uint32_t size) {
   sk_string_t* obj =
-      (sk_string_t*)SKIP_Obstack_alloc(size + sk_string_header_size);
+      (sk_string_t*)SKIP_Obstack_alloc(sk_string_header_size + size + 1);
   obj->size = size;
-  return (char*)&(obj->data);
+  char* str = (char*)&(obj->data);
```
mbouaziz: I think the cast is unneeded as you changed `sk_string_t` `data` field from `unsigned char` to `char` in a previous PR

prelude/runtime/string.c
```diff
@@ -37,15 +37,20 @@ SkipInt SKIP_String_hash(char* obj) {
...
+  str[size] = '\0';
+  return str;
 }
 
 char* sk_string_create(const char* buffer, uint32_t size) {
+  if (size > 0 && buffer[size - 1] == '\0') {
+    size--;
```
mbouaziz: This looks ugly...

And it is actually ugly because it will change what's stored in the `size` field of the resulting `sk_string_t`. Is it intended?

prelude/runtime/free.c
```diff
@@ -39,7 +39,7 @@ void sk_free_obj(sk_stack_t* st, char* obj) {
 
   // Check if we are dealing with a string
   if (SKIP_is_string(obj)) {
-    size_t memsize = get_sk_string(obj)->size;
+    size_t memsize = get_sk_string(obj)->size + 1;
```
mbouaziz, line 42: This isn't always true, is it?

prelude/runtime/string.c
jberdine: The intent here is to avoid adding a nul terminator if there already is one, which will lead to the resulting string being considered to have an embedded nul character, and blow up if passed to C. It isn't pretty. I audited all the call sites, and it looks like the only way for that to happen is if creating a string from an array of uint8 where the last element is 0 or to create a string as the contents of a file where the last byte of the file is 0. Keeping the extra nul in those cases seems legit as they are part of the explicit input data. So I've updated to remove this check.

prelude/runtime/string.c
jberdine: There is still the distinction between pointers and arrays: `error: incompatible pointer types initializing 'char *' with an expression of type 'char (*)[0]'`

prelude/runtime/free.c
jberdine: Did you have something specific in mind? My thinking was that for static strings their size is rounded up to a multiple of 8, but those shouldn't be given to free anyhow, and all the other strings are allocated by `sk_string_alloc` in `string.c` which does allocate this number of bytes.

compiler/src/AsmOutput.sk
jberdine: Yes, removing is ok. This test is in practice always true anyhow, since no static strings that contain an embedded nul at the end are ever created.

prelude/runtime/posix.c
```diff
@@ -297,17 +290,14 @@ int64_t SKIP_posix_isatty(int64_t fd) {
   return (char)rv;
 }
 
-int64_t SKIP_posix_mkstemp(char* template_obj) {
-  char* template = sk2c_string(template_obj);
```
jberdine, line 301: It turns out that this function relies on `sk2c_string` to make a copy since `mkstemp` will overwrite its argument. This broke sktest in a way that made it not run any tests. Fixed in latest update.

prelude/runtime/string.c
mbouaziz: Use `&obj->data[0]`?

prelude/runtime/free.c
mbouaziz: Now that you unconditionally add the extra byte, that's good

mbouaziz (approved):

prelude/runtime/string.c
jberdine: Oh, yeah, nice!

PR #226, opened 2024-05-03
Fix memory issues

gregsexton:
Found and fixed a segfault on multiple errors in the repl.

Fairly confident that I fixed an obstack leak on error in the event loop.

### DIFF
--- a/prelude/src/skstore/Context.sk
+++ b/prelude/src/skstore/Context.sk
@@ -1527,7 +1527,9 @@ fun runWithGcIntern(
       None()
     }) match {
     | Some(x) -> x
-    | None() -> return context
+    | None() ->
+      destroyObstack(pos);
+      return context
     };
     newContextOp match {
     | CContinue(underLockF) ->
--- a/sql/src/SqlEval.sk
+++ b/sql/src/SqlEval.sk
@@ -1816,14 +1816,11 @@ private mutable class SQLContext(
     (newContext, underLockF) = try {
       f(this.context with {globals => this.globals})
     } catch {
-    | err @ SKJSON.ParseError _
-    | err @ SKJSON.FieldError _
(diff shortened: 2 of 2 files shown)


gregsexton:
Some 👀 on this one would be much appreciated.

prelude/src/skstore/Context.sk
```diff
@@ -1527,7 +1527,9 @@ fun runWithGcIntern(
       None()
     }) match {
     | Some(x) -> x
-    | None() -> return context
+    | None() ->
+      destroyObstack(pos);
+      return context
```
beauby, line 1532: Yes, I remember this bit being suspicious.

sql/src/SqlEval.sk
```diff
@@ -1816,14 +1816,11 @@ private mutable class SQLContext(
...
-    | err @ SKJSON.ParseError _
-    | err @ SKJSON.FieldError _
-    | err @ SKJSON.SuperTypeError _
-    | err @ SKJSON.ReconstructError _
-    | err @ Error _ ->
-      (SKStore.cloneContextWithError(this.context, err), None())
-    | exn -> throw exn
+    | ex ->
+      cexn = SKStore.destroyObstackWithValue(pos, List[ex]).getHead();
+      throw cexn
```
beauby, line 1821: Do we want to always throw here? The code previously did not, and I suspect it might interfere with how errors are being bubbled up (although the previous clunky mechanism for bubbling errors up relied on setting a variable on the context since `destroyObstackWithValue()` was not available then).

gregsexton: It's caught and turned in to a Failure object slightly higher in the stack. I believe this mirrors the previous interface. I'm just using the exn as a way of getting an error out in a memory safe way - I assume it's safe, seems to be an established pattern.

PR #228, opened 2024-05-14
[skargo] Description of profiles

mbouaziz:
Adds a section describing profiles to `skargo build --help`:

```
 VALUES:
    PROFILE-NAME                                                            
        dev         Development profile: no optimizations, debug symbols    
        release     Release profile: with optimizations
```

`PROFILE-NAME` refers to the value name in the `OPTIONS` section:
```
        --profile <PROFILE-NAME>    Use the specified profile
```

...preceded by a few older commits that were aiming at GADTifying `Arg`s but failed because Skip doesn't have locally existential types (yet!). Though they make `Arg` fields a bit more consistent.


beauby (commented):
I am a bit uneasy about having other packages depending on skargo internals (I know it's already the case, but we should probably not entangle it more).

beauby:
LGTM! My only concern is the dependency on skargo from other packages.

mbouaziz:
I assume that you are talking about using `kProfileArgs` in `sknpm`.

It's not really internal, it's a user-facing thing. The `--release` and `--profile` args of `sknpm` are just passed as-is to `skargo`. This happens to be text-based because `sknpm` launches `skargo` via the system shell. But otherwise it would call `skargo` as a library and the profile arg would just be a parameter exposed by the library.

As for the existing dependencies on `Skargo` from `Sknpm`, I'll leave it to @skiplabsdaniel

beauby:
> I assume that you are talking about using `kProfileArgs` in `sknpm`.

Correct.
> It's not really internal, it's a user-facing thing. The `--release` and `--profile` args of `sknpm` are just passed as-is to `skargo`.

The flags are indeed part of the CLI interface, but `kProfileArgs` per se is not.
> This happens to be text-based because `sknpm` launches `skargo` via the system shell. But otherwise it would call `skargo` as a library and the profile arg would just be a parameter exposed by the library.

This is precisely the issue: skargo does not have a stable API at the moment.

That being said, given the pre-existing situation, feel free to merge as is.

mbouaziz:
@jberdine Any idea why `make fmt` is complaining in the CI whereas everything's fine locally on my machine?

jberdine:
> @jberdine Any idea why `make fmt` is complaining in the CI whereas everything's fine locally on my machine?

It is complaining about the files:
```
 M cli/tests/tests.sk
 M sknpm/src/commands/publish.sk
 M sktest/extra/src/wasm32.sk
 M sql/src/Main.sk
```
I'll pull down your branch and have a look.

jberdine:
For me locally `make fmt` also changes those files. What is your setup? If you explicitly invoke `skfmt -i cli/tests/tests.sk sknpm/src/commands/publish.sk sktest/extra/src/wasm32.sk sql/src/Main.sk` do you get changes?

PR #230, opened 2024-05-15
Fix usage of parallel in Makefile and check_format.sh

mbouaziz:
`parallel command :::` expects arguments to follow `:::` (as used correctly in `test_date.sh`)

But it seems that older versions of `parallel` (such as 20210822) behaves as if `:::` is not here if no arguments follow, where as newer versions (such as 20231122) do not and just do nothing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,11 +100,11 @@ clean:
 
 .PHONY: fmt-sk
 fmt-sk:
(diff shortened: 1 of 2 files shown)


jberdine (approved):
LGTM. Would be good to get sql/test_sql.sh and sql/test_date.sh while you are at it.

mbouaziz:
`sql/test_sql.sh` and `sql/test_date.sh` are using `:::` correctly I think

PR #231, opened 2024-05-16
[std] Support `cwd` in `Posix.Popen`.

beauby:
This commit enables setting the child process' cwd when calling `Posix.Popen::create{}`/`System.popen{}`.

### DIFF
--- a/prelude/src/stdlib/other/Posix.sk
+++ b/prelude/src/stdlib/other/Posix.sk
@@ -244,6 +244,7 @@ mutable class Popen(
     args: Array<String>,
     env: Map<String, String> = Map[],
     clear_env: Bool = false,
+    cwd: ?String = None(),
     stdin: Bool = false,
     stdout: Bool = false,
     stderr: Bool = false,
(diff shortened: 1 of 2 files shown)


gregsexton:
LGTM

prelude/src/stdlib/other/Posix.sk
```diff
@@ -277,6 +278,13 @@ mutable class Popen(
     };
     new_env = env.items().map(e -> `${e.i0}=${e.i1}`).collect(Array);
 
+    // NOTE: Since `posix_spawn()` does not provide a way for changing the
+    // working directory of the child process, we rely on GNU env.
+    cwd match {
+    | Some(path) -> !args = Array["env", "-C", path].concat(args)
+    | None() -> void
+    };
+
```
jberdine, line 287: We can't use `posix_spawn_file_actions_addchdir` I guess? This makes me think it would be better to give up on posix_spawn and just use (v)fork/exec so that we can `chdir` and whatever else is needed in the child between `fork` and `exec`. Wrapping the whole thing in `env` is just a heavy-weight version of that right?

beauby: Yeah at this stage `posix_spawn_file_actions_addchdir_np` is a non-standard extension. I agree that `posix_spawn` is a bit too limiting for our use, and that we should move to `(v)fork`/`exec`. I'll open an issue.

PR #234, opened 2024-05-16
[skargo] Handle `--color` flag.

beauby:


skargo/src/console.sk
```diff
@@ -18,12 +18,22 @@ base class ConsoleVerbosity uses Orderable {
...
   const kHeaderColor: TermColor.Color = TermColor.Green();
   const kErrorColor: TermColor.Color = TermColor.Red();
   const kWarningColor: TermColor.Color = TermColor.Yellow();
   const kNoteColor: TermColor.Color = TermColor.Cyan();
 
+  static fun create(verbosity: ConsoleVerbosity, color: String): Console {
+    use_colors = color match {
+    | "always" -> true
+    | "never" -> false
+    | "auto" | _ -> IO.stderr().isatty()
```
jberdine, line 31: This seems too eager to me. What about https://pubs.opengroup.org/onlinepubs/7908799/xcurses/has_colors.html ? At the minimum I'd say that hardcoding a check that TERM is not dumb or emacs would be needed.

beauby: I agree that the check is very basic for now but keep in mind the previous default behavior was `always`. I am not too keen on adding a dependency on curses just for this, but I'm in favor of adding a `SKARGO_TERM_COLOR` env var to provide a stop-gap for the time being. What do you think?

jberdine: As a stop-gap, I reckon checking that TERM is not dumb or emacs would be just as easy to implement and better for users. Opinions @gregsexton ?

gregsexton: Ah, don't worry. Colors are not a problem, Emacs handles them just fine. There's some logic to strip out the escape codes and use text properties to color the text.

What emcas shell mode does not support is arbitrarily repositioning the cursor and redrawing text. It does support repositioning on the current line, so things like progress bars, spinners, etc work. But not generally. So you can't run things like htop or ncurses stuff - stuff that wants to manage the whole display.

[sorry I can't help myself] And so why do I use this?? I think a lot of people see no htop as a deal breaker, but for me it's a huge feature. I want as many things as possible to behave like a log and to have a uniform (emacs of course) interface. This way I can go back and inspect things, interact with the buffer like I do any other, etc etc. E.g. I often prefer pidstat to htop. How do you search for something in htop again? No problem with an emacs buffer.

PR #235, opened 2024-05-17
Fix: propagate updates when in a transaction

gregsexton:
I noticed that if tailing a reactive view built on a table I:

- do an insert in to the source table, the tail outputs the update instantly

- do the same insert wrapped in a transaction, it does not update
  until the condvar times out

Digging in, this seems to be because the context isn't update()d and
so the changed dir names do not include the view, so notify never
triggers the condvar.

### DIFF
--- a/sql/src/SqlEval.sk
+++ b/sql/src/SqlEval.sk
@@ -1485,6 +1485,7 @@ class Evaluator{options: Options, user: ?UserFile} {
       error(pos, "OR REPLACE not supported in transactions")
     } else if (inTransaction) {
       this.insertTable(context, table, rows);
+      context.update();
       None()
     } else if (onConflict is Some(P.OCUpdate(_))) {
       onConflict match {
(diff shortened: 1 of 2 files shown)


gregsexton:
I figure this is the right fix rather than updating at the end. It makes sense that within a txn you would be able to read the effect of a write on downstream views after it has happened.

bennostein (approved):
Looks good to me -- I agree about this being the right semantics, it would be a sharp edge for views and tables to get out of sync within a transaction like that.

edit: local sequence number test failure looks legitimate, though -- I guess that logic relies on the context not updating within a transaction.

pikatchu:
I agree with the fix. Maybe it might be worth adding a mechanism to make sure the update was called whenever the context was modified?

If we tripped over it here, there might be other places like that.

gregsexton:
> might be worth adding a mechanism to make sure the update was called whenever the context was modified?

Already on my todo list. This isn't the first time we've been bitten. Are there any situations you can think of where we don't automatically want to update though?

gregsexton:
> local sequence number test failure looks legitimate, though -- I guess that logic relies on the context not updating within a transaction.

Yeah was relying on the gen sym counter. Now that we call update it will be bumped. I just updated the test expectation so tests should pass now.

PR #236, opened 2024-05-17
Postgres data connector proof of concept

gregsexton:
Syncs postgres tables with SKDB tables. Not yet productionised.


gregsexton:
I forgot this was still open. Merging in. It stands alone.

PR #237, opened 2024-05-20
[skc] Consistently use absolute paths in `Makefile`.

beauby:
Using relative paths in `$PATH` causes issues when changing the cwd.


mbouaziz (approved):
Seems legit

PR #238, opened 2024-05-21
[skfmt] Minor arg improvements

mbouaziz:
I am plugging skfmt in a vscode extension. When formatting-on-save, I get both the filename and the buffer contents. I have to use `skfmt` with stdin because the new contents is not written yet.
- Though `skfmt` failed are reporting syntax errors on stdin. Here is a fix.
- I also want those errors to be reported with the correct filename, hence a new option to do it.

### DIFF
--- a/compiler/src/skfmt.sk
+++ b/compiler/src/skfmt.sk
@@ -42,6 +42,7 @@ untracked fun main(): void {
   filenames = args.getArray("files");
   for (file in filenames) {
     contents = if (file == "-") {
+      invariant(!inplace, "Cannot use both - and -i");
       !file = "<stdin>";
       buf = mutable Vector[];
       IO.stdin().read_to_end(buf) match {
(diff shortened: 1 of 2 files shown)


compiler/src/skfmt.sk
```diff
@@ -35,11 +35,17 @@ untracked fun main(): void {
         .short("i")
         .about("Inplace edit file(s), if specified"),
     )
+    .arg(
+      Cli.Arg::string("filename")
```
jberdine: FWIW clang-format calls this option `--assume-filename`

jberdine (approved):
LGTM.

It would also be good to add use of this option to `bin/git_hooks/check_format.sh`.

compiler/src/skfmt.sk
mbouaziz: :+1: 
Renamed!

PR #244, opened 2024-05-30
[skargo] Cleaner skc invokations.

beauby:
This PR makes `skc` invokations cleaner by passing relative paths
to sources files.


gregsexton:
sheep it

beauby:
Yup, I’m just figuring out why CI is unhappy first

PR #245, opened 2024-05-30
[runtime] Give static linkage to obstack globals.

beauby:
This prevents name clashes (notably for the `end` symbol which is
defined such that its address points near the end of the `.bss`
segment, cf. [man](https://man7.org/linux/man-pages/man3/end.3.html)).

### DIFF
--- a/prelude/runtime/obstack.c
+++ b/prelude/runtime/obstack.c
@@ -33,30 +33,25 @@
      -------------
 */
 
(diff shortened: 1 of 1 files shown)


jberdine (approved):
LGTM. For future reference, it would be nice if you added the link to the man page for `end` that you sent in chat. That is nasty and anything to avoid having to hunt this down again later would be good.

PR #246, opened 2024-06-20
Docs for SKStore

pikatchu:
First attempt to describe SKStore. Still needs some work.

### DIFF
--- a/skiplang/docs/SKStore.md
+++ b/skiplang/docs/SKStore.md
@@ -1,0 +1,146 @@
+# SKStore: A generic framework for reactive programming
+
+SKStore is a store made of collections called `directories`.
+Each directory contains entries associating values to a key.
+Keys can be of any type as long as they extend the class `SKStore.Key`; values can be of any type as long as they extend the class `SKStore.File`.
+The APIs available for `SKStore` can be found in `Context.sk` and `Handle.sk`.
+
+## DirName
+
+Each directory (or collection) has a name, the name is defined with a slash separated path "/of/this/form/".
+The way to create a `DirName` is by using the primitive `DirName::create`.
+A directory name must start and finish with a slash; if it doesn't, an exception is thrown.
+The names are not used directly when manipulating directories, however, they come in handy when trying to debug the state that the system is in.
+One can navigate all the reactive values using familiar primitives (such as cd, ls etc ...).
+[comment]: <> (Can they? I've never seen that in the code and wasn't aware of any interactive shell-like thing)
+
+## Input directories
+
+Input directories are directories that can be modified by the outside world.
+To create one, you must use the primitive `mkdir` and specify a type for both keys and values
+
+```
+myInputDirectory =
+  context.mkdir(Type1::keyType, Type2::type, "/myInput/Name/")
+```
+
+Note that the fields "keyType" and "type" are inherited when one extends `SKStore.Key` and `SKStore.File` respectively.
+
+Once a directory is created, one can modify it (add/remove):
+
+```
+myInputDirectory.writeArray(context, myKey, Array[myValue])
+```
+
+Removing an element consists of writing an empty array (also called a "tomb").
+There is no semantic difference between an empty array and no entry, since SKStore returns an empty array when trying to access a missing entry.
+[comment]: <> (One way wonder why you may associate arrays only, not any type of values and arrays would be a special case (There is no sematic difference between an empty array and no entry gives a hint))
+[comment]: <> (It's not obvious that writeArray is replacing and not appending (Removing an element just consists in writing an empty array gives a hint as well))
+
+Tombs are useful to handle concurrency and to synchronize directories with the outside world.
+They are collected automatically by the system.
+
+## Eager directory
+
+An eager directory is a directory defined as the result of a "map" over one or multiple directories.
+For example:
+
+```
+sumDir = myDirOfIntegers.map(
+  IID::keyType,
+  IntFile::type,
+  context,
+  "/result/",
+  (context, writer, key, fileIterator) ~> {
+    sum = 0;
+    for(intFile in fileIterator) {
+      sum = sum + intFile.value
+    };
+    writer.set(key, IntFile(sum));
+  }
+)
+```
+
+In this example, we map over a directory of type `IntFile` and produce the sum for each entry.
+Once this map is defined, a link has been created between `myDirOfIntegers` and `sumDir`.
+Every time `myDirsOfIntegers` changes, the directory `sumDir` will be updated accordingly.
+
+
+### Direct access
+
+Sometimes, mapping over all values in a directory is not what we need; there are cases where a specific value we are interested in is determined at runtime.
+SKStore gracefully handles this case, by allowing one to directly look up values in any available directory.
+For example, let's imagine a directory that is made of names which need to be looked up in another directory.
+We would program it by using the primitive `getArray`, which can be used on any SKStore directory:
+
+```
+sumDir = myDirOfNames.map(
+  SID::keyType,
+  StringFile::type,
+  context,
+  "/result2/",
+  (context, writer, key, fileIterator) ~> {
+    values = mutable Vector[];
+    for(file in fileIterator) {
+      value = myOtherDirectory.getArray(context, SID::create(file.value));
+      values.push(value);
+    };
+    result = // .. do something with values ...
+    writer.set(key, result);
+  }
+)
+```
+[comment]: <> (It is worth noting that this automatically records the dependency, and noting the granularity of the dependency record.)
+
+### Subdirectories
+
+It is possible to map over a directory inside of a mapper function.
+The resulting directory can then be stored in a file, and made accessible to the rest of the system, because directories are also values.
+
+You can also create input directories within a mapper, but this is a very advanced feature that we don't recommend using for beginners.
+The problem is that this kind of directory will be "cleaned up" if the key that created the directory is removed.
+When that is the case, all the data that was in that input directory gets wiped out, which can lead to data loss.
+To keep things safe, we recommend you only create input directories at toplevel.
+
+## Lazy directories
+
+A lazy directory works just like a cache, except that the cache invalidation is handled automatically by SKStore.
+To define one, use the primitive `LHandle::create`:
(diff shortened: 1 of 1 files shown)


docs/SKStore.md
```diff
@@ -0,0 +1,171 @@
...
+when one extends `SKStore.Key` and `SKStore.File` respectively.
+
+Once a directory is created, one can modify it (add/remove):
+
+```
+myInputDirectory.writeArray(context, myKey, Array[myValue])
+```
+
+Removing an element just consists in writing an empty array (also
+called a "tomb"). There is no sematic difference between an empty
```
mbouaziz: Two things here:
- one way wonder why you may associate arrays only, not any type of values and arrays would be a special case (_There is no sematic difference between an empty array and no entry_ gives a hint)
- it's not obvious that `writeArray` is replacing and not appending (_Removing an element just consists in writing an empty array_ gives a hint as well)

docs/SKStore.md
```diff
@@ -0,0 +1,171 @@
...
+```
+sumDir = myDirOfNames.map(
+  SID::keyType,
+  StringFile::type,
+  context,
+  "/result2/",
+  (context, writer, key, fileIterator) ~> {
+    values = mutable Vector[];
+    for(file in fileIterator) {
+      value = myOtherDirectory.getArray(SID::create(file.value));
```
mbouaziz:
```suggestion
      value = myOtherDirectory.getArray(context, SID::create(file.value));
```
It is worth noting that this automatically records the dependency

gregsexton: And the granularity of the dep record.

jberdine:
I think we should have this in the repo, so I moved the file under skiplang, rebased, and made some small edits.

jberdine:
@pikatchu feel free to object to the hijack!

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+The keys can be of any type as long as they extend the class `SKStore.Key`, the values can be of any type as long as they extend the class `SKStore.File`.
+The APIs available for `SKStore` can be found in `Context.sk` and `Handle.sk`.
+
+## DirName
+
+Each directory (or collection) has a name, the name is defined with a slash separated path "/of/this/form/".
+The way to create a `DirName` is by using the primitive `DirName::create`.
+A directory name must start and finish with a slash, if it doesn't, an exception is thrown.
+The names are not used directly when manipulating directories, however, they come in handy when trying to debug the state that the system is in.
+One can navigate all the reactive values using familiar primitives (such as cd, ls etc ...).
```
bennostein, line 14: Can they? I've never seen that in the code and wasn't aware of any interactive shell-like thing

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+SKStore is a store made of collections called `directories`.
+Each directory contains entries associating values to a key.
+The keys can be of any type as long as they extend the class `SKStore.Key`, the values can be of any type as long as they extend the class `SKStore.File`.
+The APIs available for `SKStore` can be found in `Context.sk` and `Handle.sk`.
+
+## DirName
+
+Each directory (or collection) has a name, the name is defined with a slash separated path "/of/this/form/".
+The way to create a `DirName` is by using the primitive `DirName::create`.
+A directory name must start and finish with a slash, if it doesn't, an exception is thrown.
```
bennostein:
```suggestion
A directory name must start and finish with a slash; if it doesn't, an exception is thrown.
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
+# SKStore: A generic framework for reactive programming
+
+SKStore is a store made of collections called `directories`.
+Each directory contains entries associating values to a key.
+The keys can be of any type as long as they extend the class `SKStore.Key`, the values can be of any type as long as they extend the class `SKStore.File`.
```
bennostein:
```suggestion
Keys can be of any type as long as they extend the class `SKStore.Key`; values can be of any type as long as they extend the class `SKStore.File`.
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+Each directory (or collection) has a name, the name is defined with a slash separated path "/of/this/form/".
+The way to create a `DirName` is by using the primitive `DirName::create`.
+A directory name must start and finish with a slash, if it doesn't, an exception is thrown.
+The names are not used directly when manipulating directories, however, they come in handy when trying to debug the state that the system is in.
+One can navigate all the reactive values using familiar primitives (such as cd, ls etc ...).
+
+## Input directories
+
+Input directories are directories that can be modified by the outside world.
+To create one, you must use the primitive `mkdir`, the type of the key and of the values must be specified:
```
bennostein:
```suggestion
To create one, you must use the primitive `mkdir` and specify a type for both keys and values
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+
+```
+myInputDirectory =
+  context.mkdir(Type1::keyType, Type2::type, "/myInput/Name/")
+```
+
+Note that the fields "keyType" and "type" are automatically defined when one extends `SKStore.Key` and `SKStore.File` respectively.
```
bennostein:
```suggestion
Note that the fields "keyType" and "type" are inherited when one extends `SKStore.Key` and `SKStore.File` respectively.
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+
+Once a directory is created, one can modify it (add/remove):
+
+```
+myInputDirectory.writeArray(context, myKey, Array[myValue])
+```
+
+Removing an element just consists of writing an empty array (also called a "tomb").
```
bennostein:
```suggestion
Removing an element consists of writing an empty array (also called a "tomb").
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+There is no semantic difference between an empty array and no entry, since SKStore returns an empty array when trying to access a missing entry.
+
+Tombs are useful to handle concurrency and to synchronize directories with the outside world.
+They are collected automatically by the system.
```
bennostein, line 41: Is this relevant here? Feels more of an implementation detail to me

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+In this example, we map over a directory of type `IntFile` and produce the sum for each entry.
+Once this map is defined, a link has been created between `myDirOfIntegers` and `sumDir`.
+Every time `myDirsOfIntegers` changes, the directory `sumDir` will be updated accordingly.
+
+
+### Direct access
+
+Sometimes, mapping over directories is not what we need, there are cases where the value we are interested in is determined at runtime.
```
bennostein:
```suggestion
Sometimes, mapping over all values in a directory is not what we need; there are cases where a specific value we are interested in is determined at runtime.
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+SKStore gracefully handles this case, by allowing one to directly lookup values in any available directory.
```
bennostein:
```suggestion
SKStore gracefully handles this case, by allowing one to directly look up values in any available directory.
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+For example, let's imagine a directory that is made of names, and those names need to be looked up in another directory.
```
bennostein:
```suggestion
For example, let's imagine a directory that is made of names which need to be looked up in another directory.
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+  }
+)
+```
+
+### Subdirectories
+
+It is possible to map over a directory inside of a mapper function.
+The resulting directory can then be stored in a file, and made accessible to the rest of the system, because directories are also values.
+
+You can also create inputDirectories within a mapper, but this is a very advanced feature that we don't recommend using for beginners.
```
bennostein:
```suggestion
You can also create input directories within a mapper, but this is a very advanced feature that we don't recommend using for beginners.
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+  "/myDirName/",
+  (context, self, key) ~> {
+    Array[/* some value */]
+  }
+)
+```
+
+Just like in any other cache, the values are automatically evicted if they are not used often enough.
+The cache eviction policy is programmable (cf `Handle.sk`).
+Note that Lazy directories are very handy to handle recursive definitions because they pass around a reference to themselves (the parameter `self` of the closure above).
```
bennostein:
```suggestion
Note that lazy directories are very handy to implement recursive definitions because they pass around a reference to themselves (the parameter `self` of the lambda above).
```

skiplang/docs/SKStore.md
```diff
@@ -0,0 +1,142 @@
...
+# Guarantees
+
+SKStore is a simple yet powerful API to define reactive systems.
+It provides the following guarantees:
+
+- The system is transactional.
+  Either the transaction succeeded and the whole system updated, or the transaction failed and nothing changed.
+There is no such thing as "half" updates.
+
+- The system is concurrent, multiple readers and multiple writers can interact with SKStore at the same time, without blocking each-other (except at commit time).
```
bennostein:
```suggestion
- The system is concurrent: multiple readers and multiple writers can interact with SKStore at the same time without blocking each-other (except at commit time).
```

bennostein (approved):
Bunch of grammar/prose tweaks, a few comments, otherwise LGTM

skiplang/docs/SKStore.md
jberdine: At this point, this whole file is documenting the implementation, so seems ok.

skiplang/docs/SKStore.md
jberdine: I don't know, but will add a comment with your question.

PR #247, opened 2024-06-21
[sql] Add support for column-extraction schema migrations, and replication with legacy schema clients

bennostein:
Finally got this working to a point I'm happy with in the TS integration tests -- start there (`apitests.ts` / `testSchemaChanges`) to get a sense of what this feature adds.  I'm going to also try hammering on this with soak/replication tests, so may uncover as-of-yet unknown issues, but putting this up now in case people have cycles/want to take a look

The core thing we're adding is the ability to "extract" a column from a table to a one-to-many relationship with another table.

For example, we can go from a table with schema
`CREATE TABLE users (id INTEGER PRIMARY KEY, username TEXT, email TEXT)` 
to two tables
`CREATE TABLE users (id INTEGER PRIMARY KEY, username TEXT)` and `CREATE TABLE user_emails (user_id INTEGER, email TEXT, skdb_original INTEGER)`.

Crucially, we want to be able to serve clients connecting on _either_ the old or the new schema.  To do so, we set up what I'm calling "legacy views" here, in which we maintain a view joining together the new `users` and `user_emails` tables, from which we can serve a client who doesn't know about the split.

On the read path, this doesn't involve much work for the server: just tail that view, but call it by the name that the client knows about.
On the write path, it's a bit more complicated: we take legacy-schema rows and split out the writes to the corresponding tables, then let them bubble up to the legacy view from there.  This requires some messing with the sources/writers as the writes bubble up, to make sure we tail all the new rows in the legacy view but don't reflect writes back to writers.

There are some pretty wide-ranging changes to support all that.  I tried to break things up into a commit history where each logical change can be viewed separately, but it's quite possible I've mixed that up or included some change that isn't strictly necessary here.

EDIT: see [notion](https://www.notion.so/Schema-Migration-35496239d18e49a792440e0eadbb8780) for a design doc with much greater detail.


jberdine:
I haven't looked closely yet, but first one high-level question that I expect to come up. When taking a write to the legacy view and splitting it up into multiple writes to the new tables, are there atomicity concerns to make sure that what used to be a single write of a single row ends up like a transaction? And if there are transactions involving the legacy views, that gets more involved, or not? I'm just wondering if you have some design comments around this that could help reading the code.

bennostein:
> are there atomicity concerns to make sure that what used to be a single write of a single row ends up like a transaction? And if there are transactions involving the legacy views, that gets more involved, or not?

Yes, but all of the normal atomicity/transactional consistency stuff still works under these changes,  e.g. if a write violates a check constraint, is rejected for privacy reasons, etc. then it is rejected.  A single data message in the replication protocol consists of a sequence of table diffs; previously those were applied one-to-one (logic with the flavor of `syntacticTableDiffs.map(semantics).each(apply)`), now we first interpret each syntactic table diff as potentially multiple table diffs to apply (something like `syntacticTableDiffs.flatMap(distribute-to-extracted-column-tables).map(semantics).each(apply)`).   But all of that is wrapped in the same type of transaction as before.

Notably, when a column is extracted I generate a view checking the constraint that there is at most one "original" value for each primary key in the new table.  So if that gets violated, the transaction will be rejected.

gregsexton:
Looks good to me. As in I don't spot anything concerning. But replication is very subtle. I'll run the simulation tests now with this checked out. But would be very good to add a schema migration to these tests and have two clients replicate with different schemas.

gregsexton:
Simulation tests are broken actually. They need to be updated for the breaking change in the tail spec. I'm seeing the error "Tailer must specify an expected schema for each table".

You can run them with `make test-replication`. These and the soak tests (`make test-soak`) are worth running to be sure replication is solid.

bennostein:
> Simulation tests are broken actually. They need to be updated for the breaking change in the tail spec. I'm seeing the error "Tailer must specify an expected schema for each table".
> 
> You can run them with `make test-replication`. These and the soak tests (`make test-soak`) are worth running to be sure replication is solid.

Those breakages are actually unrelated and predate this diff.  Fixed in #274

gregsexton:
Tests all run ok then? You should ship this if they are good. I think the simulation tests are more interesting than soak for this diff.

PR #248, opened 2024-07-01
SKStore for JS

skiplabsdaniel:
First version of the JS SKStore API.

### DIFF
--- a/prelude/ts/src/sk_types.ts
+++ b/prelude/ts/src/sk_types.ts
@@ -1,6 +1,11 @@
 export type float = number;
 export type int = number;
 export type ptr = number;
+export type ErrorObject = {
+  message: string;
+  stack?: string[];
+  cause?: ErrorObject;
+};
 
 export type Opt<T> = T | null;
 
@@ -35,7 +40,7 @@ class SkRuntimeExit extends Error {
   }
 }
 
-class SkException extends Error {
+class SkError extends Error {
   constructor(msg: string) {
     super(msg);
   }
@@ -184,6 +189,7 @@ interface Exported {
   SKIP_initializeSkip: () => void;
   SKIP_skfs_end_of_init: () => void;
   SKIP_callWithException: (fnc: ptr, exc: int) => ptr;
+  SKIP_getExceptionId: (exn: ptr) => ptr;
   SKIP_getExceptionMessage: (skExc: ptr) => ptr;
   SKIP_get_persistent_size: () => int;
   SKIP_get_version: () => number;
@@ -213,7 +219,8 @@ export class Utils {
   private stderr: Array<string>;
   private stddebug: Array<string>;
   private mainFn?: string;
-  private exceptions: Error[];
+  private exception?: Error;
+  private stacks: Map<ptr, string>;
 
   exit = (code: int) => {
     let message =
@@ -222,7 +229,7 @@ export class Utils {
   };
 
   constructor(exports: WebAssembly.Exports, env: Environment, mainFn?: string) {
-    this.exceptions = [];
+    this.stacks = new Map();
     this.states = new Map();
     this.args = new Array();
     this.current_stdin = 0;
@@ -256,24 +263,13 @@ export class Utils {
     new_stdin: string = "",
   ) => {
     this.args = [this.mainFn ?? "main"].concat(new_args);
-    this.exceptions = [];
+    this.exception = undefined;
+    this.stacks = new Map();
     this.current_stdin = 0;
     this.stdin = utf8Encode(new_stdin);
     this.stdout = new Array();
     this.stderr = new Array();
     this.stddebug = new Array();
-  };
-
-  exnCause = () => {
-    let cause: Error | undefined = undefined;
-    while (this.exceptions.length > 0) {
-      let exc = this.exceptions.pop();
-      if (cause) {
-        (exc as any).cause = cause;
-      }
-      cause = exc;
-    }
-    return cause;
   };
 
   runCheckError = <T>(fn: () => T) => {
@@ -284,7 +280,7 @@ export class Utils {
     }
     if (this.stderr.length > 0) {
       let error = new Error(this.stderr.join(""));
-      (error as any).cause = this.exnCause();
+      (error as any).cause = this.exception;
       throw error;
     }
     return res;
@@ -304,7 +300,8 @@ export class Utils {
       if (exn instanceof SkRuntimeExit) {
         exitCode = exn.code;
       } else {
-        (exn as any).cause = this.exnCause();
+        if (this.exception && this.exception != exn)
+          (exn as any).cause = this.exception;
         throw exn;
       }
     }
@@ -344,7 +341,7 @@ export class Utils {
       }
       message = lines.join("\n");
       let error = new SkRuntimeExit(exitCode, message?.trim());
-      (error as any).cause = this.exnCause();
+      (error as any).cause = this.exception;
       throw error;
     }
     return this.stdout.join("");
@@ -482,21 +479,28 @@ export class Utils {
     this.exports.SKIP_skfs_end_of_init();
   };
   etry = (f: ptr, exn_handler: ptr) => {
+    let err: Error | null = null;
     try {
       return this.call(f);
-    } catch (exn) {
-      this.exceptions.push(exn as Error);
-      let exception =
-        exn instanceof SkException
-          ? null
-          : new Exception(exn as Error, this.state);
+    } catch (exn: any) {
+      if (this.exception && this.exception != exn) exn.cause = this.exception;
+      err = exn as Error;
+      this.exception = err;
+      let exception: Exception | null = null;
+      if (!(err instanceof SkError)) {
+        exception = new Exception(err, this.state);
+      }
       return this.callWithException(exn_handler, exception);
+    } finally {
+      if (this.exception == err) {
+        this.exception = undefined;
+      }
     }
   };
   ethrow = (skExc: ptr, rethrow: boolean) => {
     if (this.env && this.env.onException) this.env.onException();
-    if (rethrow && this.exceptions.length > 0) {
-      throw this.exceptions.pop();
+    if (rethrow && this.exception) {
+      throw this.exception;
     } else {
       let skMessage =
         skExc != null && skExc != 0
@@ -517,9 +521,18 @@ export class Utils {
           message = "SKFS Internal error";
         }
       }
-      throw new SkException(message);
+      const err = new SkError(message);
+      if (err.stack) this.stacks.set(skExc, err.stack);
+      throw err;
     }
   };
+  replace_exn(oldex: ptr, newex: ptr) {
+    const stack = this.stacks.get(oldex);
+    if (stack) {
+      this.stacks.delete(oldex);
+      this.stacks.set(newex, stack);
+    }
+  }
   deleteException = (exc: int) => {
     this.state.exceptions.delete(exc);
   };
@@ -529,6 +542,64 @@ export class Utils {
       return this.state.exceptions.get(exc)!.err.message;
     } else {
       return "Unknown";
+    }
+  };
+
+  getExceptionStack = (exc: int) => {
+    if (this.state.exceptions.has(exc)) {
+      return this.state.exceptions.get(exc)!.err.stack ?? "";
+    } else {
+      return "";
+    }
+  };
+
+  getErrorObject: (skExc: ptr) => ErrorObject = (skExc: ptr) => {
+    let skMessage =
+      skExc != null && skExc != 0
+        ? this.exports.SKIP_getExceptionMessage(skExc)
+        : null;
+    let message =
+      skMessage != null && skMessage != 0
+        ? this.importString(skMessage)
+        : "SKFS Internal error";
+    let errStack = this.stacks.get(skExc);
+    let lines = message.split("\n");
+    if (lines[0].startsWith("external:")) {
+      let external = lines.shift()!;
+      message = lines.join("\n");
+      let id = parseInt(external.substring(9));
+      if (this.state.exceptions.has(id)) {
+        const exception = this.state.exceptions.get(id);
+        if (exception) {
+          message = exception.err.message;
+          errStack = exception.err.stack;
+        } else {
+          message = "Unknown error";
+        }
+      } else if (message.trim() == "") {
+        message = "SKFS Internal error";
+      }
+    }
+    const toObject: (error: Error) => ErrorObject = (error?: Error) => {
+      const errcause = (error as any).cause;
+      const errstack = (error as any).stack?.split("\n") as string[];
+      if (errstack) {
+        return errcause
+          ? { message, cause: errcause, stack: errstack }
+          : { message, stack: errstack };
+      } else {
+        return errcause ? { message, cause: errcause } : { message };
+      }
+    };
+    if (errStack) {
+      const stack: string[] = errStack.split("\n") as string[];
+      return this.exception
+        ? { message, cause: toObject(this.exception), stack }
+        : { message, stack };
+    } else {
+      return this.exception
+        ? { message, cause: toObject(this.exception) }
+        : { message };
     }
   };
 
--- a/prelude/ts/src/sk_types.ts
+++ b/prelude/ts/src/sk_types.ts
@@ -163,6 +163,7 @@ export interface Environment {
   encodeUTF8: (str: string) => Uint8Array;
   onException: () => void;
   base64Decode: (base64: string) => Uint8Array;
+  base64Encode: (toEncode: string, url?: boolean) => string;
   fs: () => FileSystem;
   sys: () => System;
   crypto: () => Crypto;
@@ -205,6 +206,13 @@ export interface WasmSupplier {
 function utf8Encode(str: string): Uint8Array {
   return new TextEncoder().encode(str);
 }
+
+export type Main = (new_args: Array<string>, new_stdin: string) => string;
+
+export type App = {
+  environment: Environment;
+  main: Main;
+};
 
 export class Utils {
   private exports: Exported;
@@ -638,14 +646,28 @@ export class Utils {
   };
 
   runWithGc = <T>(fn: () => T) => {
+    this.stddebug = new Array();
     let obsPos = this.exports.SKIP_new_Obstack();
     try {
       let res = fn();
       this.exports.SKIP_destroy_Obstack(obsPos);
-      return res;
+      if (
+        res !== null &&
+        typeof res === "object" &&
+        (("__isArrayProxy" in res && res.__isArrayProxy) ||
+          ("__isObjectProxy" in res && res.__isObjectProxy)) &&
+        "clone" in res
+      ) {
+        const clone = res.clone as () => T;
+        return clone();
+      } else return res;
     } catch (ex) {
       this.exports.SKIP_destroy_Obstack(obsPos);
       throw ex;
+    } finally {
+      if (this.stddebug.length > 0) {
+        console.log(this.stddebug.join(""));
+      }
     }
   };
 
--- a/prelude/ts/src/sk_types.ts
+++ b/prelude/ts/src/sk_types.ts
@@ -562,19 +562,16 @@ export class Utils {
   };
 
   getErrorObject: (skExc: ptr) => ErrorObject = (skExc: ptr) => {
-    let skMessage =
-      skExc != null && skExc != 0
-        ? this.exports.SKIP_getExceptionMessage(skExc)
-        : null;
-    let message =
-      skMessage != null && skMessage != 0
-        ? this.importString(skMessage)
-        : "SKFS Internal error";
+    if (skExc == null || skExc == 0) {
+      return { message: "SKStore Internal error" };
+    }
+    let message = this.importString(
+      this.exports.SKIP_getExceptionMessage(skExc),
+    );
     let errStack = this.stacks.get(skExc);
     let lines = message.split("\n");
     if (lines[0].startsWith("external:")) {
       let external = lines.shift()!;
-      message = lines.join("\n");
       let id = parseInt(external.substring(9));
       if (this.state.exceptions.has(id)) {
         const exception = this.state.exceptions.get(id);
@@ -585,7 +582,7 @@ export class Utils {
           message = "Unknown error";
         }
       } else if (message.trim() == "") {
-        message = "SKFS Internal error";
+        message = "SKStore Internal error";
       }
     }
     const toObject: (error: Error) => ErrorObject = (error?: Error) => {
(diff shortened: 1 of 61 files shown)


prelude/ts/src/sk_types.ts
```diff
@@ -532,6 +545,64 @@ export class Utils {
...
+        : null;
+    let message =
+      skMessage != null && skMessage != 0
+        ? this.importString(skMessage)
+        : "SKFS Internal error";
+    let errStack = this.stacks.get(skExc);
+    let lines = message.split("\n");
+    if (lines[0].startsWith("external:")) {
+      let external = lines.shift()!;
+      message = lines.join("\n");
```
bennostein: We still have `message` here, no need to reconstruct it from `lines` right?

prelude/ts/src/sk_types.ts
```diff
@@ -532,6 +545,64 @@ export class Utils {
...
+    if (this.state.exceptions.has(exc)) {
+      return this.state.exceptions.get(exc)!.err.stack ?? "";
+    } else {
+      return "";
+    }
+  };
+
+  getErrorObject: (skExc: ptr) => ErrorObject = (skExc: ptr) => {
+    let skMessage =
+      skExc != null && skExc != 0
```
bennostein: This seems like a good place for an early return on the false case of this `?` (i.e. `if (skExc == null || skExc ==0) {return ...}`)

Following the logic through it all seems to make sense, but it's a bit tricky with all of the nesting and combinations of `if` statements and `?` expressions

prelude/ts/src/sk_types.ts
```diff
@@ -532,6 +553,64 @@ export class Utils {
...
+        ? this.exports.SKIP_getExceptionMessage(skExc)
+        : null;
+    let message =
+      skMessage != null && skMessage != 0
+        ? this.importString(skMessage)
+        : "SKFS Internal error";
```
bennostein: Here (and line 588) we should probably say "SKStore" not "SKFS"

skstore/src/Extern.sk
```diff
@@ -0,0 +1,668 @@
...
+@export("SKIP_SKStore_asyncLazy")
+fun asyncLazy(
+  context: mutable Context,
+  name: String,
+  getFunPtr: UInt32,
+  lazyFunPtr: UInt32,
+): String {
+  getHandle = FnHandle::make(getFunPtr);
+  lazyHandle = FnHandle::make(lazyFunPtr);
+  remove = f ~> f != "params" && f != "ksuid";
```
bennostein: Could this have a more descriptive name than "remove"? e.g. `fieldFilter` or similar

skstore/src/Skdb.sk
```diff
@@ -0,0 +1,182 @@
+module alias P = SQLParser;
+
+module SKDBStore;
```
bennostein: I don't understand this choice of module name -- it contains utilities for mapping to/from JSON tables, could we perhaps name it something related to that?  Or if I'm misunderstanding, would you add some comment documenting here?

skstore/src/Skdb.sk
```diff
@@ -0,0 +1,182 @@
...
+      rowKey = SKDB.RowKey(rowValues, dirDescr.kinds);
+      writer.set(rowKey, rowValues)
+    },
+  );
+  context.maybeGetEagerDir(converted.getDirName()).each(dir ->
+    dir.copyTo(context, dirDescr.dirName)
+  )
+}
+
+fun decodeParams(input: SKJSON.CJObject): Map<String, P.Value> {
```
bennostein, line 88: What's this for?  Seems very similar to the `decodeParams` in SqlEval.sk, but I don't see any callers here.

skstore/ts/tests/tests.ts
```diff
@@ -0,0 +1,509 @@
...
+      ],
+    ],
+    new Sum(),
+  );
+  eager3.mapTo(output, (key, it) => {
+    return [key, it.first()];
+  });
+}
+
+async function testMultiMapReduceRun(
```
pikatchu, line 321: Why is this one async?

skiplabsdaniel: It's to provide a global mechanism for all tests.
For async tests, It's used to wait for the async calls before doing the checks.
It's possible to split the tests in two categories.

skstore/src/Utils.sk
```diff
@@ -0,0 +1,230 @@
...
+  path = context.currentArrow() match {
+  | Some(arrow) ->
+    b64 = base64(arrow.key);
+    `${arrow.childName}${b64}/${name}/`
+  | None() -> `/sk/${name}/`
+  };
+  DirName::create(path)
+}
+
+fun parentName(context: mutable Context): DirName {
```
bennostein: What is this used for? It also seems to fix a convention of `/sk/` as the "global" parent (along with the line 207 case above) -- what's the story there?

bennostein (commented):
Still working through the changes (especially the typescript api/impl) but thought I'd post a first batch of comments now

skstore/src/Extern.sk
```diff
@@ -0,0 +1,668 @@
...
+const kSessionDir: DirName = DirName::create("/sk_prv/sessions/");
+const kGraphDir: DirName = DirName::create("/sk_prv/graph/");
+
+const kJSONTypes: Types<JSONID, JSONFile> = Types(
+  JSONID::keyType,
+  JSONFile::type,
+);
+
+class JSONID(value: SKJSON.CJSON) extends Key {
+  //
```
bennostein: Dead comment?

skstore/src/Extern.sk
```diff
@@ -0,0 +1,668 @@
...
+    values: mutable NonEmptyIterator<JSONFile>,
+  ): SKJSON.CJSON {
+    applyConvertToRowFun(this.getId(), key.value, values)
+  }
+}
+
+class ExternAccumulator(
+  accumulator: FnHandle,
+) extends Accumulator<JSONFile, JSONFile> {
+  //
```
bennostein, line 184: and another

skstore/src/Extern.sk
```diff
@@ -0,0 +1,668 @@
...
+@export("SKIP_SKStore_asyncResult")
+fun asyncResult(
+  callId: String,
+  lazyName: String,
+  key: SKJSON.CJSON,
+  params: SKJSON.CJSON,
+  value: SKJSON.CJObject,
+): Float {
+  SKStore.runLockedWithResult(context ~> {
+    resultsDirs = DirName::create(lazyName + "sk_prv/results/");
```
bennostein, line 324: For my understanding: what's the "prv" in "sk_prv"?

skstore/src/Extern.sk
```diff
@@ -0,0 +1,668 @@
...
+      if (current.isEmpty()) {
+        throw AsyncValueSaveError(key, params, "Query info not found")
+      };
+      cvalue = current[0].value match {
+      | x @ SKJSON.CJObject _ -> x
+      | _ -> throw AsyncValueSaveError(key, params, "Invalid query info")
+      };
+      ksuid = SKJSON.getString(cvalue, "ksuid");
+      if (!ksuid.map(v -> callId == v).default(false)) {
+        // The received response is deprecated
```
bennostein, line 341: Can you elaborate on this? What is deprecated? If we return void here, we won't hit the context update on line 418, and presumably will "succeed" at line 420, which seems like it may mask some actual issues.

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+ * error - The error message
+ */
+export type Failure = {
+  status: "failure";
+  error: string;
+};
+
+/**
+ * Type return by an async function call if a data is the same (for exemple '304 Not Modified' code for http)
+ * payload - The valuable data a the type
```
bennostein: Is this documentation wrong or should there be an additional `payload` field on the `Unchanged` type def? (and type parameter `V`)

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+  previous?: { payload: V; metadata?: M };
+};
+
+/**
+ * Type return by an async lazy reactive handle an error occurs.
+ * error - The error message
+ * previous - The value of the previous call if exist.
+ */
+export type Error<V extends TJSON, M extends TJSON> = {
+  status: "failure";
```
bennostein, line 100: this status field can't be used to distinguish from the previous `Failure` type -- could we give it a different code (and possibly more descriptive name) to avoid confusion?

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+   */
+  asyncLazy<K extends TJSON, V extends TJSON, P extends TJSON, M extends TJSON>(
+    get: (key: K) => P,
+    call: (key: K, params: P) => Promise<AValue<V, M>>,
+  ): LHandle<K, Loadable<V, M>>;
+
+  log(object: TJSON): void;
+}
+
+export interface Handles {
```
bennostein: Why is this is exported twice? (also on line 97, with slight difference in `apply`'s signature)

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+  V1 extends TJSON,
+  K2 extends TJSON,
+  V2 extends TJSON,
+> = (key: K1, it: NonEmptyIterator<V1>) => Iterable<[K2, V2]>;
+
+/**
+ * The handle entry mapper function to write data into table
+ * @param key - the mapped handle entry key
+ * @param {NonEmptyIterator} it - an iterator on values avalable for a key
+ * @returns {R} the table entry corresponding to eager map key value pair
```
bennostein, line 150: Why does this return just one `R` and not an `Iterable<R>`?  It seems likely/common that you would want to generate a row for each value in `it`

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+    mapper: Mapper<K, V, K2, V2>,
+  ): EHandle<K2, V2>;
+  /**
+   * Map over each eager reative map entry and apply mapper function
+   *  then reduce the when the given accumulator
+   * @param {Mapper} mapper - function to apply to the eager map entry
+   * @param {Accumulator} accumulator - reduction manager
+   * @returns {EHandle} The the resulting eager reactive map handle
+   */
+  mapReduce<K2 extends TJSON, V2 extends TJSON, V3 extends TJSON>(
```
bennostein, line 248: Why require a map + reduce together? Wouldn't the interface be more flexible if we defined map/reduce separately? i.e. you could write `x.map(foo).reduce(bar)` to replace any `x.mapReduce(foo,bar)`, but as far as I can see, the only way to write `x.reduce(foo)` using just `mapReduce` is with an identity-function mapper that creates an unnecessary intermediate directory.

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+   * Insert an entry in the table
+   * @param entries - the entries to insert in the table
+   * @param update - indicate if the value need to be updated in case of index conflict
+   * @throws {Error} when not update and an index contraints is broken
+   */
+  insert(entry: R[], update?: boolean): void;
+  /**
+   * Update an entry in the table
+   * @param row - the table entry to update
+   * @param updates - the column values updates
```
bennostein, line 311: Can you give some more detail on the expected structure of this argument?  Looking through the implementation, it looks like it's expected to be an object whose keys are column names in the table, and whose values are constants to overwrite those columns with -- strings/numbers/booleans (passed straight through) or an object (which we stringify for storage).  Is that right?

If so, that's strictly less expressive than a SQL update statement, since we're restricting values to be constants instead of arbitrary expressions.  Doesn't seem like a huge issue (since one can always fall back to the actual SQL on these tables) but want to be sure I'm understanding correctly.

bennostein: And similarly for other `JSONObject` types in this interface -- in `select`, `deleteWhere`, `watch`, `watchChanges`, users will definitely need to know what structure of object to pass in (or expect in their callbacks).

Ideally, we'd be able to specify these in a much less general type than `JSONObject` -- is there a reason that's not possible? (that keys match up with column names we of course won't be able to enforce statically. But the structure at least would be good!)

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+    update: (added: JSONObject[], removed: JSONObject[]) => void,
+  ) => { close: () => void };
+}
+
+export type Mapping<
+  K1 extends TJSON,
+  V1 extends TJSON,
+  K2 extends TJSON,
+  V2 extends TJSON,
+> = [EHandle<K1, V1>, Mapper<K1, V1, K2, V2>];
```
bennostein: Why a length-2 array instead of an object with 2 named fields?  Seems less clear and more error-prone to me this way (unless there's some benefit I'm not seeing)

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+
+export type Lazy<K extends TJSON, V extends TJSON> = (
```
bennostein: I'm confused as to what exactly this represents and why we need to export a name for it (instead: inline a function-typed argument at the one place it is used, line 366).

If it needs to be named and exported explicitly could we give it a more specific/descriptive name than "Lazy"?

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+    mappings: Mapping<K1, V1, K2, V2>[],
+  ): EHandle<K2, V2>;
+  /**
+   * Map over each entry of each eager reative map and apply the corresponding mapper function
+   *  then reduce the when the given accumulator
+   * @param {Mapping} mappings - the handles to combine
+   * @param {Accumulator} accumulator - reduction manager
+   * @returns {EHandle} The the resulting eager reactive map handle
+   */
+  multimapReduce<
```
bennostein, line 405: Same comment as for `mapReduce` above: can we just define `reduce` directly and let users compose with multimap if/when they need?

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+export interface SKStoreFactory extends Shared {
+  runSKStore(
+    init: (skstore: SKStore, ...tables: TableHandle<TJSON[]>[]) => void,
+    app: App,
+    tables: MirrorSchema[],
+    connect?: boolean,
+  ): Promise<Table<TJSON[]>[]>;
+}
+
+export interface SKStore {
```
bennostein, line 375: I'm also a bit confused about the module/interface structure here.  Why are these specific functions (`lazy`, `multimap`, `multimapReduce`, `asyncLazy`, and `log`) members of this SKStore interface and everything else at the top-level?

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+  SKIP_SKStore_multimapReduce(
+    ctx: ptr,
+    name: ptr,
+    mappings: ptr,
+    accumulator: int,
+    accInit: ptr,
+  ): ptr;
+}
+
+export interface SKJSON extends Shared {
```
bennostein, line 483: What's going on with this interface / how does it relate to the code in `skstore_skjson.ts`?  As far as I can see this is defined but never actually implemented.

skstore/ts/src/skstore_api.ts
```diff
@@ -0,0 +1,588 @@
...
+    mappings: Mapping<K1, V1, K2, V2>[],
+    accumulator: Accumulator<V2, V3>,
+  ): string;
+
+  noref: () => Context;
+  notify?: () => void;
+}
+
+/**
+ * Collect function file line colum
```
bennostein: This docstring doesn't say much! (and I don't understand the comment on line 580, either)

- why is this one function in this file being implemented?  does it really belong here in the "`skstore_api`"?

- what assumptions are being made with the regex (and magic constant 2 added to the offset) that might break if some detail changes with the underlying stack representation?

bennostein (commented):
Another batch of comments/questions.

I've got a few commits worth of doc improvements/changes that I'll push soon -- mostly I stuck to things that I understand well, but if anything there seems off to you, happy to chat or change things!

bennostein:
Those changes now stacked at https://github.com/skiplabsdaniel/skdb/pull/1 -- decided not to push directly to your fork to keep reviewing simple

skstore/src/Skdb.sk
skiplabsdaniel: It's the part of SKStore that relates to SKDB so I called it SKDBStore. This contains utilities for going from/to SQL tables to/from SKStore JSON directories. If the name is confusing, I'll think of a better one. I plan to add comment, I wanted to wait until the stabilization for it.

skstore/src/Skdb.sk
skiplabsdaniel: Indeed, it's the version that uses SKJSON, and the code that call it no more in that version. We probably need to use only SKJSON in SKDB to prevent this kind of thing.

skstore/src/Utils.sk
skiplabsdaniel: Useless function from older version. Indeed, as we mixing SKDB tables directories and SKStore directories and as the directory name are fully generated, I choose to have a convention of '/sk/' as the "global" parent for the pure SKStore side, just a habit for mixing skstore computation graphs.

skstore/src/Extern.sk
skiplabsdaniel: A bad habit to have one line between class definition and first function signature.
Easier to read for me

skstore/src/Extern.sk
skiplabsdaniel: Same

skstore/src/Extern.sk
jberdine: doesn't seem bad to me

skstore/src/Extern.sk
skiplabsdaniel: Just to specify its a private internals directory.

skstore/src/Extern.sk
skiplabsdaniel: The response is deprecated because new lazy computation was fired  for the concerned key so the current response is useless because it correspond to an old state.

skstore/ts/src/skstore_api.ts
skiplabsdaniel: Indeed it's false, bad copy/paste

skstore/ts/src/skstore_api.ts
skiplabsdaniel: The two types are not used in the same context. I use to have a different status, but finally I kept the same name to not have to create a new object. But if it confusing, i'll change it.

skstore/ts/src/skstore_api.ts
skiplabsdaniel: No reason, squash/rebase effect.

skstore/ts/src/skstore_api.ts
skiplabsdaniel: Internally it's a mapMeduce, the reducer is part of the map for efficiency memory issues I guess, we can separate them at user level, but we lose a bit of memory.

skstore/ts/src/skstore_api.ts
skiplabsdaniel: It's just a first quick version of something that not use SQL, because I had the remark during the demo I have done.
That functions are here only to have something running quickly, I did not expect that function to be really used, (except watch and  watchChanges) because the changes shall be performed by the mirroring.
But you are right, if they are here to stay, they need better typing and documentation.

skstore/ts/src/skstore_api.ts
skiplabsdaniel: Indeed, it's not part of skstore api, it's just a utility function to get function metadata.
Like for insert/update/delete just here to have a first version, I should have made this clear in the comment, I need to spend some time to have a better version.

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -0,0 +1,429 @@
...
+  },
+  getOwnPropertyDescriptor(hdl: WasmHandle, prop: string) {
+    const fields = hdl.objectFields();
+    const idx = fields.get(prop);
+    if (idx === undefined) return undefined;
+    const value = getValueAt(hdl, idx, false);
+    return {
+      configurable: true,
+      enumerable: true,
+      writable: false,
```
jberdine, line 164: AFAIU `writable: false` has the effect that attempted writes have no effect. I wonder if there ought to be a `set` function that throws so that the programming error isn't silent. It seems that if there is a `set` function, it gets called even with `writable: false`.

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -0,0 +1,429 @@
...
+            res.push(getValueAt(hdl, i, true));
+          }
+          return res;
+        };
+      if (prop === "forEach")
+        return <T>(
+          callbackfn: (value: T, index: number, array: T[]) => void,
+          thisArg?: any,
+        ) => {
+          self.toJSON().forEach(callbackfn, thisArg);
```
jberdine, line 194: Just for my understanding, this could instead loop through the elements like `toJSON` does internally, and pass the restult of `getValueAt` to `callbackfn` directly without first calling `toJSON`, right? I don't have intuition whether such allocation avoiding optimizations are worthless or not in this context?

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -0,0 +1,429 @@
...
+      return undefined;
+    }
+  },
+  has(hdl: WasmHandle, prop: string): boolean {
+    if (prop === "__isArrayProxy") return true;
+    if (prop === "__pointer") return true;
+    if (prop === "length") return true;
+    if (prop === "clone") return true;
+    if (prop === "toJSON") return true;
+    return false;
```
jberdine, line 207: This does not report the indices converted to strings as keys, while `get` and `ownKeys` handle that case. Is this difference intended?

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -0,0 +1,429 @@
...
+  if (typeof value === "object") {
+    if (Array.isArray(value)) {
+      return value.map(clone) as T;
+    } else if (aValue.__isArrayProxy) {
+      const res: any[] = [];
+      const length: number = aValue.length;
+      for (let i = 0; i < length; i++) {
+        res.push(clone(aValue[i]));
+      }
+      return res as T;
```
jberdine, line 238: Would it make sense to define enough functionality on the array proxy objects so that `value.map(clone)` would work in this case too?

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -0,0 +1,429 @@
...
+    } else if (aValue.__isObjectProxy) {
+      const res: any = {};
+      for (const key of aValue.keys) {
+        res[key] = clone(aValue[key]);
+      }
+      return res as T;
```
jberdine, line 244: Maybe I don't understand proxy objects well enough, but I expected that since there is the `ownKeys` function, the code below for non-proxy objects would work also for proxy objects. Am I confused?

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -0,0 +1,429 @@
...
+}
+
+export interface SKJSON extends Shared {
+  importJSON: (value: ptr, copy?: boolean) => any;
+  exportJSON: <T>(v: T) => ptr;
+  importOptJSON: (value: Opt<ptr>, copy?: boolean) => any;
+  importString: (v: ptr) => string;
+  exportString: (v: string) => ptr;
+  runWithGC: <T>(fn: () => T) => T;
+}
```
jberdine, line 303: This also appears in skstore_types.ts, is the duplication intentional?

jberdine (commented):
Some questions and minor comments.

PR #249, opened 2024-07-01
Show possible fields on invalid "."

pikatchu:
If you write `a.b` and `b` is a field that does not exist. You will have a suggestion like "did you mean ...". This is useful, but that suggestion is not always correct. This diff keeps that suggestion, but also list all the fields that are accessible, directly in the error message.


beauby (approved):
LGTM.

One tangential remark: do we really need all the `String.upper`/`String.uc`/`String.uppercase` aliases?

pikatchu:
> LGTM.
> 
> One tangential remark: do we really need all the `String.upper`/`String.uc`/`String.uppercase` aliases?

Probably not.

PR #250, opened 2024-07-02
Definition when argument number mismatch

pikatchu:
During a call, if the number of arguments does not match the number of parameters in the definition, we were not displaying anything other than "invalid number of arguments".

This PR adds the position where the definition was found.


beauby (approved):
Nice.

PR #251, opened 2024-07-04
Add filterMap and findMap functions to Iterator

jberdine:


mbouaziz (approved):
LGTM. Though why not using `for in / return` in `findMap`

jberdine:
> why not using `for in / return` in `findMap`

I just followed the style of `find` and didn't think of it, but I guess it would be equivalent. I can change it, I suppose most people find the `for in / return` style easier to read than the explicit loop.

prelude/src/core/language/Iterator.sk
```diff
@@ -205,6 +214,16 @@ mutable base class .Iterator<+T> {
...
 
+  // s.findMap(f) is Some(u) = f(t) for the first t in s where f is Some(_)
+  overridable mutable fun findMap<U>(f: T -> ?U): ?U {
+    for (t in this) {
+      u = f(t);
+      if (u.isSome()) break u
+    } else {
+      None()
+    }
+  }
```
jberdine, line 225: @mbouaziz How about this version?

mbouaziz: go

mbouaziz (approved):

prelude/src/core/language/Iterator.sk
```diff
@@ -117,6 +117,15 @@ mutable base class .Iterator<+T> {
     }
   }
 
+  overridable mutable fun filterMap<U>(f: T -> ?U): mutable Iterator<U> {
```
gregsexton, line 120: mapcat - or whatever we call it, flatmap? - is the more general/powerful way of achieving this. Maybe this is more efficient? I imagine it's more ergonomic to use? Just wanted to call this out in case you hadn't considered flatmap.

prelude/src/core/language/Iterator.sk
gregsexton: Could also map().dropWhile(isNone).next(), but I guess less efficient?

jberdine: Maybe the compiler would be able to resolve and inline things enough that the efficiency would be the same, it isn't impossible, but not trivial either. But my motivation was more about the convenience of the API exposed by Iterator than pure efficiency at this point. My impression is that Iterator isn't meant to have a minimal interface, but to prefer many operations for user convenience since Iterator is in a sense a common interchange type for collections. I would be ok with aiming for minimality, but this might be the one type where I think it makes the most sense to not be minimal.

prelude/src/core/language/Iterator.sk
jberdine: Yes, flatMap would work for this, though we would need Option::iterator(). Like findMap, maybe the compiler could optimize them to the same code, but it would involve flattening a nested loop into an unnested one, so maybe not easy.

PR #252, opened 2024-07-08
Add mapWhile function to Iterator

jberdine:


jberdine:
I'm not absolutely sure, but I don't think this is definable in terms of existing functions.

prelude/src/core/language/Iterator.sk
```diff
@@ -354,6 +365,23 @@ mutable class MapIterator<T, +U>(
...
+  }
+
+  mutable fun next(): ?U {
+    this.base.next() match {
+    | Some(x) -> this.f(x)
+    | None() -> None()
+    }
+  }
+}
+
```
mbouaziz: No need for this since you're using `yield`

jberdine: Doh! Leftover from previous iteration, removed in https://github.com/SkipLabs/skdb/pull/252/commits/a9b3bd1b7ad73126a4493b73c9df931629ec3584

mbouaziz (approved):

prelude/src/core/language/Iterator.sk
```diff
@@ -69,6 +69,17 @@ mutable base class .Iterator<+T> {
     mutable MapIterator(this, f)
   }
 
+  // i.mapWhile(f) is similar to i.map(f).takeWhile(isSome).map(Option.get)
+  // but terminates as soon as f returns None()
```
gregsexton, line 73: wouldn't the above also stop mapping as soon as f returns none? Is map not lazy on an iterator?

jberdine: Indeed. I simplified down to this and didn't pay attention that in the end it wasn't needed at all, thanks!

mbouaziz: It is lazy but if you `.collect(Array)` then `i.map(f).takeWhile(isSome).map(Option.get)` will still go through till the end of `i`

jberdine: I don't think I understand, can you expand on this? Do you mean that `i.map(f).takeWhile((x) -> x.isSome()).collect(Array)` will call `f` on elements of `i` after a `None()` is produced? The implementation of `takeWhile` is:
```
  overridable mutable fun takeWhile(p: T -> Bool): mutable Iterator<T> {
    for (v in this) {
      if (!p(v)) {
        break void
      } else {
        yield v
      }
    }
  }
```
While I haven't found docs on specifically what `break` and `yield` mean, my understanding of this is that the iterator produced by `takeWhile` will be shorter than its argument if an element that does not satisfy `p` is found.

mbouaziz: No, actually @gregsexton is right

gregsexton: From time to time it happens.

mbouaziz: I liked `mapWhile` because it doesn't use `Object.get` which may throw (though won't, but could, but won't, but could).

And as we found out, laziness is quite nasty: https://github.com/SkipLabs/skjs/issues/88

PR #254, opened 2024-07-17
Revert "Add mapWhile function to Iterator"

jberdine:
Reverts SkipLabs/skdb#252


jberdine:
As discussed in #252 this is not needed.

PR #269, opened 2024-07-25
Support generating shared libraries.

beauby:
This PR adds support for `--emit=cdylib` in `skc`, which generates a shared object from a skip library.
In order to trigger it from `skargo`, add the following to `Skargo.toml`:
```
[lib]
type = ["sklib", "cdylib"]
```
which causes `skargo build` to generate both `libfoo.sklib` and `libfoo.so`.

### DIFF
--- a/prelude/runtime/runtime64_specific.cpp
+++ b/prelude/runtime/runtime64_specific.cpp
@@ -270,18 +270,36 @@ void SKIP_unsetenv(char* name) {
 void SKIP_memory_init(int pargc, char** pargv);
 void sk_persist_consts();
 
-int main(int pargc, char** pargv) {
+void sk_init(int pargc, char** pargv) {
   sk_saved_obstack_t* saved;
   argc = pargc;
   argv = pargv;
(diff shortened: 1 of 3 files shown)


prelude/runtime/runtime64_specific.cpp
```diff
@@ -270,18 +270,36 @@ void SKIP_unsetenv(char* name) {
...
-  std::set_terminate(terminate);
   SKIP_memory_init(pargc, pargv);
   saved = SKIP_new_Obstack();
   SKIP_initializeSkip();
   sk_persist_consts();
   SKIP_destroy_Obstack(saved);
+}
+
+#ifdef SKIP_LIBRARY
+__attribute__((constructor)) static void lib_init() {
```
beauby, line 285: There are valid points for not making this a constructor here, and letting downstream choose when it decides to init the skip runtime. Merging as is for now, but this will probably be revisited soon (removing `__attribute__((constructor))`, removing `static`, and renaming the function to something like `SKIP_init()`).

PR #273, opened 2024-08-02
[skstore] Object Oriented API

skiplabsdaniel:


gregsexton:
Is this API instead of the previous or as well as?

bennostein:
Instead of -- this is what lets us check immutability etc. dynamically instead of using the skjs typechecker / analyzer.

Still some polishing of the dev experience to be done here, so I'm pushing some examples through and seeing how it feels with the OO API

PR #274, opened 2024-08-02
Simulation test fixes

bennostein:
These tests had gotten stale in two ways:

 * Client-specified schemas are now required when tailing
 * Rebuild semantics have been tweaked and resets removed from the protocol

 (these changes are orthogonal to the column-extraction PR #247, so sending as separate PR)


gregsexton (approved):
Thanks for fixing!

PR #276, opened 2024-08-09
Implement JS Iterable interface on NonEmptyIterators

bennostein:
This enables things like for-each loops and passing NonEmptyIterators to functions that expect an iterable input

### DIFF
--- a/skstore/ts/src/skstore_api.ts
+++ b/skstore/ts/src/skstore_api.ts
@@ -210,7 +210,7 @@ export interface Accumulator<T extends TJSON, V extends TJSON> {
 /**
  * A mutable iterator with at least one element
  */
-export interface NonEmptyIterator<T> {
+export interface NonEmptyIterator<T> extends Iterable<T> {
   /**
    * Return the next element of the iteration.
    *   `first` cannot be called after `next`
--- a/prelude/src/skstore/Handle.sk
+++ b/prelude/src/skstore/Handle.sk
@@ -65,8 +65,22 @@ mutable class NonEmptyIterator<+T>(
   mutable isPastFirstValue: Bool = false,
   mutable materializedIter: ?Array<T> = None(),
 ) extends FileIterator<T> {
+  // Materialize `iter` into an array to support multiple/repeat iterations
+  // This operation is deferred until necessary, to avoid materialization in
+  // cases where only the `first` element is accessed
+  private mutable fun materializeIter(): void {
+    this.materializedIter match {
+    | None() ->
+      array = this.iter.collect(Array);
+      this.!materializedIter = Some(array);
+      this.!iter = (array.iterator() : mutable Iterator<T>)
+    | Some(_) -> void
+    };
+  }
+
   mutable fun next(): ?T {
     if (this.isPastFirstValue) {
+      this.materializeIter();
       this.iter.next()
     } else {
       this.!isPastFirstValue = true;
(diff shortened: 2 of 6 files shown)


skstore/ts/src/skstore_api.ts
```diff
@@ -210,7 +210,7 @@ export interface Accumulator<T extends TJSON, V extends TJSON> {
 /**
  * A mutable iterator with at least one element
  */
-export interface NonEmptyIterator<T> {
+export interface NonEmptyIterator<T> extends Iterable<T> {
```
gregsexton, line 213: This doesn't make sense really. An Iterable gives you an Iterator. Should this implement a different interface or maybe be called something else? `NonEmptySequence`?

bennostein: It's a wrapper around a skip iterator under the hood, so does actually need to be treated as such, not as a (reusable) sequence.  We could instead automatically materialize an array or something for mappers, which I wouldn't necessarily be opposed to but might have more memory overhead in some cases.

And this interface is fairly important for modern JS syntax: for-each loops, spread syntax, destructuring, etc.

Is the objection that this itself is an iterator and shouldn't produce another iterator?  We could follow the iterator protocol and return `{value, done}` objects from `next` but that's a pain to use directly.

gregsexton: Can you iterate over it twice? e.g. if I use it in a for loop and then another?

bennostein: Nope, still a skip iterator under the hood.  Do you interpret "Iterable" as "Iterable (many times)"? I haven't seen that in docs anywhere

gregsexton: Yeah, that's the usual definition. So an Iterable can provide an iterator (use once) as many times as you like. That's why I said it's odd to have an iterator implement iterable.

bennostein: See the "iterable iterator" snippet [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol) -- I don't think that's the interpretation of iterable in JS-world.

Though those docs do say "when possible, it's better for iterable[Symbol.iterator]() to return different iterators that always start from the beginning".  We could do that by materializing the full set of values upfront (would simplify some things for sure) but Julien had memory concerns IIRC

gregsexton: Ha, was just linking that to you. Javascript 🤦

jberdine: Sorry if this is noise, but I'm confused. Don't the underlying skip iterators need some explicit copy to be done at some point in order to iterate them more than once, since in general they are stateful and iterating more than once needs more than one copy of the internal state? So maybe it "makes sense in JS" for an iterator to implement iterable, but wouldn't the `iterator()` function need to be something like `this.clone().iterator()`?

bennostein: > Don't the underlying skip iterators need some explicit copy to be done at some point in order to iterate them more than once, since in general they are stateful and iterating more than once needs more than one copy of the internal state? 

Not confused, that's right!

> wouldn't the iterator() function need to be something like this.clone().iterator()?

That's a good idea, doing that in the skip heap instead of on the js side... I'll implement that and see how it goes

bennostein: Just updated to do this -- the new version supports iterating over the same collection many times, materializing the contents to an array the first time any element other than the first is accessed.

PR #278, opened 2024-08-12
[sknpm] Do not indent output of failed subprocesses

jberdine:
If a subprocess produces an error, some tools parse the messages and
expect them to start at the left margin. To avoid friction with these,
do not indent the contents of stdout and stderr when reporting the
output of failed subprocesses.

### DIFF
--- a/sknpm/src/sknpm.sk
+++ b/sknpm/src/sknpm.sk
@@ -216,17 +216,14 @@ private fun exec(
   if (!p.success()) {
     if (console.verbosity < verbosity) {
       console.error(
-        "command exited with non-zero status\n\n" +
-          "Caused by:\n" +
-          "  process did not exit successfully: \`" +
-          args[0] +
-          `\` (exit status: ${p.exitcode()})\n` +
-          "  --- stdout\n" +
-          p.stdout.split("\n").map(l -> "  " + l).join("\n") +
-          "\n" +
-          "  --- stderr\n" +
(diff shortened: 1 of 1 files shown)


jberdine:
Hmm, but the output is still something like:
```
  process did not exit successfully: `tsc` (exit status: 2)
  --- stdout
target/wasm32-unknown-unknown/release/ts/prv/skstore_impl.ts(71,31): error TS2314: Generic type 'Param' requires 2 type argument(s).
```
Note that the file that the error message points to is a copy in the `target` directory, not the actual _source_ `skstore/ts/src/prv/skstore_impl.ts`. Maybe we can get tsc to report using a different file name. I don't know why we are making copies of the sources though.

skiplabsdaniel:
Very good point, there's no fundamental reason to make a copy, and looking at that it's better to avoid it. 

LGTM

jberdine:
I was looking into the copies a bit more and it is done by the `copyAndConvertImport` function. This isn't always just a copy, and I'm not confident that I fully understand the original need to do this. Also, it is not immediately obvious to me that the processing that function does will preserve line numbers, so just changing the names of the files in the reports may not be enough.

skiplabsdaniel:
I plan to make a version that gets rid of anything too specific (including copyAndConvertImport).

jberdine:
Closing in favor of the plan to remove copyAndConvertImport.

PR #280, opened 2024-08-14
[skstore] Allow mappers to close only over handles

jberdine:
This diff removes the class type argument from map functions in the
SKStore API, and changes the parameters for the Mapper class
constructors from effectively
```
...params: any[]
```
to
```
eparams: EHandle<any, any>[],
lparams: LHandle<any, any>[]
```

The benefit of these changes is that calls to the `map` functions no
longer need to specify the type of the class. This helps since type
inference is not strong enough to elaborate the type parameters
automatically. Also, specifying the types of each parameter as only
either `EHandle` or `LHandle` gives API users help to only close over
state that the SKStore incremental computation engine is aware of.

A drawback is that splitting the params into a pair of arrays rather
than one varargs list is less convenient at call sites. Additionally,
the `constructor` functions of mapper classes have a uniform
signature that is only very weakly typed: the number and order of the
actual class constructor parameters must be ensured manually.

### DIFF
--- a/skstore/ts/examples/sum.ts
+++ b/skstore/ts/examples/sum.ts
@@ -34,7 +34,11 @@ class T2SIdentify<K extends TJSON, V extends TJSON>
 }
 
 class Add implements Mapper<number, number, number, number> {
-  constructor(private other: EHandle<number, number>) {}
+  private other: EHandle<number, number>;
+
+  constructor(eparams: EHandle<any, any>[], lparams: LHandle<any, any>[]) {
+    this.other = eparams[0];
+  }
 
   mapElement(
     key: number,
--- a/skstore/ts/examples/sum.ts
+++ b/skstore/ts/examples/sum.ts
@@ -5,6 +5,7 @@ import type {
   TJSON,
   Mapper,
   EHandle,
+  LHandle,
   NonEmptyIterator,
   OutputMapper,
 } from "skstore";
@@ -62,13 +63,10 @@ export function initSKStore(
   input2: TableHandle<[number, number]>,
   output: TableHandle<[number, number]>,
 ) {
-  const eager1 = input1.map<number, number, typeof T2SIdentify>(T2SIdentify);
-  const eager2 = input2.map<number, number, typeof T2SIdentify>(T2SIdentify);
-  const eager3 = eager1.map<number, number, typeof Add>(Add, eager2);
-  eager3.mapTo<[number, number], typeof ToOutput<number, number>>(
-    output,
-    ToOutput,
-  );
+  const eager1 = input1.map(T2SIdentify<number, number>, [], []);
+  const eager2 = input2.map(T2SIdentify<number, number>, [], []);
+  const eager3 = eager1.map(Add, [eager2], []);
+  eager3.mapTo(output, ToOutput, [], []);
 }
 
 export function scenarios() {
--- a/skstore/ts/src/skstore_api.ts
+++ b/skstore/ts/src/skstore_api.ts
@@ -14,7 +14,6 @@ export type TJSON = number | JSONObject | boolean | TJSON[] | string;
 
 export type TTableHandle = any;
 export type TTable = any;
-export type Param = any;
 
 export type DBType = "TEXT" | "JSON" | "INTEGER" | "FLOAT" | "SCHEMA";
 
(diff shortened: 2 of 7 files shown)


jberdine:
This is draft PR for discussion, I don't know if we should go this way, but maybe worth looking at and thinking about.

jberdine:
FWIW, so far my attempts to rewrite this using the isomorphism between `EHandle<any, any>[], LHandle<any, any>[]` and `(EHandle<any, any> | LHandle<any, any>)[]` have failed. I have been able to get examples to type-check, but haven't been able to get the type-checker to reject bad code. More investigation needed.

skstore/ts/examples/sum.ts
```diff
@@ -33,7 +34,11 @@ class T2SIdentify<K extends TJSON, V extends TJSON>
 }
 
 class Add implements Mapper<number, number, number, number> {
-  constructor(private other: EHandle<number, number>) {}
+  private other: EHandle<number, number>;
+
+  constructor(eparams: EHandle<any, any>[], lparams: LHandle<any, any>[]) {
+    this.other = eparams[0];
+  }
```
jberdine, line 41: Here is an example of how this approach gives very weak typing: the constructors must uniformly accept an array of `EHandles` and an array of `LHandles` (with no knowledge of their key and value types) and initialize the class fields by picking out the appropriate array elements. This has to be gotten right at map call sites manually. This isn't great at all.

skstore/ts/examples/sum.ts
```diff
@@ -62,13 +67,10 @@ export function initSKStore(
...
-  const eager2 = input2.map<number, number, typeof T2SIdentify>(T2SIdentify);
-  const eager3 = eager1.map<number, number, typeof Add>(Add, eager2);
-  eager3.mapTo<[number, number], typeof ToOutput<number, number>>(
-    output,
-    ToOutput,
-  );
+  const eager1 = input1.map(T2SIdentify<number, number>, [], []);
+  const eager2 = input2.map(T2SIdentify<number, number>, [], []);
+  const eager3 = eager1.map(Add, [eager2], []);
+  eager3.mapTo(output, ToOutput, [], []);
```
jberdine, line 73: But note how much less type instantiation boiler-plate is needed here.

skstore/ts/src/skstore_api.ts
```diff
@@ -331,13 +308,13 @@ export interface TableHandle<R extends TJSON[]> {
...
-  >(
-    mapper: C,
-    ...params: EMParameters<K, V, R, C>
+  map<K extends TJSON, V extends TJSON>(
+    mapper: new (
+      eparams: EHandle<any, any>[],
+      lparams: LHandle<any, any>[],
+    ) => EntryMapper<R, K, V>,
+    eparams: EHandle<any, any>[],
+    lparams: LHandle<any, any>[],
```
bennostein, line 317: Could cut down boilerplate further by providing default (empty-array) arguments here (and similar for other map functions throughout)

```suggestion
    eparams: EHandle<any, any>[] = [],
    lparams: LHandle<any, any>[] = [],
```

jberdine: I don't know JS well enough to know how it works when you have two optional args and you only want to pass one. I need to see if that is that common/easy.

jberdine:
Just for clarity, this is just one approach, and there is at least one more that I will investigate. So definitely don't spend any great amount of time reviewing this.

skstore/ts/src/skstore_api.ts
bennostein: you can only leave off (positional) optional arguments after the last provided argument.

so if I had `foo(x=0,y=0,z=0) {...}` and wanted to provide `y`, I would have to provide `x` too but could still elide `z`.

pikatchu:
I think we can close this. @skiplabsdaniel has a more flexible approach that seems to make TS type-inference happy.

jberdine:
> I think we can close this. @skiplabsdaniel has a more flexible approach that seems to make TS type-inference happy.

Yes, agreed. I think there is still the question of whether we want to try to give users some help from the type-checker to ensure that all the parameters to class constructors are handles, or if the best we can do is to only check this at runtime. Thoughts?

PR #281, opened 2024-08-16
[skstore:api] Decompose the API according to the number of parameters

skiplabsdaniel:
Facilitates typescript inference, reducing the need to annotate types


pikatchu:
Awesome! What is the difference between mapAll and map?

skiplabsdaniel:
mapAll is the function that allow to map with a constructor that have any number of parameter, ( Probably need to be rename mapAny or mapN ), and map is to map with a constructor that have no parameter.

PR #283, opened 2024-08-20
[skstore] manage deep freeze, add parameters check

skiplabsdaniel:


skstore/ts/src/prv/skstore_impl.ts
```diff
@@ -1765,3 +1799,33 @@ function toUpdateWhereQuery(
...
+  } else if (type == "object") {
+    const jso = value as any;
+    if ((value as any).__sk_frozen) {
+      return;
+    } else if (jso.constructor.name) {
+      throw new Error(
+        "class '" + jso.constructor.name + "' cannot be manage by skstore.",
+      );
+    } else if (Object.isFrozen(jso)) {
+      if (Array.isArray(jso)) {
```
pikatchu, line 1824: So arrays cost of O(n) every time?
Could we do better?

skstore/ts/src/prv/skstore_impl.ts
```diff
@@ -1765,3 +1799,33 @@ function toUpdateWhereQuery(
...
+    return;
+  } else if (type == "object") {
+    const jso = value as any;
+    if ((value as any).__sk_frozen) {
+      return;
+    } else if (jso.constructor.name) {
+      throw new Error(
+        "class '" + jso.constructor.name + "' cannot be manage by skstore.",
+      );
+    } else if (Object.isFrozen(jso)) {
```
pikatchu, line 1823: Why is this allowed? Why not only support objects with the tag __sk_frozen?
Is it for convenience when passing arrays?

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -141,8 +142,12 @@ export const reactiveObject = {
     if (idx === undefined) return undefined;
     return getValueAt(hdl, idx, false);
   },
+  set(_hdl: WasmHandle, _prop: string, _value: any) {
```
pikatchu, line 145: Good :)

pikatchu (commented):
Great PR! I just would like to better understand why arrays work the way they do. Could we not also stamp them with a tag? It could not be a property, because that would affect the keys, but maybe in the prototype?

skstore/ts/src/prv/skstore_impl.ts
skiplabsdaniel: This is only the case if the Object.freeze function is used and not the one supplied. Otherwise, the Array is tagged in the same way as an object, and when the check is run, it's returned earlier.

skstore/ts/src/prv/skstore_impl.ts
skiplabsdaniel: It was just temporary.
I needed to do some testing.
No more the case.

jberdine:
Do I understand correctly that with this, all the parameters to mapper class constructors are checked to already be deeply frozen at runtime. So the expectation is that users will themselves call freeze on the parameters before calling skstore map functions? For this purpose, there is a (deep) `freeze` function in skstore.ts. But if users use the shallow `Object.freeze` recursively over the whole object graph, it will still work, but the check will be slower.

Would it be a simpler and more user-friendly API to instead have the skstore functions do the deep freezing themselves wherever necessary, and provide users a function to make a deeply frozen clone of an object that they could use in the cases where they want to keep an unfrozen copy of the parameter?

Is the main question whether it is better to make users explicitly call (deep) `freeze` on map parameters that are neither primitive values nor skstore handles, to put it right in front of their faces that there is a deep immutable copy when reading the code, or to implicitly make the copy ourselves so that the user code looks simpler in the usual case?

Either way, I think it would be simpler, clearer, and harder to misuse if we renamed `freeze` to `deepFreeze` and made our `check` function fail unless that one was called, removing the slow path that lets users use `Object.freeze`.

Freezing an already frozen object is fast, so I guess there is no performance consideration to favor one approach over the other.

PR #284, opened 2024-08-20
[skdb] Support FFI interface.

beauby:
A few changes to allow using SKDB from rust via FFI (basically enabling the rust and python clients available at [skiplabs/skdb-rs](https://github.com/SkipLabs/skdb-rs)).

### DIFF
--- a/prelude/src/stdlib/ffi.sk
+++ b/prelude/src/stdlib/ffi.sk
@@ -1,0 +1,7 @@
+module FFI;
+
+// TODO: This should probably return a Result<> in case invalid utf8 is provided.
+@cpp_extern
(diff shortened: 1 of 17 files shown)


prelude/src/stdlib/ffi.sk
```diff
@@ -0,0 +1,7 @@
+module FFI;
+
+// TODO: This should probably return a Result<> in case invalid utf8 is provided.
+@cpp_extern
```
jberdine, line 4: Why not `@cpp_extern("sk_string_create")` here and then no need to define `SKIP_FFI_utf8_bytes_to_string` in string.c?

PR #285, opened 2024-08-20
[JS] Extend "get" interfaces and apply consistent naming

bennostein:
Previously, JS handles on SKStore collections had inconsistent/potentially-confusing "get" operations: eager handles had `get : K -> V`  and `maybeGet : K -> ?V`, and lazy handles had just `get : K -> V`.

These suffice for cases where there's at most one value per key, but are specialized versions of the generic `get` operation which can return multiple values (as in e.g. `getArray` or `getIter` in the skip implementation).

So, I've exposed more operations to the JS runtime and renamed things consistently, offering `get : K -> V[]`, `getSingle : K -> V`, and `maybeGetSingle : K -> ?V` on all JS handles.  Open to suggestion on renamings -- I settled on this to make as clear as possible that `get` is the generic/natural way of accessing collections and `getSingle`/`maybeGetSingle` are specialized cases.

### DIFF
--- a/skstore/ts/src/skstore_api.ts
+++ b/skstore/ts/src/skstore_api.ts
@@ -235,14 +235,28 @@ export interface NonEmptyIterator<T> {
 
 /**
  * A _Lazy_ Handle on a reactive collection, whose values are computed only when queried
- * using `get`
  */
 export interface LHandle<K extends TJSON, V extends TJSON> {
   /**
-   * Get (and potentially compute) a value of a lazy reactive collection
+   * Get (and potentially compute) all values mapped to by some key of a lazy reactive
+   * collection.
    * @throws {Error} when the key does not exist
    */
-  get(key: K): V;
+  get(key: K): V[];
+
+  /**
+   * Get (and potentially compute) a value of a lazy reactive collection.
+   * If multiple values are mapped to by the key, any of them can be returned
+   * @throws {Error} when the key does not exist
+   */
+  getSingle(key: K): V;
+
+  /**
+   * Get (and potentially compute) a value of a lazy reactive collection, if one exists.
+   * If multiple values are mapped to by the key, any of them can be returned.
+   * @throws {Error} when the key does not exist
+   */
+  maybeGetSingle(key: K): Opt<V>;
 }
 
 export interface ALHandle<K extends TJSON, V extends TJSON, M extends TJSON>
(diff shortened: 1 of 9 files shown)


pikatchu:
What is the reasoning behind the name getSingle? Is the idea that the preferred way of getting values should be getIter?

If so, shall we remove it altogether?

@skiplabsdaniel ?

bennostein:
My reasoning is: a collection is a map from keys to arbitrarily many values.  That's the mental model we want people to have, that's what you operate over when you map, etc.  So, the "default" `get` should have that shape -- returning 1 or more values for the given key.  This is what we call `getArray` or `getIter` in Skip -- here I'm proposing that we just call it `get` in JS for discoverability/to nudge people towards using the general-case operation.

Sometimes, your program will be set up so that each key only has one value, so you actually want an accessor of type `K -> V`.  But that's a special case which may fail if there are many values, so it should have a name making that clear.  I chose `getSingle` but it could also be `getUnique`, `getOnlyValue`, etc. I don't feel too strongly what the particular name is, just that it feels wrong for an unsafe special-case to be called `get`

skstore/ts/src/skstore_api.ts
```diff
@@ -235,14 +235,28 @@ export interface NonEmptyIterator<T> {
...
    */
-  get(key: K): V;
+  getArray(key: K): V[];
+
+  /**
+   * Get (and potentially compute) a value of a lazy reactive collection.
+   * If multiple values are mapped to by the key, any of them can be returned
+   * @throws {Error} when the key does not exist
+   */
+  getFirst(key: K): V;
```
jberdine: FWIW I find the `getFirst` name a bit odd when any of the multiple keys can be returned. To me `first` indicates an ordering. Maybe `getOne` or `choose`?

skstore/ts/src/skstore_api.ts
```diff
@@ -235,14 +235,28 @@ export interface NonEmptyIterator<T> {
...
  */
 export interface LHandle<K extends TJSON, V extends TJSON> {
   /**
-   * Get (and potentially compute) a value of a lazy reactive collection
+   * Get (and potentially compute) all values mapped to by some key of a lazy reactive
+   * collection.
    * @throws {Error} when the key does not exist
    */
-  get(key: K): V;
+  getArray(key: K): V[];
```
jberdine: Maybe I'm confused, but is there actually a distinction between a key "not existing" and mapping to an empty array of values? Looking at the skip code, it seems that there is no throw case for `getArray`, and that the a key "existing" amounts to it mapping to a non-empty array of values.

skstore/ts/src/skstore_api.ts
```diff
@@ -235,14 +235,28 @@ export interface NonEmptyIterator<T> {
...
+   * @throws {Error} when the key does not exist
+   */
+  getFirst(key: K): V;
+
+  /**
+   * Get (and potentially compute) a value of a lazy reactive collection, if one exists.
+   * If multiple values are mapped to by the key, any of them can be returned.
+   * @throws {Error} when the key does not exist
+   */
+  maybeGetFirst(key: K): Opt<V>;
```
jberdine: Here it is also unclear what the difference between returning None vs throwing is.

skstore/ts/src/skstore_api.ts
bennostein: Right, thanks -- copied too much of the docstring there from `get` (which does throw) and should weaken the language for `maybe`/`array` versions.

skstore/ts/src/skstore_api.ts
bennostein: Renamed to `getOne` after chatting with Julien today!  Thanks for the catch/suggestion.

PR #286, opened 2024-08-21
[JS] Add support for mappers returning multiple values per key

bennostein:
Previously, if a mapper returned multiple values bound to the same key then only one was written and the others were silently dropped.  I changed the behaviour here to match my intuition/expectation that they would all be written.

This came up in #285 but is somewhat orthogonal and changes semantics instead of just extending, so submitting separately.

### DIFF
--- a/skstore/src/Extern.sk
+++ b/skstore/src/Extern.sk
@@ -96,6 +96,18 @@ fun writerSet(
   writer.set(JSONID(key), JSONFile(value))
 }
 
+@export("SKIP_SKStore_writerSetArray")
+fun writerSetArray(
+  writer: mutable TWriter<JSONID, JSONFile>,
+  key: SKJSON.CJSON,
+  values: SKJSON.CJArray,
+): void {
+  files = values match {
+  | SKJSON.CJArray(xs) -> xs.map(x -> JSONFile(x))
+  };
+  writer.setArray(JSONID(key), files)
+}
+
 @export("SKIP_SKStore_iteratorFirst")
 fun iteratorFirst(values: mutable NonEmptyIterator<JSONFile>): SKJSON.CJSON {
   if (values.isPastFirstValue) {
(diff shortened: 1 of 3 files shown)


skiplabsdaniel:
It seems a dangerous modification at this stage, as it breaks the semantics of SKStore, which is one value per source. If you need several values per source, you can make an array. The fact that this is done silently is indeed a problem, an exception should be returned on the Skip side, and maybe a map should be returned on the TS side rather than an Iterable of array.

pikatchu:
Just spoke with Daniel over the phone. This is incorrect "it breaks the semantics of SKStore, which is one value per source". SKStore supports multiple value per source (an array of value to be precise). Daniel never noticed because he never used it this way, but it's there.

I think there are cases where you would want to "set" a whole array at once, so it would make sense to have a primitive called `setArray`.

There are other cases where you might want to accumulate, for example, you are walking a datastructure and you want to accumulate, so it would make sense to also implement something like `append`.

What is weird is to mix the two ... I would suggest to throw an exception if setArray is called when the key already holds a value.

bennostein:
I agree an `append` might be nice but we'd have to design a user-facing API for it that makes sense. Not sure exactly what that would look like -- the current semantics are designed for our model of "write the logic for the current state and we'll update everything for you", and allowing to accumulate instead at the same source might lead to some of those invariants breaking.  Will land this for now and we can of course add that later if/when we've got a use case for it.

skiplabsdaniel:
If there are several write functions at writer level, it's better to expose them to the user in the mapper (To allow fine-tuned management) instead of returning an iterable.

bennostein:
This behaves exactly the same as the previous version except that it doesn't drop multiple bindings on the same key -- what's the fine-tuned management that you have in mind?

skiplabsdaniel:
You're right, I can manage my needs with javascript code. I need to rest.

PR #288, opened 2024-08-27
Fix skstore/ts examples: maybeGetFirst -> maybeGetOne

mbouaziz:
Just tried `make check` in `skstore/`... and fixed according to your recent rename.


bennostein (approved):
Ahh, missed these, thanks Mehdi!

PR #289, opened 2024-08-27
Add a `ValueMapper` interface for structure-preserving maps

bennostein:
`ValueMapper` is a specialized form of `Mapper` which re-uses the input collection's key structure in the output collection and thus doesn't need to consider keys.

For cases where users just want to map values and preserve the key structure, this will simplify and cut down on boilerplate -- instead of defining the fully-general mapper `(K1, NonEmptyIterator<V1>) => Iterable<[K2, V2]>`, they can just write the `(V1) => V2` transformation directly.

### DIFF
--- a/skstore/ts/src/prv/skstore_impl.ts
+++ b/skstore/ts/src/prv/skstore_impl.ts
@@ -84,7 +84,7 @@ class EHandleImpl<K extends TJSON, V extends TJSON> implements EHandle<K, V> {
     const mapperObj = new mapper(...params);
     Object.freeze(mapperObj);
     if (!mapperObj.constructor.name) {
-      throw new Error("The class must have a name.");
+      throw new Error("Mapper classes must be defined at top-level.");
     }
     const eagerHdl = this.context.map(
       this.eagerHdl,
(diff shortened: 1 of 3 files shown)


skstore/ts/src/prv/skstore_impl.ts
```diff
@@ -84,7 +86,7 @@ class EHandleImpl<K extends TJSON, V extends TJSON> implements EHandle<K, V> {
     const mapperObj = new mapper(...params);
     Object.freeze(mapperObj);
     if (!mapperObj.constructor.name) {
-      throw new Error("The class must have a name.");
+      throw new Error("Mapper classes must be named.");
```
jberdine: Is what we really want `Mapper classes must be defined at top-level` and being named is just a way to approximately check that? I wonder if a user seeing this error for an attempt to use a nested class would react by trying to add a name property to their class.

skstore/ts/src/prv/skstore_impl.ts
```diff
@@ -239,6 +241,164 @@ class EHandleImpl<K extends TJSON, V extends TJSON> implements EHandle<K, V> {
     return this.mapN(mapper, p1, p2, p3, p4, p5, p6, p7, p8, p9);
   }
 
+  mapValuesN<
+    V2 extends TJSON,
```
jberdine: I wonder what the feeling is about using `V` and `V2` as opposed to e.g. `V` and `W`? My impression is that, especially when speaking out loud, it is easier to have either all variables like this be numbered, or to use different letters, but mixing is harder. When there are only a few, my feeling is that using different letters is easier.

(I realize this is not specific to this PR, and feel free to ignore, but it seems like a reasonable point to discuss the question.)

bennostein: I don't have strong feelings about this so switched from V2 to W throughout!  I agree that mixing numbered and non-numbered type variables gets a bit confusing.

PR #291, opened 2024-08-27
Forbid NaN as a collection key

bennostein:
Add a dynamic check that mappers don't produce NaN keys, since those mess up the comparisons and orderings that we need to keep everything happy on the Skip side

### DIFF
--- a/skstore/ts/src/prv/skstore_impl.ts
+++ b/skstore/ts/src/prv/skstore_impl.ts
@@ -37,6 +37,18 @@ import type { MirrorDefn, Params, SKDBSync } from "#skdb/skdb_types.js";
 
 type Query = { query: string; params?: JSONObject };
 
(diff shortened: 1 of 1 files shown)


gregsexton:
What about null and undefined?

Undefined compares as greater than every (maybe most) things. So maybe it is ok?

Hilariously though, null is coerced to a string and sorted by code point so will likely end up in the middle of a collection. Source: https://devtips.theanubhav.com/posts/undefined-sort

gregsexton:
I guess neither of these are actual problems for comparison though, unlike NaN.

bennostein:
Yep, null should be fine as it compares/orders sensibly in both JS and Skip (in its embedded representation).  but NaN non-reflexive equality is borked at the level of the IEEE spec

undefined is forbidden at the typescript level -- it's not a member of the `TJSON` type bounding all of our type variables.  But probably still worth checking here, given the amount of trust merited by the typescript type system...

PR #292, opened 2024-08-27
Apply OO style fixes to simplify API; run full sknpm check on ./skstore

bennostein:
This applies some suggestions from @skiplabsdaniel on proper javascript OO style, and in doing so simplifies the interfaces I added in #289 

Also runs and applies fixes from `cd ./skstore && ../build/sknpm check --test`, which runs a more thorough typescript check than the `sknpm test` I had been using.  Pending changes from Daniel/Lukas to get that wired into CI.


skstore/ts/src/skstore_api.ts
```diff
@@ -167,16 +167,24 @@ export type MParameters<
...
-export type VMParameters<
+export abstract class ValueMapper<
+  K extends TJSON,
   V1 extends TJSON,
   V2 extends TJSON,
-  C extends new (...params: Param[]) => ValueMapper<V1, V2>,
-> = C extends new (...params: infer P) => ValueMapper<V1, V2> ? P : never;
+> implements Mapper<K, V1, K, V2>
+{
+  abstract mapValue(value: V1): V2;
```
mbouaziz: This function could take the key as second argument, it's usual in the javascript world

bennostein: Do you mean like in 3df1c39 ? nice! not sure when you'd use that over just the value, but might as well have the option

PR #293, opened 2024-08-29
Lint SkStore

mbouaziz:
Add `ESLint` and enable more strict typechecking rules for `skstore/ts/src`.
Only minor changes, no red flags were found.
A few rules are still disabled because they were too noisy.

To run the linter: `npm run lint`

### DIFF
--- a/prelude/ts/src/sk_types.ts
+++ b/prelude/ts/src/sk_types.ts
@@ -702,7 +702,7 @@ export interface Text {
 export class Raw implements Text {
   text: string;
 
-  constructor(text: string, category?: string) {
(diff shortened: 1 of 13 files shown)


bennostein:
lints all look good to me! (modulo the type error in skstore-wasm test -- that's the result of adding more strict rules to the CI check, right?)

mbouaziz:
#294 should fix this

mbouaziz:
Maybe non-`Error` exceptions were intended. @skiplabsdaniel, were they?

skiplabsdaniel:
LGTM

PR #294, opened 2024-08-29
[sknpm] Add modules instead of replacing them

mbouaziz:
The dark magic of `sknpm` replaces the line `/*--MODULES--*/` with several lines of `import` followed by `modules = [...];`, expecting a pre-existing `var modules = [];`.
Linters expect this line to be `const modules = [];` because the array doesn't seem to be changed.
Fine, let's push modules instead of replacing the contents of the array, which was brittle if we didn't start with an empty array.


skiplabsdaniel:
LGTM

PR #295, opened 2024-08-29
Simplify parameters

mbouaziz:
Quantify over parameters rather than a constructor.

I'm wondering whether this would be enough to get rid of `map1,2,...`

Only 1 commit, stacked over #293

`make -C skstore check` is happy

### DIFF
--- a/skstore/ts/src/prv/skstore_impl.ts
+++ b/skstore/ts/src/prv/skstore_impl.ts
@@ -19,13 +19,8 @@ import type {
   JSONObject,
   TJSON,
   Param,
-  EMParameters,
-  MParameters,
-  OMParameters,
   LazyCompute,
-  LParameters,
   AsyncLazyCompute,
-  ALParameters,
   NonEmptyIterator,
   ALHandle,
 } from "../skstore_api.js";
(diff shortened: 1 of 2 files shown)


jberdine:
Do you get the same type errors when passing the wrong parameters to map functions in clients before and after this change?

mbouaziz:
> Do you get the same type errors when passing the wrong parameters to map functions in clients before and after this change?

It seems so.
We need to add except-error tests to make sure we don't break this statement.

Btw, I could get rid of the decomposition of the API and Typescript seems happy: https://github.com/mbouaziz/skdb/tree/undecompose

Am I missing something?

jberdine:
Also check things like in the examples directory (make check in skstore dir). A previous iteration did not split out the various arities, but typescript inference went bonkers. It is easy to make it happy, harder to make it unhappy when it should be.

jberdine:
But for sure it not working well before doesn't mean it can't. Maybe parameterizing on the params type rather than the constructor helps. I'm wary though. If an actual constructor take two params of different types, does the simplified version correctly check the number and types of params at the call sites?

mbouaziz:
`make -C skstore check` is happy.

> If an actual constructor take two params of different types, does the simplified version correctly check the number and types of params at the call sites?

From what I tested, yes.

skiplabsdaniel:
LGTM

PR #296, opened 2024-08-29
[nfc] Improve consistency of terminology/naming

bennostein:
Here's a first batch of renamings based on the glossary I posted in discord, making sure we consistently use "Collection" in code and comments and renaming EntryMapper/TableMapper to InputMapper (for symmetry/consistency with OutputMapper).  Should be no functional change


Makefile
```diff
@@ -129,7 +129,7 @@ setup-git-hooks: .git/hooks/pre-commit
 
 .PHONY: test
 test:
-	$(MAKE) --keep-going SKARGO_PROFILE=dev SKDB_WASM=sql/target/wasm32-unknown-unknown/dev/skdb.wasm SKDB_BIN=sql/target/host/dev/skdb test-skfs test-native tstest-sql
+	$(MAKE) --keep-going SKARGO_PROFILE=dev SKDB_WASM=sql/target/wasm32-unknown-unknown/dev/skdb.wasm SKDB_BIN=sql/target/host/dev/skdb sktest-prelude test-native tstest-sql tstest-skstore
```
jberdine, line 132: The definition of the test-skfs target below should probably also be removed

jberdine (approved):
This LGTM. Not necessarily about this PR, but reading it made me think of it. When I see the term "name" it is not clear whether or not the name values are meant to be unique identifiers. Maybe it is just me. But in places where the name is simply informative, I think it would make sense to use the term "name", while in places where a unique identifier is needed, the term "id" or similar would be clearer.

jberdine:
Another question is whether "TableCollection" is redundant in the sense that all tables are collections (of rows), so why not just "Table"? Or perhaps "EagerTable" to emphasize that they are EagerCollections.

bennostein:
> Another question is whether "TableCollection" is redundant in the sense that all tables are collections (of rows), so why not just "Table"? Or perhaps "EagerTable" to emphasize that they are EagerCollections.

Ahh, this one point had a bunch of discussion between me and Daniel: I wanted to just call this "Table" as you suggest.  But there's a different interface "Table" which provides SQL-like select/insert/update/delete/etc. operations.  So, in a sense, this is a "read-only" table which can be used in mappers, while that's a "read-write" table which very much can not be used in mappers.  This was what I came up with to convey that this is a table-shaped type of collection which can be used like other collections (i.e. in mappers) but I'm open to other versions too.  e.g. I considered calling this `Table` and that `WritableTable` but couldn't come up with a way to explain that without getting too down in the weeds

Makefile
bennostein: I left it in as it's clearly labelled as for "backwards compatibility" and I wasn't sure if anything/anyone else was depending on it -- there's nothing in the repo that does, so might as well remove I suppose

jberdine:
> > Another question is whether "TableCollection" is redundant in the sense that all tables are collections (of rows), so why not just "Table"? Or perhaps "EagerTable" to emphasize that they are EagerCollections.
> 
> Ahh, this one point had a bunch of discussion between me and Daniel: I wanted to just call this "Table" as you suggest. But there's a different interface "Table" which provides SQL-like select/insert/update/delete/etc. operations. So, in a sense, this is a "read-only" table which can be used in mappers, while that's a "read-write" table which very much can not be used in mappers. This was what I came up with to convey that this is a table-shaped type of collection which can be used like other collections (i.e. in mappers) but I'm open to other versions too. e.g. I considered calling this `Table` and that `WritableTable` but couldn't come up with a way to explain that without getting too down in the weeds

Ah ok then, `TableCollection` makes sense. I will have this naming question in mind, but looks good as is.

bennostein:
Closing, subsumed into #300 .

PR #299, opened 2024-08-30
[RFC] Stronger type for ptr

mbouaziz:
In JS, we use `type ptr = number` to represent pointers. Nothing prevents us from mixing `ptr` and `number`.
A first step is to make `ptr` (what Typescript people call) _opaque_ (it corresponds to OCaml `private` types, not abstract types), i.e. it is `number & fake_something` so that we can convert `ptr` to `number` but not the other way around, unless using `as` or `any`, of course...
Under undefined circumstances (pun intended), it guarantees we're not mixing both types.

The second step is to make `ptr` parametric so we know what they refer to in the Skip world (after being compiled to LLVM and Wasm). Again, it avoids mixing things like `ptr<int>` with `ptr<string>`. Like in C, you can always can between pointer, via `void*` (`ptr<any>`).

### DIFF
--- a/skstore/ts/src/prv/skstore_skjson.ts
+++ b/skstore/ts/src/prv/skstore_skjson.ts
@@ -388,7 +388,7 @@ class LinksImpl implements Links {
         return fromWasm.SKIP_SKJSON_endCJArray(arr);
       } else if (typeof value == "object") {
         if (value.__isObjectProxy || value.__isArrayProxy) {
-          return value.__pointer as number;
+          return value.__pointer as ptr;
         } else {
           const obj = fromWasm.SKIP_SKJSON_startCJObject();
           for (const key of Object.keys(value as object)) {
--- a/skstore/ts/src/prv/skstore_skjson.ts
+++ b/skstore/ts/src/prv/skstore_skjson.ts
@@ -1,5 +1,6 @@
 // prettier-ignore
 import type { int, ptr, float, Links, Utils, ToWasmManager, Environment, Opt, Shared, } from "#std/sk_types.js";
+import type * as Internal from "./skstore_internal_types.js";
 
 export enum Type {
   /* eslint-disable no-unused-vars */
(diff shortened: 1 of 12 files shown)


skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -13,34 +14,43 @@ export enum Type {
...
+  SKIP_SKJSON_asFloat: (json: ptr<Internal.CJSON>) => float;
+  SKIP_SKJSON_asBoolean: (json: ptr<Internal.CJSON>) => boolean;
+  SKIP_SKJSON_asString: (json: ptr<Internal.CJSON>) => ptr<Internal.String>;
+  SKIP_SKJSON_asArray: (json: ptr<Internal.CJSON>) => ptr<Internal.CJArray>;
+  SKIP_SKJSON_asObject: (json: ptr<Internal.CJSON>) => ptr<Internal.CJObject>;
+
+  SKIP_SKJSON_fieldAt: (
+    json: ptr<Internal.CJObject>,
+    idx: int,
+  ) => ptr<Internal.String>; // Should be Opt<...>
```
bennostein, line 30: Are the semantics here that this comes back as a zero pointer if it's None on the skip side?

I ran into a few cases like that and added some "if the (skip) pointer is 0 just use a (javascript) null instead of trying to dereference" logic in previous PRs.  It's a bit odd how "Opt" in JS means "nullable" and not the usual tagged sum, but given that interpretation this seems fine as is -- do you agree or plan to make this an explicit option?

skstore/ts/src/prv/skstore_skjson.ts
```diff
@@ -267,17 +277,30 @@ function clone<T>(value: T): T {
...
-  SKIP_SKJSON_endCJObject: (obj: ptr) => ptr;
-  SKIP_SKJSON_startCJArray: () => ptr;
-  SKIP_SKJSON_addToCJArray: (arr: ptr, value: ptr) => void;
-  SKIP_SKJSON_endCJArray: (arr: ptr) => ptr;
-  SKIP_SKJSON_createCJNull: () => ptr;
-  SKIP_SKJSON_createCJInt: (v: int) => ptr;
-  SKIP_SKJSON_createCJFloat: (v: float) => ptr;
-  SKIP_SKJSON_createCJString: (str: ptr) => ptr;
-  SKIP_SKJSON_createCJBool: (v: boolean) => ptr;
+  SKIP_SKJSON_startCJObject: () => ptr<Internal.VectorPairStringCJSON>;
```
bennostein: Given that this (and similarly for VectorCJSON a few lines down) are only used here, could we drop those specialized definitions from skstore_internal_types and just define aliases here? opened a PR against your branch showing what I mean

bennostein (approved):
Looks great(!) to me, nice idea and implementation.  A couple low level comments/questions/suggestions that you should feel free to ignore, looks solid as is.

skstore/ts/src/prv/skstore_skjson.ts
mbouaziz: I don't know, I've seen a few places where the `Option<T>` was translated into `Opt<ptr>` (btw `Opt<T>` is not standard Typescript, it is defined in `sk_types` as `T | null`).
Maybe @skiplabsdaniel knows more

skstore/ts/src/prv/skstore_skjson.ts
mbouaziz: You're right, I'll do that

bennostein: done in https://github.com/mbouaziz/skdb/pull/1 if that looks alright to you :)

skstore/ts/src/prv/skstore_skjson.ts
bennostein: > btw `Opt<T>` is not standard Typescript, it is defined in `sk_types`

Yep, sorry, that's what I meant by "in JS".  I'm happy i you want to make it `Opt<ptr<...>>`, I just mean that the actual value there may be the number `0` not the value `null` -- so if we do make it `Opt`, the implementation should do that check on the value coming back from skip

jberdine: I'm not sure if it makes it to the js interface, but note that the skip option type is represented as a flat null pointer for `None` or the value of the `Some` case directly where possible. In C interop code one has to take account of that when manipulating values of option type. I don't see how the wasm code would be different, so that mapping to JS null is likely the most accurate.

PR #300, opened 2024-09-02
[TS skip runtime] Rename source files, directories, package artifacts

bennostein:
This one does the proper renaming of files, package artifacts, and
directories to strip out references to "SKStore" from the user-facing
interface in favor of "Skip Runtime".

In anticipation of future language bindings, I've called the top-level
directory in the repo `skipruntime-ts`, so we can do e.g.
`skipruntime-java`, `skipruntime-py`, `skipruntime-rs`, etc. going
forward. (or combine and just name as `skipruntime`)

I've renamed the package to `skip-runtime` and made the according
changes throughout.

User-facing TS source files `./skstore_*.ts` became `./skipruntime_*.ts`,
except for the package top-level `skip-runtime.ts`.

(5 commits, stacked on #296)


bennostein:
To be clear: consider this an RFC on this particular naming/structure -- open to other suggestions!

bennostein:
Also, note that the test failure is just due to out-of-date prebuilt compiler/runtime with `skfs`-referencing names.  Will push an updated docker image when this lands, all is green with a runtime built on this branch.

jberdine (commented):
This LGTM but I don't understand why we end up in a place with both "skip runtime" and "skruntime". Having various hyphenation and capitalization forms to fit the various conventions that are in place in the different ecosystems makes sense, but it seems superfolous to have both "skip runtime" and the only slightly shorter "skruntime". Am I missing something?

bennostein:
> why we end up in a place with both "skip runtime" and "skruntime"?

My initial thinking was to match the other toplevel projects in this repo: skbuild, skdate, sknpm, sktest, etc.  The name of that directory is not a huge deal as it won't be visible to users, but it bled into the TS sources themselves for... no real a priori reason I can think of now.  Though it is fun to say ;)

I'll go through and switch sources from `skruntime_*.ts` to `skip_runtime_*.ts`; I agree that consistency there would be good.

PR #302, opened 2024-09-02
Undecompose API according to the number of parameters

mbouaziz:
With #295, it looks like we don't need this decomposition anymore.

Let me know if you had specific cases where it was needed, we should add expect-error tests to make sure we can rely on typescript to infer things correctly, at least in regular use cases


pikatchu:
SHIPIT!

pretty sure @skiplabsdaniel will be ok with it.

PR #308, opened 2024-09-04
First iteration of API docs workflow

jberdine:
This PR adds a template Docusaurus site and adds api docs generated by
typedoc to it using the docusaurus-plugin-typedoc Docusaurus
plugin. This needs the development/source docs site to be in the same
repo as the sources, as there are relative paths in the docusaurus
config file to the sources. A separate deployment step will be needed
to generate the docs and push them to the repo that feeds the live
site. This also means that we can iterate on the docs in the source
repo without publishing them immediately, and then choose to publish
them in sync with publishing packages.

The workflow should be just:
```
- $ cd /path/to/repo/root
- $ make run-docs # to run the docs site locally
- visit http://localhost:3000/docs/api
- edit the TSDoc comments in the sources
- $ cd /path/to/repo/root
- $ make docs # to regenerate the api docs
```

The documentation for the tags that typedoc recognizes is
[here](https://typedoc.org/guides/tags/).

The API docs themselves look less than great. There is an alternative
plugin to link Docusaurus and typedoc that seems to generate better
looking output,
[docusaurus-plugin-typedoc-api](https://github.com/milesj/docusaurus-plugin-typedoc-api),
but it is currently incompatible with Docusaurus 3.5.


jberdine:
Some screenshots just for reference:
image
image

skipruntime-ts/package.json
jberdine, line 1: @skiplabsdaniel after the fact I see that sknpm is looking for a file named npm.json, and maybe I broke something with this renaming. Is the file sknpm looks for meant to be different from the package.json files that other tools look for? Should we have both files, or change sknpm to look for package.json?

PR #309, opened 2024-09-04
[skipruntime-ts] More types

mbouaziz:
While adding code, I'm always hurting myself with weakly-typed stuff... Here are a few more (more-or-less) typed things: proxies, WasmHandle, handles, importJSON, exportJSON.

### DIFF
--- a/prelude/ts/src/sk_types.ts
+++ b/prelude/ts/src/sk_types.ts
@@ -951,12 +951,15 @@ export async function runUrl(
   return await start(modules, buffer, env, main);
 }
 
+export const sk_isArrayProxy: unique symbol = Symbol();
+export const sk_isObjectProxy: unique symbol = Symbol();
+
 export function cloneIfProxy<T>(v: T): T {
   if (
     v !== null &&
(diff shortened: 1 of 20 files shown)


prelude/ts/src/sk_types.ts
```diff
@@ -959,3 +950,15 @@ export async function runUrl(
...
+export function cloneIfProxy<T>(v: T): T {
+  if (
+    v !== null &&
+    typeof v === "object" &&
+    (("__isArrayProxy" in v && v.__isArrayProxy) ||
+      ("__isObjectProxy" in v && v.__isObjectProxy)) &&
+    "clone" in v
+  )
+    return (v.clone as () => T)();
+  return v;
```
jberdine: You don't like braces?

jberdine (approved):

prelude/ts/src/sk_types.ts
mbouaziz: Fixed. I wrongly assumed the formatter would add them for me

PR #310, opened 2024-09-05
SkJSON/skruntime: make CJArray parametric

mbouaziz:
Allows to reduce runtime checks when deconstructing well-typed JSON arrays.

Based on #309


mbouaziz:
I couldn't use `CJArray` for ranges because it needed to be optional and we get `CJNull()` rather `None()` so I had to use `CJSON` instead of a more precise type

PR #311, opened 2024-09-05
Add ranges to SKRuntime map/mapReduce

mbouaziz:


skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -32,6 +32,13 @@ export type MirrorSchema = {
   expected: ColumnSchema[];
   filter?: DBFilter;
 };
+export class MapOptions<K extends TJSON> {
+  constructor(public ranges: [K, K][] | null = null) {}
+}
+export type WithOptions<Params extends Param[], K extends TJSON> = [
+  ...Params,
+  MapOptions<K>?,
+];
```
jberdine: I'm not sure I understand this. The `MapOptions` is optional, which is the same as `MapOptions<K> | undefined` (right?), so if it is omitted, there will be some unknown number of `Params` followed by an `undefined`, right? That doesn't seem to agree with the imlementation of `splitMapParams` which uses `instanceof MapOptions`. Am I confused?

mbouaziz: `T?` in a tuple means the element can be omitted

BUT you are right, it also means it can be present with type `undefined`, which is not what I want, unfortunately.

This is indeed ambiguous: when the user writes `(..., undefined)` we don't know whether they meant no `MapOptions` (and the last parameter is `undefined`) or an undefined `MapOptions`.

I've added a last commit which differentiates the two cases in the frontend API and prevents the user from passing an `undefined` `MapOptions`.

There's still a possible ambiguity if the user has a mapper whose last parameter is a `MapOptions`, but why would they?
Maybe I could rely on a function which creates a record with a private unique symbol instead of relying on classes + `instanceof`. Not sure it's worth it for now

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -322,6 +325,16 @@ export interface EagerCollection<K extends TJSON, V extends TJSON> {
...
 
+  /**
+   * Create a new eager collection by mapping some computation over this one (with options)
+   * @param {Mapper} mapper - function to apply to each element of this collection
+   * @returns {EagerCollection} The resulting (eager) output collection
+   */
+  map<K2 extends TJSON, V2 extends TJSON, Params extends Param[]>(
+    mapper: new (...params: Params) => Mapper<K, V, K2, V2>,
+    ...paramsAndOptions: [...Params, MapOptions<K>]
+  ): EagerCollection<K2, V2>;
```
jberdine, line 338: With 2 declarations like this, would it be possible to make the `MapOptions` not optional in the underlying implementation, so that the first `map` fills in the default value of `MapOptions`. The `MapOptions` argument could be moved to come before the `...params`, and then maybe no magic fiddling with things coming after the rest args based on type tags or whatever.

mbouaziz: Yes and no.
Moving the options before the arguments is possible, indeed.
But in the implementation, there's gonna be some magics anyway because there can only be a single implementation for the two signatures (overloading is a typescript thing and doesn't exist once translated into javascript).

Whether we want the options before the parameters or after, I don't care.

Before probably reduces the fiddling in the implementation. But does it make the API simpler? What is better from the user point of view?
I've put the options at the end because they are optional and that's where people expect optional things to be.

jberdine: Ok, I see, if there can be only 1 implementation it doesn't really help. I guess opinions vary, but I think I would gravitate to having a `mapWithOptions` function that takes the `MapOptions` unconditionally and before the `...params`, and a `map` function that is just a wrapper that does not take `MapOptions` and passes the default value to `mapWithOptions`. It just seems to me that having an optional argument after a vararg list is too tricky, but maybe that is what people would prefer.

mbouaziz: I'll merge as is and we'll discuss it tomorrow

PR #313, opened 2024-09-06
Improve types and tweak some names for refresh tokens

bennostein:


skipruntime-ts/ts/src/skipruntime_service.ts
```diff
@@ -51,13 +52,14 @@ export type OutputDefinition = {
...
 
 export interface GenericSkipService {
-  // name / duration in milliseconds
-  tokens?: Record<string, number>;
   localInputs(): Record<string, InputDefinition>;
   remoteInputs(): Record<string, RemoteInputs>;
 
   outputs(): Record<string, OutputDefinition>;
 
+  // name / duration in milliseconds
```
jberdine, line 60: Both name and duration are present, right, not one or the other? If so:
```suggestion
  // name, duration in milliseconds
```

skipruntime-ts/ts/src/skipruntime_service.ts
```diff
@@ -66,14 +68,15 @@ export interface GenericSkipService {
...
   database: Database;
-  inputs: string[];
+  tables: string[];
 };
 
 export interface SimpleSkipService {
   inputTables?: string[];
   remoteTables?: Record<string, SimpleRemoteInputs>;
+
   // name / duration in milliseconds
```
jberdine, line 78: same / vs , here

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -16,6 +16,8 @@ export type TTableCollection = any;
 export type TTable = any;
 export type Param = any;
 
+export type RefreshToken = number & { __type_brand: "SkipRefreshToken" };
```
jberdine: The implementation of `freeze` does `Object.defineProperty(value, "__sk_frozen"` which seems to be playing a similar role. Maybe it is worthwhile to use just one mechanism in both places?

skipruntime-ts/ts/src/skipruntime_runner.ts
```diff
@@ -95,10 +96,10 @@ function toWriters(
 }
 
 class SimpleToGenericSkipService implements GenericSkipService {
-  tokens?: Record<string, number>;
+  refreshTokens?: Record<string, RefreshToken>;
```
jberdine, line 99: I don't understand the distinction between `Token`s and `RefreshToken`s. Each Token has a name and duration, and an array of them seems isomorphic to this `Record<string, RefreshToken>` which is like an association list right?

skipruntime-ts/ts/src/skipruntime_api.ts
mbouaziz: No, here, it's a hack for opaque types, it doesn't convey any value.

Feel free to use `Internal.Opaque<number, "SkipRefreshToken">`

bennostein: Right, the idea here is not to enforce immutability but to make RefreshToken an opaque type that one can't accidentally mix up with `number`.  I followed what seems like TS ecosystem standard of "branded" types but will switch to using Mehdi's Opaque type constructor instead if that achieves the same!

skipruntime-ts/ts/src/skipruntime_runner.ts
bennostein: `Token` here is an implementation detail of the `TimedQueue` which manages timing the token refreshes; users should only be dealing with `RefreshToken`.  I've now renamed and shuffled things around to make that more clear and avoid the name clash.

The types are indeed isomorphic but they have different interpretations: a `Token` (now `TQ_Token`) is an identifier paired with a _duration_ which the `TimedQueue` uses to schedule periodic refreshes. A `RefreshToken` on the other hand is a handle on the actual value being refreshed -- set to the current timestamp on refresh, though it should just be used as an opaque dependency in the computation graph, e.g. to trigger polls of external non-reactive dependencies.

PR #314, opened 2024-09-06
fix linter failure

jberdine:
to unbreak CI


jberdine:
@bennostein this is why https://github.com/SkipLabs/skdb/pull/313 failed CI I think

mbouaziz (approved):

PR #318, opened 2024-09-09
[skipruntime-ts] Add init function for local tables

skiplabsdaniel:


mbouaziz (approved):
What's the intended use case?

bennostein:
> What's the intended use case?

The idea is to give the ability to "hydrate" local state/inputs from a database or other service _before_ initializing the compute graph -- so that asynchronous process happens all at once and not piecemeal with the reactive compute churning as it tries to keep up.

bennostein (approved):

PR #319, opened 2024-09-09
[skipruntime-ts] Add take (limit)

mbouaziz:
Based on #321

### DIFF
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -511,6 +511,46 @@ fun sliced(
     .toString()
 }
 
+fun noFilter<K: Key, V: File>(
+  _context: readonly Context,
+  _key: K,
+  _value: File,
+): Bool {
+  true
+}
+
+fun countElementAsOne<V: File>(_value: File): Int {
+  1
+}
+
+fun removeElementAsOne<V: File>(_value: V, _toRemove: Int): File {
+  invariant_violation("removeElementAsOne called on singleton value");
+}
+
+@export("SkipRuntime_take")
+fun take(
+  context: mutable Context,
+  handleId: String,
+  name: String,
+  limit: UInt32,
+): String {
+  sourceDirName = DirName::create(handleId);
+  eager = EHandle(JSONID::keyType, JSONFile::type, sourceDirName);
+  dirName = sourceDirName.sub(name);
+  eager
+    .filter(
+      context,
+      dirName,
+      limit.toInt(),
+      noFilter,
+      countElementAsOne,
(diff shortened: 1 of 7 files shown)


skipruntime-ts/src/Extern.sk
```diff
@@ -487,6 +487,46 @@ fun sliced(
...
+  eager = EHandle(JSONID::keyType, JSONFile::type, sourceDirName);
+  dirName = sourceDirName.sub(name);
+  eager
+    .filter(
+      context,
+      dirName,
+      limit.toInt(),
+      noFilter,
+      countElementAsOne,
+      removeElementAsOne,
```
jberdine, line 547: You don't like lambdas? :-)

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -366,6 +366,11 @@ export interface EagerCollection<K extends TJSON, V extends TJSON> {
    */
   sliced(ranges: [K, K][]): EagerCollection<K, V>;
 
+  /**
+   * Create a new eager collection by keeping the given number of the first elements.
+   */
+  take(limit: int): EagerCollection<K, V>;
```
jberdine, line 365: Is `C.take(N)` supposed to take the first `N` values in `C`, or all the values of the first `N` keys in `C`? I ask since, from the code, it seems like the latter, but the docs read as the former to me.

skipruntime-ts/src/Extern.sk
mbouaziz: If I give a name to it, it saves a comment

skipruntime-ts/ts/src/skipruntime_api.ts
mbouaziz: Currently it's the latter indeed. Real usage will tell whether we need more flexibility

jberdine: Looking over this PR again, I don't know if it is clear what "first" means. As far as I understand `C.take(N)` will produce a collection with `N` elements if `C` already has `N` or more elements, and the elements included will be those with the smallest keys according to the order on the underlying Key type. But if one of those elements is then removed from `C`, will it be repaced in the result of take by a new one? Or will the result of take have fewer than `N` elements until more are added to `C`? I haven't dug through the implementation of filter in skstore deeply enough to answer these questions. Any insights @pikatchu ?

jberdine: Discussed offline, all good with "first" as it is now.

jberdine (approved):

PR #321, opened 2024-09-10
[skipruntime-ts] Sliced instead of MapOptions

mbouaziz:
As we discussed today, instead of the `MapOptions` thing introduced in #311 for ranges, we add a new function `sliced` on eager collections to do the actual slicing before mapping.
~~But it was actually simpler to implement a `sliced` that just prepares the ranges for the next `map`/`mapReduce`, saving the need to do the intermediate slicing before mapping.
Current limitation (that can be lifted later if there is a need): you cannot slice an already sliced collection. This is ensured by the type system.~~

### DIFF
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -216,7 +216,7 @@ class ExternAccumulator(
   }
 }
 
-fun asKeyRanges(jsonRanges: SKJSON.CJSON): ?Array<KeyRange> {
+fun asKeyRangesOpt(jsonRanges: SKJSON.CJSON): ?Array<KeyRange> {
   if (jsonRanges is SKJSON.CJNull()) return None();
   ranges = SKJSON.expectArray(jsonRanges);
   Some(
@@ -244,7 +244,7 @@ fun map(
     name,
     context,
     mapHandle.map,
-    asKeyRanges(rangeOpt),
+    asKeyRangesOpt(rangeOpt),
   )
     .getDirName()
     .toString()
@@ -271,7 +271,7 @@ fun mapReduce(
     context,
     mapHandle.map,
     ExternAccumulator(accumulatorHandle, JSONFile(default)),
-    asKeyRanges(rangeOpt),
+    asKeyRangesOpt(rangeOpt),
   )
     .getDirName()
     .toString()
@@ -427,7 +427,7 @@ fun mapTable(
     name,
     context,
     mapHandle.mapTable,
-    asKeyRanges(rangeOpt),
+    asKeyRangesOpt(rangeOpt),
   )
 }
 
@@ -448,7 +448,7 @@ fun toSkdb(
     context,
     mapHandle.toRow,
     connected != 0,
-    asKeyRanges(rangeOpt),
+    asKeyRangesOpt(rangeOpt),
   )
 }
 
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -454,6 +454,39 @@ fun toSkdb(
   )
 }
 
+fun identityMap<K: Key, V: File>(
+  _context: mutable Context,
+  writer: mutable TWriter<K, V>,
+  key: K,
+  it: mutable NonEmptyIterator<V>,
+): void {
+  writer.setArray(key, it.collect(Array));
+}
+
+@export("SkipRuntime_sliced")
+fun sliced(
+  context: mutable Context,
+  handleId: String,
+  name: String,
+  jsonRanges: SKJSON.CJArray,
+): String {
+  ranges = jsonRanges match {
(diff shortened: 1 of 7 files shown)


bennostein:
This makes sense, but I am not sure that it is actually _simpler_ from the user perspective -- we've introduced multiple variants of EagerCollection and TableCollection, with overlapping but not identical supported operations.

If I understand correctly, the reason for the no-slicing-a-sliced-collection restriction is to avoid computing the intersection of two sets of ranges, nothing fundamental, no problem.  But not being able to "get" without mapping is because there is no actual materialized sliced collection; the sliced collection just holds onto some ranges to be used whenever its mapped over.  Is that right?

I'd argue that this is premature optimization wrt what we talked about earlier, and that it is simpler for "slice" to apply a no-op mapper with the given ranges, producing a regular EagerCollection which supports all the usual operations.  If we want to avoid materializing those intermediate results, that seems a job for the abstract transducer-y representation that we tallked about earlier.

Am I understanding right? What do you think?

mbouaziz:
Right.

The implementation was easier because there was no need to write a no-op map and the rangesOpt argument was already there. I agree it clutters up the public API.
I'll rewrite as an eager slicing

mbouaziz:
Rewritten version. The hard part was getting a name for the slice

skipruntime-ts/ts/src/internals/skipruntime_module.ts
```diff
@@ -368,6 +384,14 @@ export class ContextImpl implements Context {
     ) as TJSON[];
   }
 
+  /* Must produce a valid SKStore key ideally with no collision */
+  keyOfJSON(value: TJSON): string {
```
bennostein, line 388: Pros and cons of generating names this way (b64-encoding input names and joining those) vs. how I did it for merge (stripping slashes off of input names and concatenating) -- these will end up more compact on the one hand but on the other hand less human-readable.  I don't feel too strongly either way but worth keeping in mind and potentially unifying the two in the future.

mbouaziz: I wouldn't have done differently if the name was generated on the skip side.
Currently SKStore directory names must be `'/'`-separated paths, each path element must be either an integer or start with a letter or underscore followed by alpha-numerical chars or `'-'` and `'_'`.
But my input is any valid JSON value.

mbouaziz: Wait... does the name really has to be unique?

bennostein (approved):
LGTM!  I'll rebase on here once you merge and reuse the same identity map function over on #322

skipruntime-ts/ts/src/internals/skipruntime_module.ts
mbouaziz: I'll merge as is so we move forward but let's simplify later if we can

bennostein: Well, the resulting DirName/handleID does need to be unique.  But what you have here achieves that, IIUC: the "name" passed to `SkipRuntime_slice` is used as a sub-directory name (at Utils.sk line 45) on the base eager handle, so the resulting DirName is unique for each base eager collection / ranges combination

mbouaziz: It looks conflict-free but, to be fully accurate and fully nit picking :stuck_out_tongue_winking_eye:, it isn't because I use for delimiters characters that can be found in values. We should probably allow more characters in dirnames. Or I could have encoded the length before each value.

jberdine: My understanding is that the underlying SKStore Context's dirs field maps DirNames to Dirs, and everything gets ahold of a Dir through that map. So if a Dir has multiple DirNames, that is ok though perhaps confusing, but no two distinct Dirs can have the same DirName or all hell breaks loose. Right? And currently the ts layer is getting uniqueness of DirNames by mostly ad hoc means. Would it make sense to have a gensym sort of operation on the context that would e.g. generate and suffix a number onto the desired name in order to ensure it is unique with respect to the dirs in the context? If we ever make a mistake and there is a name clash, it's going to be hell for us and users to debug.

skipruntime-ts/src/Extern.sk
```diff
@@ -448,10 +450,43 @@ fun toSkdb(
...
+  _context: mutable Context,
+  writer: mutable TWriter<K, V>,
+  key: K,
+  it: mutable NonEmptyIterator<V>,
+): void {
+  writer.setArray(key, it.collect(Array));
+}
+
+@export("SkipRuntime_sliced")
+fun sliced(
```
jberdine, line 467: Sorry to come late and nitpick names, but the other operation names like map and reduce are all verbs, while this one is a noun...

PR #322, opened 2024-09-10
[skipruntime-ts] Drop `multimap` from public API in favor of `merge`

bennostein:
This is a slightly less space efficient but much more understandable/clean interface to my eyes.

See b99cb86 for an example of the difference in user code.

### DIFF
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -279,6 +279,32 @@ fun mapReduce(
     .toString()
 }
 
+@export("SkipRuntime_union")
+fun union(
+  context: mutable Context,
+  name: String,
+  collections: SKJSON.CJArray,
+): String {
+  mappings = collections match {
+  | SKJSON.CJArray(v) ->
+    v.map(handleID -> {
+      ehandle = EHandle(
+        JSONID::keyType,
+        JSONFile::type,
+        DirName::create(SKJSON.asString(handleID)),
+      );
+      id_mapper = (_ctx, writer, key, vals) ~> {
+        //identity mapper function: write values straight through
+        writer.setArray(key, vals.collect(Array))
+      };
+      (ehandle, id_mapper)
+    })
+  };
+  eagerMultiMap(JSONID::keyType, JSONFile::type, name, context, mappings)
+    .getDirName()
+    .toString()
+}
+
 @export("SkipRuntime_multimap")
 fun multimap(
   context: mutable Context,
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -295,11 +295,7 @@ fun merge(context: mutable Context, collections: SKJSON.CJArray): String {
         JSONFile::type,
         DirName::create(handleID),
       );
-      id_mapper = (_ctx, writer, key, vals) ~> {
-        //identity mapper function: write values straight through
-        writer.setArray(key, vals.collect(Array))
-      };
-      (ehandle, id_mapper)
+      (ehandle, identityMap)
     })
   };
   eagerMultiMap(JSONID::keyType, JSONFile::type, name, context, mappings)
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -280,19 +280,19 @@ fun mapReduce(
 }
 
 @export("SkipRuntime_union")
-fun union(
-  context: mutable Context,
-  name: String,
-  collections: SKJSON.CJArray,
-): String {
+fun union(context: mutable Context, collections: SKJSON.CJArray): String {
+  name = "union";
   mappings = collections match {
   | SKJSON.CJArray(v) ->
-    v.map(handleID -> {
+    v.map(handleID_json -> {
+      handleID = SKJSON.asString(handleID_json);
       ehandle = EHandle(
         JSONID::keyType,
         JSONFile::type,
-        DirName::create(SKJSON.asString(handleID)),
+        DirName::create(handleID),
       );
+      invariant(handleID.endsWith("/"));
+      !name = name + handleID.splitLast("/").i0;
       id_mapper = (_ctx, writer, key, vals) ~> {
         //identity mapper function: write values straight through
         writer.setArray(key, vals.collect(Array))
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -279,9 +279,9 @@ fun mapReduce(
     .toString()
 }
 
-@export("SkipRuntime_union")
-fun union(context: mutable Context, collections: SKJSON.CJArray): String {
-  name = "union"; // to be concatenated with the names of inputs within the loop
+@export("SkipRuntime_merge")
+fun merge(context: mutable Context, collections: SKJSON.CJArray): String {
+  name = "merge"; // to be concatenated with the names of inputs within the loop
   mappings = collections match {
   | SKJSON.CJArray(v) ->
     v.map(handleID_json -> {
--- a/skipruntime-ts/src/Extern.sk
+++ b/skipruntime-ts/src/Extern.sk
@@ -281,18 +281,20 @@ fun mapReduce(
 
 @export("SkipRuntime_union")
 fun union(context: mutable Context, collections: SKJSON.CJArray): String {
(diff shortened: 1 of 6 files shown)


skipruntime-ts/src/Extern.sk
```diff
@@ -277,6 +277,32 @@ fun mapReduce(
...
+    v.map(handleID -> {
+      ehandle = EHandle(
+        JSONID::keyType,
+        JSONFile::type,
+        DirName::create(SKJSON.asString(handleID)),
+      );
+      id_mapper = (_ctx, writer, key, vals) ~> {
+        //identity mapper function: write values straight through
+        writer.setArray(key, vals.collect(Array))
+      };
```
mbouaziz: Let's share our identity mappers (see #321)

skipruntime-ts/ts/src/internals/skipruntime_types.ts
```diff
@@ -326,6 +324,18 @@ export interface FromWasm {
     value: ptr<Internal.CJObject>,
   ): number;
 
+  SkipRuntime_union(
+    ctx: ptr<Internal.Context>,
+    name: ptr<Internal.String>,
+    collections: ptr<Internal.CJArray>,
```
mbouaziz: You can even specify `Internal.CJArray<Internal.CJString>`

skipruntime-ts/ts/src/internals/skipruntime_types.ts
```diff
@@ -112,6 +112,11 @@ export interface Context {
     rangeOpt?: [K, K][] | null,
   ) => void;
 
+  union: <K extends TJSON, V extends TJSON>(
+    name: string,
+    ...collections: EagerCollection<K, V>[]
```
mbouaziz: I don't have a strong opinion on this, I'm still not sure of why we have so many layers, but it looks like at this level, we're already in the world of strings and not collections anymore (the `import type EagerCollection` triggered me)

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
```diff
@@ -414,6 +414,16 @@ export class SKStoreImpl implements SKStore {
     });
   }
 
+  union<K extends TJSON, V extends TJSON>(
+    ...collections: EagerCollection<K, V>[]
+  ) {
+    if (collections.length < 2)
+      throw new Error("Union requires at least two input collections.");
```
mbouaziz: This could be "enforced" with types

mbouaziz: Fixed in later version

skipruntime-ts/src/Extern.sk
```diff
@@ -277,6 +277,32 @@ fun mapReduce(
     .toString()
 }
 
+@export("SkipRuntime_union")
+fun union(context: mutable Context, collections: SKJSON.CJArray): String {
+  name = "union";
```
jberdine: Just to check, it is ok for all the unions have the same name?

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -355,42 +355,6 @@ export interface EagerCollection<K extends TJSON, V extends TJSON> {
...
   ): EagerCollection<K2, V3>;
 
-  /**
-   * Create a new eager reactive collection by mapping some computation `mapper` over this
-   * one and then reducing the results with `accumulator` (with options)
-   * @param {Mapper} mapper - function to apply to each element of this collection
-   * @param {Accumulator} accumulator - function to combine results of the `mapper`
-   * @returns {EagerCollection} An eager collection containing the output of the accumulator
-   */
-  mapReduce<
```
jberdine: Is removing mapReduce and mapTo intended?

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -603,39 +580,6 @@ export interface SKStore {
...
-   * @param {Mapping[]} mappings - the input collections and corresponding mappers
-   * @returns {EagerCollection} An eager collection containing the combined outputs of the `mappings`
-   */
-  multimap<
-    K1 extends TJSON,
-    V1 extends TJSON,
-    K2 extends TJSON,
-    V2 extends TJSON,
-  >(
-    mappings: Mapping<K1, V1, K2, V2>[],
```
jberdine, line 569: I think this and the one below are the only uses of the `Mapping` type and it could be removed too.

skipruntime-ts/ts/src/internals/skipruntime_types.ts
pikatchu: What is the name for? I thought those where generated for the user in the TS layer.

bennostein: This is from a previous version -- the current diff generates the name on the skip side.

skipruntime-ts/src/Extern.sk
bennostein: Hey no fair, I wrote mine first ;)

Sounds good, I'll rebase and do that once yours lands in trunk

jberdine (commented):
I think the `union` API makes sense, though the term `union` makes me think of sets, and expect that duplicate values for the same key will be deduped, but AFAIU they are not. I'm not sure whether that expectation is just me though, and which behavior is likely to be wanted by users.

Another thing that occurs to me is whether it would make sense to have an operation like `union` here, but that takes an `Accumulator` and gives all the values in all the combined collections to it in order to produce the output value. Then both deduping `union` and nondeduping `join` operations could be defined as wrappers using simple accumulators. This is more what I had been thinking before regarding a `merge` operation.

skipruntime-ts/ts/src/skipruntime_api.ts
bennostein: Ahh, nope, messed up rebasing too quickly, got caught out by there now being multiple overloaded versions of mapReduce/mapTo

skipruntime-ts/src/Extern.sk
bennostein: They don't -- this gets concatenated with input collections' names within the loop a few lines down.  Will make a note in comment.

bennostein (commented):
> I think the `union` API makes sense, though the term `union` makes me think of sets, and expect that duplicate values for the same key will be deduped, but AFAIU they are not. I'm not sure whether that expectation is just me though, and which behavior is likely to be wanted by users.

I'll rename to `merge` to avoid that confusion.  I agree with your reasoning here, set union isn't a perfect analogy.

> Another thing that occurs to me is whether it would make sense to have an operation like `union` here, but that takes an `Accumulator` and gives all the values in all the combined collections to it in order to produce the output value. Then both deduping `union` and nondeduping `join` operations could be defined as wrappers using simple accumulators. This is more what I had been thinking before regarding a `merge` operation.

Good idea!  I'll leave for a future PR to avoid continuing to add here but I think it would make sense (and be pretty easy) to add such an operation, possibly just as an optional parameter on this function.

jberdine:
> I'll rename to `merge` to avoid that confusion. I agree with your reasoning here, set union isn't a perfect analogy.

> Good idea! I'll leave for a future PR to avoid continuing to add here but I think it would make sense (and be pretty easy) to add such an operation, possibly just as an optional parameter on this function.

Sounds good to use a separate PR, but I'd think about eventual naming now. In APIs I've used, `merge` tends to be the operation with a user-specified resolution, while things like `union`, `join`, `concat` don't have such flexibility. So I'd say the current `union` is better named `join` or `concat`. But if the accumulator-accepting one gets called `reduce`, say, than using `merge` for `union` seems fine to me.

I would hesitate to go the optional arg route since the accumulator-accepting one is definitely more general and union is a special case, but more importantly, the type of the output values could be different than the input ones in the accumulator-accepting case, while union will need to fix the output type to be the same as the input.

jberdine:
> Another thing that occurs to me is whether it would make sense to have an operation like `union` here, but that takes an `Accumulator` and gives all the values in all the combined collections to it in order to produce the output value. Then both deduping `union` and nondeduping `join` operations could be defined as wrappers using simple accumulators. This is more what I had been thinking before regarding a `merge` operation.

Thinking more, maybe taking literally an `Accumulator` is not a good idea, since here it is only used when the keys are the same, and there will be no need for the `dismiss` function.

PR #323, opened 2024-09-12
[docs] Skeleton layout based on Charles' proposal

jberdine:


jberdine:
image
image

bennostein (approved):
Structure etc. look good to me!

Sidenote: specifying table-of-contents order by putting a `sidebar_position` index into each sub-page's metadata seems crazy!  Any idea why docusaurus does it that way instead of all in one place?

jberdine:
Yeah, not a fan of the index in each file approach myself. I'm not sure it is the only way, but haven't dug into it so far. There was some mention that _by default_ the sidebar is generated from the filesystem, making me think there is another way.

PR #325, opened 2024-09-12
Explicitly define sidebar instead of autogenerating

bennostein:
This is just generally nicer, but also will let us add some structure/hierarchy in the sidebar as we add content.

See: https://docusaurus.io/docs/sidebar/items


bennostein:
Oh, and I should also mention that I reordered things according to my proposal on discord.  Feel free to shuffle around further of course

jberdine:
Does the "Skip API" entry still appear in the sidebar?

jberdine (approved):
LGTM. I pushed a commit that re-adds the api to the sidebar.

PR #328, opened 2024-09-13
[skipruntime-ts] A few cleanups on the API

mbouaziz:
- remove unused types
- unified handling of SK-frozen objects
- more precise type for `Param`

### DIFF
--- a/skipruntime-ts/ts/src/internals/skipruntime_impl.ts
+++ b/skipruntime-ts/ts/src/internals/skipruntime_impl.ts
@@ -110,9 +110,9 @@ class EagerCollectionImpl<K extends TJSON, V extends TJSON>
     return this.context.maybeGetOne(this.eagerHdl, key);
   }
 
-  size = () => {
+  size(): int {
     return this.context.size(this.eagerHdl);
-  };
+  }
 
   sliced(ranges: readonly [K, K][]): EagerCollection<K, V> {
     const sliceName =
@@ -381,7 +381,8 @@ export class TableImpl<R extends TJSON[]> implements Table<R> {
       query.params ? toParams(query.params) : undefined,
     );
   }
-  watch = (update: (rows: JSONObject[]) => void, feedback: boolean = false) => {
+
+  watch(update: (rows: JSONObject[]) => void, feedback: boolean = false) {
     const name = feedback
       ? this.getName() + serverResponseSuffix
       : this.getName();
@@ -391,13 +392,13 @@ export class TableImpl<R extends TJSON[]> implements Table<R> {
       query.params ? toParams(query.params) : {},
       update,
     );
-  };
+  }
 
-  watchChanges = (
+  watchChanges(
     init: (rows: JSONObject[]) => void,
     update: (added: JSONObject[], removed: JSONObject[]) => void,
     feedback: boolean = false,
-  ) => {
+  ) {
     const name = feedback
       ? this.getName() + serverResponseSuffix
       : this.getName();
@@ -408,7 +409,7 @@ export class TableImpl<R extends TJSON[]> implements Table<R> {
       init,
       update,
     );
-  };
+  }
 }
 
 export class SKStoreImpl extends SkFrozen implements SKStore {
@@ -481,9 +482,11 @@ export class SKStoreFactoryImpl implements SKStoreFactory {
     private createKey: (key: string) => Promise<CryptoKey>,
   ) {}
 
-  getName = () => "SKStore";
+  getName() {
+    return "SKStore";
+  }
 
-  runSKStore = async (
+  async runSKStore(
     init: (
       skstore: SKStore,
       tables: Record<string, TableCollection<TJSON[]>>,
@@ -492,7 +495,7 @@ export class SKStoreFactoryImpl implements SKStoreFactory {
     remotes: Record<string, Remote> = {},
     tokens: Record<string, number> = {},
     initLocals?: (tables: Record<string, Table<TJSON[]>>) => Promise<void>,
-  ): Promise<Record<string, Table<TJSON[]>>> => {
+  ): Promise<Record<string, Table<TJSON[]>>> {
     const context = this.context();
     const tables = await mirror(
       context,
@@ -519,7 +522,7 @@ export class SKStoreFactoryImpl implements SKStoreFactory {
       result[key] = (value as TableCollectionImpl<TJSON[]>).toTable();
     }
     return result;
-  };
+  }
 }
 
 function toWs(entryPoint: EntryPoint) {
--- a/skipruntime-ts/ts/src/internals/skipruntime_impl.ts
+++ b/skipruntime-ts/ts/src/internals/skipruntime_impl.ts
@@ -860,3 +860,60 @@ export function check<T>(value: T): void {
     throw new Error("'" + type + "' cannot be manage by skstore.");
   }
 }
+
+/**
+ * _Deep-freeze_ an object, returning the same object that was passed in.
+ *
+ * This function is similar to
+ * {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze | `Object.freeze()`}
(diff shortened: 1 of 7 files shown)


skipruntime-ts/ts/src/internals/skipruntime_impl.ts
```diff
@@ -877,27 +877,23 @@ export function freeze<T>(value: T): T {
...
+      Object.defineProperty(frozenValue, "__sk_frozen", {
         enumerable: false,
         writable: false,
         value: true,
       });
-      for (const key of Object.keys(jso)) {
-        jso[key] = freeze(jso[key]);
-      }
-      return Object.freeze(jso) as T;
+      return Object.freeze(frozenValue) as T;
```
jberdine: The commit message isn't right, this isn't just a "simplification" since it (significantly) changes the semantics. Am I confused or doesn't this change `freeze` from an in-place mutation operation like `Object.freeze` to an operation that makes a deep copy that is frozen, leaving the argument value unchanged?

This semantic change is something that I have discussed with Julien, and agree it should be made. I would put such semantic changes in separate PRs though. And I think that `freeze` is not a good name for the new operation that makes a deep copy. Perhaps rename to e.g. `frozenClone`. I'll need to update the docs in https://github.com/SkipLabs/skdb/pull/326.

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
```diff
@@ -47,21 +47,47 @@ function assertNoKeysNaN<K extends TJSON, V extends TJSON>(
...
+    enumerable: false,
+    writable: false,
+    value: true,
+  }) as T & Frozen;
+}
+
+function isSkFrozen(x: any): x is Frozen {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  return sk_frozen in x && x[sk_frozen];
+}
```
jberdine: A way to help the poor type checker:
```suggestion
function isSkFrozen(x: any): x is Frozen {
  let is_frozen = false;
  if (sk_frozen in x) {
    is_frozen = x[sk_frozen];
  };
  return is_frozen;
}
```

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
```diff
@@ -47,21 +47,47 @@ function assertNoKeysNaN<K extends TJSON, V extends TJSON>(
...
+
+abstract class SkFrozen implements Frozen {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore error TS2564: Property '[sk_frozen]' has no initializer and is not definitely assigned in the constructor.
+  [sk_frozen]: true;
```
jberdine:
```suggestion
  // tsc misses that Object.defineProperty in the constructor inits this
  [sk_frozen]!: true;
```

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
```diff
@@ -47,21 +47,47 @@ function assertNoKeysNaN<K extends TJSON, V extends TJSON>(
...
+}
+
+function sk_freeze<T extends object>(x: T): T & Frozen {
+  return Object.defineProperty(x, sk_frozen, {
+    enumerable: false,
+    writable: false,
+    value: true,
+  }) as T & Frozen;
+}
```
jberdine: I don't understand why we don't also call `Object.freeze` here. The call sites in the definition of `freeze` include a call to `Object.freeze`, but I don't see why the call sites of the `SkFrozen` constructor should not also call `Object.freeze`. In which case better to just do it here. I suppose that the superconstructor call would need to be moved after the subclass field initializations, but that would be ok I expect.

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
```diff
@@ -47,21 +47,47 @@ function assertNoKeysNaN<K extends TJSON, V extends TJSON>(
 }
 export const serverResponseSuffix = "__skdb_mirror_feedback";
 
+export const sk_frozen: unique symbol = Symbol();
+
+export interface Frozen {
```
jberdine: To avoid potential confusion from people associating "frozen" with what `Object.freeze` achieves, maybe this should be renamed to e.g. `DeepFrozen` or `Constant`. For reference, the docs for `Object.freeze` state that it results in an _immutable_ but not necessarily _constant_ object. So `Constant` would be keeping with that terminology.

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -13,7 +13,14 @@ export type JSONObject = { [key: string]: TJSON | null };
...
 
-export type Param = any;
+export type Param =
+  | null
+  | boolean
+  | number
+  | string
+  | Frozen
+  | readonly Param[]
+  | { readonly [k: string]: Param };
```
jberdine, line 23: Do we want to allow arrays and objects that are not themselves `Frozen` but which might eventually contain `Frozen` data? It would be more conservative to, at least initially, require arrays and objects to be `Frozen`. I don't trust `readonly` in TS to achieve anything. Also, IIUC the `check` function will blow at runtime on non-frozen arrays and objects.

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
mbouaziz: Not a simplification, indeed. I'll remove the commit from this PR.
Don't we may want both, though?

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
mbouaziz: Thanks! `return sk_frozen in x && x[sk_frozen] === true;` works as well

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
mbouaziz: <3

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
mbouaziz: Typescript expects the superconstructor calls to be first, otherwise I get
```
A 'super' call must be the first statement in the constructor to refer to 'super' or 'this' when a derived class contains initialized properties, parameter properties, or private identifiers.ts(2376)
```

I think we don't call `Object.freeze` on our classes because we think we know what is going on. Though anyone can add properties to them if we don't so that'd be better. I'll add a commit that does that

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
mbouaziz: Changing for `Constant`

skipruntime-ts/ts/src/skipruntime_api.ts
mbouaziz: The goal was not to make `check` always pass but start with reducing the cases where it would blow.

Currently `check` accepts more things than values returned by `freeze`, it accepts values that are manually deeply frozen.

jberdine: Ah yes, check will accept some manually deep-frozen values. But it is slow in that case as it has to do a deep traversal every time. I seem to recall, but can't currently find, some discussion where check accepting in these cases was an intermediate state but not what we want eventually. Do you recall @skiplabsdaniel ?

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
jberdine: I also thought that maybe we would want both, but discussing with @pikatchu settled on going with a copying version only initially, as the advantage of the non-copying one is only to save one copy and the usability drawbacks of deeply freezing users objects in-place isn't worth it. We can always add an in-place operation later if it becomes a performance issue for someone, but better to initially steer people to the friendlier interface.

PR #329, opened 2024-09-13
Cookbook: database example

pikatchu:
Reactive service using a database to backup one of its tables. We use SQLite in the example, but we could have chosen anything.


skipruntime-ts/ts/examples/database.ts
```diff
@@ -0,0 +1,156 @@
...
+/*****************************************************************************/
+// The read path, we want to find a user
+/*****************************************************************************/
+
+type User = {
+  name: string;
+  country: string;
+};
+
+type Response = { status: "error"; msg: string } | { status: "ok"; user: User };
```
jberdine, line 86: This makes me realize an issue with `TJSON`: as written `Request` is not a subtype of `TJSON` since there is nothing that says that a union of TJSONs is a TJSON. Not sure how to react yet.

skipruntime-ts/ts/examples/database.ts
```diff
@@ -0,0 +1,156 @@
...
+    );
+    console.log("Initializing done.");
+    console.log("Service is ready!");
+  }
+
+  reactiveCompute(
+    _store: SKStore,
+    requests: EagerCollection<string, TJSON>,
+    inputCollections: Record<string, EagerCollection<string, TJSON>>,
+  ): SimpleServiceOutput {
```
jberdine, line 165: It seems that `SimpleServiceOutput` should be parametric. As it is, we cannot give a better/tighter type to `update`.

jberdine (commented):
I don't know if it is my setup, but when I run this what I observe is:
```
$ cd skipruntime-rs
$ make bunrun-database
...
Initializing done.
Service is ready!
```
and in another shell
```
$ make bunclient-database
cd ts/examples && bun run database-client.ts
> {"type":"init","payload":[]}
play 1
>> {"type":"get","command":"getUser","payload":"123"}
>> {"type":"post","command":"set","payload":[{"key":"123","value":{"name":"daniel","country":"UK"}}]}
> {"type":"update","payload":{"added":[{"id":"3rMcURdTnxcaE7kapXhfcwt-BA4","response":"{\"status\":\"ok\",\"user\":{\"country\":\"FR\",\"name\":\"daniel\"}}","skdb_author":"","skdb_access":"read-write"}],"removed":[]}}
make: *** [Makefile:83: bunclient-database] Error 2
```
Note that while the output looks as expected, it is exiting with failure. Is that intentional? Also, the last line for "update" is sometimes displayed and sometimes not displayed. I don't know what is going on there at the moment.

jberdine:
@pikatchu I pushed some commits to your branch instead of making a bunch of little code suggestions. Feedback on whether or not they are a good idea welcome.

skipruntime-ts/ts/examples/database.ts
```diff
@@ -0,0 +1,173 @@
...
+    const cmd = it.first();
+    const value = this.users.maybeGetOne(cmd.payload as GetUser);
+    const user = value as User;
+    const computed: Response =
+      cmd.command == "getUser"
+        ? { status: "ok", user }
+        : { status: "error", msg: "Unknown command" };
+    return Array([key, computed]);
+  }
+}
```
bennostein, line 104: Bit simpler with ValueMapper, since you're not modifying key structure at all here.

```suggestion
class Request implements ValueMapper<string, Command, Response> {
  constructor(private users: EagerCollection<string, TJSON>) {}

  mapValue(cmd: Command): Response {
    const value = this.users.maybeGetOne(cmd.payload as GetUser);
    const user = value as User;
    return cmd.command == "getUser"
        ? { status: "ok", user }
        : { status: "error", msg: "Unknown command" };
  }
}
```

skipruntime-ts/ts/examples/database.ts
jberdine: Update: After more testing it seems that for better or worse tsc does the structural expansion while checking subtyping, which lets Response <: TJSON go through.

PR #330, opened 2024-09-16
Some docs & minor fixes for refresh tokens

bennostein:
Not getting into a full worked example for the docs until seeing the external call API, but wrote some high-level overview text and fixed up some comments/identifiers/etc. in the code after our meeting on friday clarified some things for me.

### DIFF
--- a/www/docs/sources_destinations.md
+++ b/www/docs/sources_destinations.md
@@ -14,7 +14,7 @@ are kept as fresh as your application requires.
 ## Overview
 
 A reactive service can define any number of refresh tokens, each of which
-associates a string identifier with a refresh frequency, given in milliseconds.
+associates a string identifier with a refresh interval, given in milliseconds.
 Skip then maintains an internal data structure with a timestamp for each token
 that is refreshed at the specified frequency.
 
--- a/www/docs/sources_destinations.md
+++ b/www/docs/sources_destinations.md
@@ -1,9 +1,32 @@
 # Data sources and destinations
 
+Skip is designed to carefully track the computation and data dependencies of
+your application, so that outputs can be efficiently updated if (and only if!)
+any inputs they depend upon change.  However, when those inputs are from
+non-Skip sources -- databases, external systems or APIs, non-reactive internal
+systems, and the like -- it is crucial to ensure that your reactive service
+still produces up-to-date results.
+
+Skip provides a simple and easy-to-use mechanism to address this situation: by
+defining and using _refresh tokens_, you can ensure that any non-reactive inputs
+are kept as fresh as your application requires.
+
 ## Overview
+
+A reactive service can define any number of refresh tokens, each of which
+associates a string identifier with a refresh frequency, given in milliseconds.
+Skip then maintains an internal data structure with a timestamp for each token
+that is refreshed at the specified frequency.
+
+When a Skip computation calls `SKStore.getRefreshToken` to get the latest
+timestamp for some refresh token, it incurs a dependency on that token.  Thus,
+it will be re-executed periodically and can produce outputs or intermediate data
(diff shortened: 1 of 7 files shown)


www/docs/sources_destinations.md
```diff
@@ -1,9 +1,32 @@
 # Data sources and destinations
 
+Skip is designed to carefully track the computation and data dependencies of
+your application, so that outputs can be efficiently updated if (and only if!)
+any inputs they depend upon change.  However, when those inputs are from
```
jberdine:
```suggestion
any inputs they depend on change.  However, when those inputs are from
```
since "upon" is more formal than "on" and the rest of our prose is not super formal, but your call

www/docs/sources_destinations.md
```diff
@@ -1,9 +1,32 @@
...
+
 ## Overview
 
+A reactive service can define any number of refresh tokens, each of which
+associates a string identifier with a refresh frequency, given in milliseconds.
+Skip then maintains an internal data structure with a timestamp for each token
+that is refreshed at the specified frequency.
+
+When a Skip computation calls `SKStore.getRefreshToken` to get the latest
+timestamp for some refresh token, it incurs a dependency on that token.  Thus,
```
jberdine: It isn't clear at this point, to me at least, why someone would want "the latest timestamp"? Is making that call expected to be done only in order to establish the dependency?

jberdine (approved):

www/docs/sources_destinations.md
```diff
@@ -19,9 +19,11 @@ Skip then maintains an internal data structure with a timestamp for each token
...
 
 When a Skip computation calls `SKStore.getRefreshToken` to get the latest
-timestamp for some refresh token, it incurs a dependency on that token.  Thus,
-it will be re-executed periodically and can produce outputs or intermediate data
-structures which depend on the latest timestamp.
+timestamp for some refresh token, it incurs a dependency on that token.
+Thus, it will be re-executed periodically, enabling you to poll external
+services.  By using the token's timestamp value, you can also reactively
+maintain any outputs or intermediate data structures which depend on the latest
+"current" timestamp.
```
jberdine, line 26: Thanks for revising, but I'm still confused, sorry. Reading and thinking about this, I wonder if there is a confusion about the word "timestamp" itself. AFAIK the term "timestamp" means (a record of) the time at which something happened. But here you seem to use it to refer to some other value, not just a time. Or maybe it is just a time value, and I'm not understanding how such a thing gets "used" to produce an output. I mean, the output usually isn't going to actually be a function of the time value, it just needs to get re-evaluated with some frequency, right? Maybe `getRefreshToken` should be named something more like `subscribeToRefreshToken` to indicate that just calling the function in `mapElement` and immediately throwing away the result has the effect of causing later re-evaluations of the `mapElement` and later functions.

bennostein: Right, by "timestamp" here I mean "the integer number of seconds since the unix epoch".  If you're doing something lazy that is a function of the time value, you can use it as is.  I've got a commit in my current working branch which also exposes refresh tokens as a singleton collection, i.e.
`getRefreshTokenEager: (key: string) => EagerCollection<string, RefreshToken>` (along with the current `getRefreshToken : (key:string) => RefreshToken`)
which makes things simpler, I think, for cases where you really do want to set up some eager computation that is a function of the actual time.  (motivated by the oncall rotation example which I've picked back up).

Let's chat more over VC later.  As I see it there are two core use cases for refresh tokens: creating dependencies (without using the actual value) to schedule polls/recomputations, and setting up reactive computations that care about the "current" time. We've focused on the former to this point, but it's only minor additions needed to cover the latter.

PR #332, opened 2024-09-18
Simplifications and improvements on database example

jberdine:


skipruntime-ts/ts/src/package.json
```diff
@@ -1,15 +1,10 @@
...
-    "@eslint/js": "^9.9.1",
-    "@stylistic/eslint-plugin-js": "^2.6.4",
-    "@types/node": "^22.5.5",
-    "@typescript-eslint/parser": "^8.3.0",
-    "eslint-plugin-jsdoc": "^50.2.2",
-    "prettier": "3.2.5",
-    "typescript-eslint": "^8.3.0"
+    "@eslint/js": "",
+    "@types/node": ""
   }
```
mbouaziz: I think you can completely remove the `devDependencies` from modules if they are defined at the workspace root

package.json
```diff
@@ -1,12 +1,12 @@
...
   "type": "module",
   "devDependencies": {
-    "@eslint/js": "^9.9.1",
-    "@stylistic/eslint-plugin-js": "^2.6.4",
-    "@types/node": "^22.5.5",
-    "@typescript-eslint/parser": "^8.3.0",
-    "eslint-plugin-jsdoc": "^50.2.2",
-    "prettier": "3.2.5",
-    "typescript-eslint": "^8.3.0"
+    "@eslint/js": "",
```
mbouaziz: About this commit: yes we haven't carefully analyzed compatibility constraints but we know things work with the given versions.
If you do not specify versions, there is a risk that an update of a random package breaks the CI unrelatedly to the PR

skipruntime-ts/ts/examples/database.ts
```diff
@@ -100,32 +82,35 @@ async function update(
...
-      for (const entry of cmd.payload) {
-        await db.run(
-          "INSERT OR REPLACE INTO data (id, object) VALUES ($id, $object)",
-          {
-            $id: entry.key,
-            $object: JSON.stringify(entry.value),
-          },
-        );
-        writer.set(entry.key, entry.value);
+  await new Promise((resolve) => {
```
mbouaziz: Do we really need this? I think we don't

skipruntime-ts/ts/examples/utils.ts
```diff
@@ -77,7 +77,7 @@ class Player {
 
   play(idx?: number) {
     let run = true;
-    if (idx ?? 0 > 0) {
+    if ((idx ?? 0) > 0) {
```
mbouaziz: or use `play(idx = 0) {`

skipruntime-ts/ts/src/package.json
jberdine: I can remove eslint/js, but something re-adds types/node by itself.

package.json
jberdine: Ok, I'll leave the versions in. I am bothered by how we have a bunch of package.json files with differing versions for the same dependencies strewn aroung the repo.

skipruntime-ts/ts/examples/database.ts
jberdine: I don't know. On one hand, the code before this PR is just type-incorrect as the function returns `void` and should return `Promise<void>`. And, not that I trust it, but the linter complains. On the other hand, returning a promise that immediately resolves seems superfluous. Initially I just wanted to use the `return` operation of the `Promise` monad, but there doesn't seem to be one in JS. Would putting `await new Promise((resolve) => resolve(undefined))` at the end instead of putting the body of this code in the resolve continuation make you happier?

jberdine: The other changes in this commit are the ones that fix the failures and odd behavior of the test that I commented about earlier (cc @pikatchu), but FTR if I omit the `await new Promise` parts, I still cannot reproduce the test failures.

jberdine: My personal view is that I would rather go with one of the type-correct options, even if we cannot observe a runtime failure with the option that has less code. Also, in this case the linter was able to detect a real issue that the type-checker failed to notice, so I am inclined to leave it enabled, which also supports going with one of the type-correct options.

mbouaziz: I think we should change the return type of `update` in (`Simple`)`ServiceOutput` to `void | Promise<void>`. It is OK to `await` a non-`Promise`, it's a no-op.

jberdine: Like https://github.com/SkipLabs/skdb/pull/333?

skipruntime-ts/ts/src/package.json
```diff
@@ -1,15 +1,9 @@
...
   },
   "devDependencies": {
-    "@eslint/js": "^9.9.1",
-    "@stylistic/eslint-plugin-js": "^2.6.4",
-    "@types/node": "^22.5.5",
-    "@typescript-eslint/parser": "^8.3.0",
-    "eslint-plugin-jsdoc": "^50.2.2",
-    "prettier": "3.2.5",
-    "typescript-eslint": "^8.3.0"
+    "@types/node": "^22.5.5"
```
mbouaziz: I don't want this comment to be blocking but I'm wondering whether we should really remove typing dependencies from this `package.json`. Won't they be needed by our users once the package is published?

My ChatGPT-helped understanding is that we should keep in `dependencies` packages that contain types that we expose. So let's keep things in `devDependencies` for now.

skipruntime-ts/ts/examples/package.json
```diff
@@ -1,5 +1,8 @@
 {
   "type": "module",
+  "scripts": {
+    "lint": "eslint --config ../../../eslint.config.js ."
```
mbouaziz: Note:
I intentionally did not enable eslint and very strict tsc rules on tests and examples because we probably won't expect it from our users.
I.e. I don't want to "cheat" when writing the examples.
Not sure it's very reasonable though

skipruntime-ts/ts/examples/database.ts
mbouaziz: Do we still need the `await new Promise ...` in this commit?

skipruntime-ts/ts/examples/utils.ts
```diff
@@ -66,7 +66,7 @@ class Player {
 
   start(idx: number) {
     const scenario = this.scenarios[idx - 1];
-    if (!scenario) {
+    if (scenario.length == 0) {
```
mbouaziz: Hmm, not sure about this change. Don't we want to check `scenario !== undefined`?

skipruntime-ts/ts/examples/utils.ts
```diff
@@ -75,10 +75,10 @@ class Player {
     }
   }
 
-  play(idx?: number) {
+  play(idx = 0) {
     let run = true;
-    if (idx ?? 0 > 0) {
-      run = this.start(idx!);
+    if (idx > 0) {
+      run = this.start(idx);
```
mbouaziz, line 81: Nit: reading a bit farther, I think we should use `number | undefined` or `number | null` rather than this 1-based index. But not important => skip

skipruntime-ts/ts/examples/sum.ts
```diff
@@ -37,11 +38,12 @@ class Add implements Mapper<string, TJSON, string, TJSON> {
     it: NonEmptyIterator<TJSON>,
   ): Iterable<[string, TJSON]> {
     const v = it.first() as number;
-    const ev = this.other.maybeGetOne(key) as number;
+    const ev = this.other.maybeGetOne(key) as Opt<number>;
```
mbouaziz: You can definitely use tighter types: replace the `TJSON` in this class by `number`, and `inputCollections` in `reactiveCompute` by `inputCollections: Record<string, EagerCollection<string, number>>,` or even `: { input1: EagerCollection<string, number>, input2: ... }`, it will make `tsc` (unsoundly) happy.

skipruntime-ts/ts/src/package.json
jberdine: Do you mean that perhaps `@types/node` should be in `dependencies` instead of `devDependencies`, or that it would be removed entirely? I tried to remove it entirely, but something run by `make check-all` adds it back.

skipruntime-ts/ts/examples/package.json
jberdine: Hmm, I was thinking that the example code should be a demonstration of what we want to encourage users to do, and so making TS as close to a reasonable language as possible there as well seems like the way to go. I don't know though. I just found myself duplicating the config and so deduped it.

skipruntime-ts/ts/examples/utils.ts
jberdine: Although the linter thinks that `scenario !== undefined` is always false.

skipruntime-ts/ts/examples/utils.ts
jberdine: Not disagreeing, but curious why you think that.

skipruntime-ts/ts/src/package.json
mbouaziz: `@types/node` may be a special case, but I don't think what I wrote applies to any of our `devDependencies` yet. We don't re-expose types from our dependencies

skipruntime-ts/ts/examples/utils.ts
mbouaziz: That's because we disabled `noUncheckedIndexedAccess` in `tsconfig.json`. It was too noisy. Ideally we'll re-enable it

skipruntime-ts/ts/src/package.json
jberdine: Ok, so to summarize: If we re-expose any types of a dependency, we should put it in `dependencies`, and if we only need them for building/linting/etc, we should put them in `devDependencies`.

skipruntime-ts/ts/examples/utils.ts
mbouaziz: 0 is used to mean something else than an index, so we can use another type. `"kamoulox"` works as well

skipruntime-ts/ts/examples/utils.ts
jberdine: So since we don't want `noUncheckedIndexedAccess` to be reported everywhere, it complains if we ever do check the index access. FFS.

skipruntime-ts/ts/examples/database.ts
jberdine: I will squash the commits that intro and elim `await new Promise` and get the ones in sum and sheet too.

skipruntime-ts/ts/examples/utils.ts
jberdine: Fixed now by using different type-checking options when linting

skipruntime-ts/ts/src/package.json
jberdine: In the current state of the tip of this PR I have one remaining question about this: the examples/package.json has non-dev deps on sqlite3 and ws since it actually uses them at runtime, and also a dep on @types/sqlite3 which I'm unsure whether it should stay a non-dev dep of move to a devDep.

jberdine:
In the latest update I have given up on trying to deduplicate the linting and type-checking configurations and have gone back to just diffing the files manually to see if they are in sync. Since the linter requires different type-checking options than the normal build, there are file names in the eslint config file. It should be possible, but I have failed to get the supposed support for making one eslint config file extend another to work. I timed out and went the cave-man route.

skipruntime-ts/ts/src/package.json
mbouaziz: It matters only for packages that we are going to publish.
A dev dependency will not be installed by the users, whereas a normal dependency will.

mbouaziz (approved):
:rocket:

PR #333, opened 2024-09-20
Allow SkipService init and update functions to return void

jberdine:
in addition to Promise<void>.

### DIFF
--- a/skipruntime-ts/ts/src/internals/skipruntime_process.ts
+++ b/skipruntime-ts/ts/src/internals/skipruntime_process.ts
@@ -55,9 +55,12 @@ export async function runWithServer_(
           values: [[http, "", "read-write"]],
         });
       } else if (http.type == "post") {
-        update(http, {}).catch((e: unknown) => {
-          throw e;
-        });
+        const maybePromise = update(http, {});
+        if (typeof maybePromise === "object") {
+          maybePromise.catch((e: unknown) => {
+            throw e;
+          });
+        }
       } else {
         throw new TypeError("Unknown message type");
       }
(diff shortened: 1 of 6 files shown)


skipruntime-ts/ts/src/internals/skipruntime_process.ts
```diff
@@ -55,9 +55,12 @@ export async function runWithServer_(
           values: [[http, "", "read-write"]],
         });
       } else if (http.type == "post") {
-        update(http, {}).catch((e: unknown) => {
-          throw e;
-        });
+        const maybePromise = update(http, {});
+        if (typeof maybePromise === "object") {
+          maybePromise.catch((e: unknown) => {
+            throw e;
```
mbouaziz, line 61: It seems weird to create the `Promise` and never await it, should we turn the listener to an `async` function?
ESLint complains with `@typescript-eslint/no-misused-promises` but it seems that's what people do

mbouaziz (approved):
LGTM, comment can be resolved independently

skipruntime-ts/ts/src/internals/skipruntime_process.ts
jberdine: Yeah, I really don't know. This code is type-incorrect before and after this PR. The type of `ws.on` specifies that the `message` callback returns `void`, and here we return `void` on some paths and `Promise<void>` on others. But there is no way to wait on the `Promise`. Making the callback `async` allows it to be `await`ed, but then the callback returns a `Promise` which the `ws` library does not itself await. It seems to be just fundamentally wrong to call async code in the callbacks passed to `ws`, it is expecting sync code. Maybe there is an async websockets library we should use instead.

PR #336, opened 2024-09-20
[ts runtime] External calls API

bennostein:
Adds a new `externalCall` operation on the SKStore context, which lets you set up an async lazy collection over some arbitrary async function, automatically wiring up a refresh token to govern polling frequency.

### DIFF
--- a/skipruntime-ts/ts/src/skipruntime_api.ts
+++ b/skipruntime-ts/ts/src/skipruntime_api.ts
@@ -528,6 +528,14 @@ export interface AsyncLazyCompute<
   call: (key: K, params: P) => Promise<AValue<V, M>>;
 }
 
+export interface ExternalCall<
+  K extends TJSON,
+  V extends TJSON,
+  Metadata extends TJSON,
+> {
+  call(key: K, timestamp: number): Promise<AValue<V, Metadata>>;
+}
+
 export interface SKStore extends Constant {
   /**
    * Creates a lazy reactive collection.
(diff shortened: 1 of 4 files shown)


skipruntime-ts/ts/tests/tests.ts
```diff
@@ -717,6 +714,100 @@ tests.push({
...
+  tokens: { token_4s: 4000 },
+  init: (
+    skstore: SKStore,
+    input: TableCollection<[number, number]>,
+    output: TableCollection<[number, string]>,
+  ) => {
+    const external = skstore.externalCall(
+      "token_4s",
+      MockExternal,
+    ) as AsyncLazyCollection<number, string, TJSON>;
```
mbouaziz: `as ...` unneeded

skipruntime-ts/ts/tests/tests.ts
```diff
@@ -717,6 +714,100 @@ tests.push({
...
+      success(1),
+      success(2),
+      success(3),
+    ]);
+    await timeout(1000);
+    //all loading at t = 4.5s (because refresh triggered at t = 4s)
+    check("testExternalCall", output.select({}), [
+      loading(1),
+      loading(2),
+      loading(3),
```
mbouaziz: Do we really want `loading` after initialization?
Should we use the latest value instead?

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -556,6 +564,17 @@ export interface SKStore extends Constant {
     ...params: Params
   ): AsyncLazyCollection<K, V, M>;
 
+  externalCall<
```
mbouaziz: Or just `external`?

skipruntime-ts/ts/tests/tests.ts
bennostein: Users of the interface can decide what behaviour is best for them -- I've updated the tests now to demonstrate that the previous value is still available when loading after a refresh.

mbouaziz (approved):

PR #337, opened 2024-09-23
Add target to update JS dependencies, and run it

jberdine:
Add a target so that `make update-js-deps` finds every package.json in the repo and updates the versions of dependencies to the latest.


mbouaziz (approved):
I'm not sure we have set up a CI for all projects but I'll suppose we do :crossed_fingers:

PR #338, opened 2024-09-23
Switch check targets from npm to bun

jberdine:
bun is much faster (about 17 vs 26 sec for make check-all in my setup)

### DIFF
--- a/skipruntime-ts/Makefile
+++ b/skipruntime-ts/Makefile
@@ -51,5 +51,5 @@ clean:
 	cd .. && make clean
 
 .PHONY: test
-test: build
-	cd core && npm run test
+test: bunbuild
+	cd core && bun run test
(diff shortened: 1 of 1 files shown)


jberdine:
This will fail CI until we next update the container images, so draft for now.

jberdine:
This is good to go now. Anyone have comments/concerns?

skipruntime-ts/Makefile
```diff
@@ -6,27 +6,27 @@ SCRIPT_DIR=$(shell dirname $(shell realpath $(firstword $(MAKEFILE_LIST))))
...
-check-all: build check-src check-examples
+check-all: bunbuild check-src check-examples
 
 .PHONY: build
 build:
 	cd .. && npm install && npm run build
 
 .PHONY: bunbuild
 bunbuild:
-	cd .. && bum install && bun run build
```
beauby, line 29: Oopsie

beauby (approved):

skipruntime-ts/Makefile
mbouaziz: Bummer!

PR #343, opened 2024-09-26
Resource-based architecture for skipruntime.

beauby:
This PR introduces the new resource-based architecture described in `rfc/008-architecture.org`.

Also introduces:
- removal of `sknpm` dependency,
- simplified replication protocol,
- shared `tsconfig.json`,
- shared `eslint.config.js`.


jberdine:
Maybe you found a way, but let me mention in case you are fighting with the linter, I needed to config the linter to use different type-checking options than the normal build. See https://github.com/SkipLabs/skdb/pull/332/commits/267839ece45428cf5d8976c96d1cdf2673def684 . Doing that seemed to require an extra tsconfig file, and I had trouble getting the path resolution to point to it from the eslint config to work.

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -581,3 +402,76 @@ export interface SKStore extends Constant {
...
+    key: string | number,
+    value: V[],
+  ): void;
+
+  patch<K extends TJSON, V extends TJSON>(
+    collection: string,
+    values: Entry<K, V>[],
+  ): void;
+
+  delete(collection: string, key: string | number): void;
```
bennostein, line 497: `delete` is a keyword in JS -- I think we should avoid it, seems like this might cause some issues for people, at least weird highlighting but potentially unintended semantics.

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -297,7 +235,27 @@ export interface EagerCollection<K extends TJSON, V extends TJSON>
    * @returns the value for this `key`, or null if no such value exists
    */
   maybeGetOne(key: K): Opt<V>;
+}
 
+export interface CollectionAccess<K extends TJSON, V extends TJSON>
+  extends CollectionReader<K, V> {
+  /**
+   * Get all values of an eager reactive collection, if one exists.
+   * Exist only in .
```
bennostein, line 243: Not sure what this comment is saying.

More generally, could you explain a bit about the relationship between the different kinds of collection interfaces here?  `Writer` vs. `Reader` is intuitive, but I think that the distinction between `Access` and `Reader` is a bit fuzzy

skipruntime-ts/ts/src/skipruntime_api.ts
```diff
@@ -274,12 +217,7 @@ export interface AsyncLazyCollection<
...
 
-/**
- * An _Eager_ reactive collection, whose values are computed eagerly and kept up
- * to date whenever inputs are changed
- */
-export interface EagerCollection<K extends TJSON, V extends TJSON>
-  extends Constant {
+export interface CollectionReader<K extends TJSON, V extends TJSON> {
   /**
    * Get (and potentially compute) all values mapped to by some key of a lazy reactive
```
beauby, line 224: Unrelated to this PR but this probably means to read "some key of a ~lazy~ eager reactive collection".

beauby:
> Maybe you found a way, but let me mention in case you are fighting with the linter, I needed to config the linter to use different type-checking options than the normal build. See [267839e](https://github.com/SkipLabs/skdb/commit/267839ece45428cf5d8976c96d1cdf2673def684) . Doing that seemed to require an extra tsconfig file, and I had trouble getting the path resolution to point to it from the eslint config to work.

I had noticed your changes, but I ended up enabling `noUncheckedIndexedAccess` across the board because it seems necessary for typescript to output relevant type information for eslint to do its business (and it didn't cause failures,). Are there cases where we actually want to disable it?

PR #344, opened 2024-09-28
Consolidate npm packages around npm workspaces.

beauby:
This has several benefits for a monorepo such as ours:
- shared `node_modules/` directory (at the root),
- shared (dev) dependencies (we can specify common
typescript/nodejs/eslint versions in the root `package.json`),
- installing all dependencies for all packages with `npm install` at
the root,
- building/linting all packages in one go with:
```
$ npm run build --workspaces --if-present
```
- no need to replace the `file:...` dependencies with version strings
when publishing.

One caveat is that running commands on all packages with `npm run foo
--workspaces` will run them in the order of packages within the
`workspaces` key of the root `package.json`. This matters for instance
for `npm run build --workspaces` as each package needs its
dependencies' type declarations (generated by `tsc`).

### DIFF
--- a/skdate/ts/package.json
+++ b/skdate/ts/package.json
@@ -1,4 +1,6 @@
 {
+  "name": "skdate",
+  "version": "1.0.0",
   "type": "module",
   "main": "./dist/sk_date.js",
   "types": "./dist/sk_date.d.ts",
@@ -20,10 +22,10 @@
     "@types/ws": "^8.5.12",
     "prettier": "3.2.5",
     "eslint": "^9.9.0",
-    "eslint-config-skiplabs": "file:../../eslint-config",
-    "tsconfig-skiplabs": "file:../../tsconfig"
+    "eslint-config-skiplabs": "^1.0.0",
+    "tsconfig-skiplabs": "^1.0.0"
   },
   "dependencies": {
-    "std": "file:../../prelude/ts"
+    "std": "^1.0.0"
   }
 }
--- a/skdate/ts/package.json
+++ b/skdate/ts/package.json
@@ -12,7 +12,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "lint": "eslint ."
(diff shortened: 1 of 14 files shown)


skdate/ts/package.json
```diff
@@ -15,13 +15,10 @@
     "lint": "eslint ."
   },
   "devDependencies": {
-    "typescript": "^5.6.2",
     "@playwright/test": "^1.36.2",
     "@types/express": "^4.17.21",
     "@types/node": "^22.6.0",
```
mbouaziz: Looks like `@types/node` is common too

skipruntime-ts/Makefile
```diff
@@ -5,83 +5,59 @@ SKARGO_PROFILE?=release
...
-build: prelude skjson replication
-	cd ts && npm i && npm run build
+build:
+	cd .. && npm install && npm run build
 
-bunrun-%: build
-	cd ts/examples && bun install && bun run $*.ts
+.PHONY: bunbuild
+bunbuild:
+	cd .. && num install && bun run build
```
mbouaziz: Did you mean `npm`?

sql/ts/tests/node_modules
```diff
@@ -0,0 +1 @@
+../../../node_modules/
```
mbouaziz, line 1: o_O

skipruntime-ts/Makefile
beauby: Oops I think I meant `bun`.

skdate/ts/package.json
beauby: Correct, I'll add it.

sql/ts/tests/node_modules
beauby: This is a symlink because the npm workspaces feature causes packages to be installed at the workspace root. It's a bit dirty but skdb seems to rely explicitly on some `node_modules` paths.

PR #345, opened 2024-09-30
[skipruntime-ts] Minor simplifications

mbouaziz:
A few minor simplifications that were waiting on my stack + renaming `sliced` -> `slice`

### DIFF
--- a/skipruntime-ts/ts/src/internals/skipruntime_impl.ts
+++ b/skipruntime-ts/ts/src/internals/skipruntime_impl.ts
@@ -109,7 +109,7 @@ export class EagerCollectionImpl<K extends TJSON, V extends TJSON>
     return this.context.size(this.eagerHdl);
   };
 
-  sliced(ranges: readonly [K, K][]): EagerCollection<K, V> {
+  slice(ranges: readonly [K, K][]): EagerCollection<K, V> {
     const sliceName =
       "range_" +
       ranges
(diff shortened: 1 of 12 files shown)


skipruntime-ts/ts/src/internals/skipruntime_module.ts
```diff
@@ -628,9 +628,9 @@ class NonEmptyIteratorImpl<T> implements NonEmptyIterator<T> {
     );
 
     return {
-      next() {
+      next(): IteratorResult<T> {
         const value = cloned_iter.next();
-        return { value, done: value == null } as IteratorResult<T>;
+        return value === null ? { value, done: true } : { value };
```
beauby, line 633: Does it make a difference that `done` is not explicitly set (to `false`) in the else case?

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
```diff
@@ -255,8 +255,7 @@ export class SKStoreImpl extends SkFrozen implements SKStore {
     const name = computeObj.constructor.name;
     const lazyHdl = this.context.lazy(
       name,
-      (selfHdl: LazyCollection<K, V>, key: K) =>
-        computeObj.compute(selfHdl, key),
+      computeObj.compute.bind(computeObj),
```
beauby, line 258: Open question: do we generally prefer `obj.method.bind(obj)` over `(arg1, arg2) => obj.method(arg1, arg2)`?

beauby (approved):
Minor questions, LGTM otherwise.

skipruntime-ts/ts/src/internals/skipruntime_module.ts
mbouaziz: No, it works as well.

skipruntime-ts/ts/src/internals/skipruntime_impl.ts
mbouaziz: With `(arg1, arg2) => obj.method(arg1, arg2)`, I want to simplify to just `obj.method` and then it fails... so `obj.method.bind(obj)`, although ugly, makes it explicit

beauby: In that case let's open an issue with a reminder to use that style across the board (there are instances of the lambda variant in the skipruntime client or server).

PR #350, opened 2024-09-30
[skipruntime-ts] Improvements on TimedQueue

mbouaziz:
Mainly:
- using a binary heap for the priority queue
- [RFC] do not leap re-updates (see commit with the same title)

### DIFF
--- a/skipruntime-ts/core/src/internals/skipruntime_module.ts
+++ b/skipruntime-ts/core/src/internals/skipruntime_module.ts
@@ -1238,7 +1238,7 @@ class LinksImpl implements Links {
       // Get a run uuid to build to allow function mapping reload in case of persistence
       const uuid = this.env.crypto().randomUUID();
       const jsu = skjson();
-      const time = new Date().getTime();
+      const time = Date.now();
       const result = utils.runWithGc(() =>
         Math.trunc(
           fromWasm.SkipRuntime_createFor(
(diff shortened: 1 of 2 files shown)


skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -1386,111 +1386,128 @@ class Manager implements ToWasmManager {
...
+    }
+  }
+
+  peekPriority(): number | undefined {
+    return this.heap[0]?.[0];
+  }
+
+  popValues(): T[] | undefined {
+    const top = this.heap[0];
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
beauby, line 1417: Is this really needed?

beauby (commented):
The code LGTM.

Unrelated to this PR: what's the benefit of using a `TimeQueue` vs an array of `Timeout`s created using JS's `setInterval()`?

skipruntime-ts/core/src/internals/skipruntime_module.ts
mbouaziz: We have enabled `reportUnusedDisableDirectives: "error",` so we would get an error if it wasn't needed

mbouaziz: It comes from `"noUncheckedIndexedAccess": false,` resulting in `top` not being `undefined`

beauby: I thought we had enabled `noUncheckedIndexedAccess`, I'll look into it.

PR #353, opened 2024-10-01
[skipruntime-ts] More cleanups

mbouaziz:
Biggest change is: splitting out `_init` and `_rest` from `skip-runtime.ts` so that it contains only the public exports.


skipruntime-ts/core/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "@skiplabs/tsconfig",
-  "exclude": ["dist/**/*.ts", "examples/**/*.ts", "tests/**/*.ts"]
+  "exclude": ["dist/**/*.ts"]
```
beauby: I think this is the default.
```suggestion
```

beauby (approved):
LGTM modulo the unnecessary `tsconfig.json` override.

skipruntime-ts/core/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "@skiplabs/tsconfig",
```
mbouaziz: In that case...
```suggestion
  "extends": "@skiplabs/tsconfig"
```

PR #354, opened 2024-10-01
[SkStore] Minor cleanups

mbouaziz:
...while reading the code


sql/src/SqlEval.sk
```diff
@@ -1607,7 +1607,7 @@ class Evaluator{options: Options, user: ?UserFile} {
     ifNotExists: Bool,
   ): void {
     dirName = dirDescr.dirName;
-    if (context.unsafeMaybeGetEagerDir(dirName) is Some _ && ifNotExists) {
+    if (ifNotExists && context.unsafeMaybeGetEagerDir(dirName) is Some _) {
```
beauby, line 1610: Yes.

prelude/src/core/primitives/String.sk
```diff
@@ -250,8 +250,8 @@ native class .String uses Hashable, Show, Orderable, AsBytes {
     i.rfind(f).map(_ -> i)
   }
 
-  fun chars(): Vector<Char> {
-    this.getIter().collect(Vector)
+  fun chars(): Array<Char> {
+    this.getIter().collect(Array)
```
beauby: This is fine, but shouldn't `chars()` return the iterator itself and leave it up to the caller to collect it into an `Array`?

mbouaziz: First, I'll fix the CI related to this change :stuck_out_tongue:  then I'll see whether it makes sense, `Array` may be just fine in all cases

mbouaziz: Well well, actually collecting to an `Array` is worse because it first collects to a `Vector` :disappointed: 
I'll get rid of this commit

PR #356, opened 2024-10-01
[skipruntime] Get rid of `TimedQueue`.

beauby:
It was basically a re-implementation of `setInterval()`.

### DIFF
--- a/skipruntime-ts/core/src/internals/skipruntime_module.ts
+++ b/skipruntime-ts/core/src/internals/skipruntime_module.ts
@@ -831,7 +831,7 @@ class LinksImpl implements Links {
   env: Environment;
   handles: Handles;
   skjson?: SKJSON;
-  timedQueue?: TimedQueue;
+  timedQueue?: ReturnType<typeof setInterval>[];
 
   detachHandle!: <T>(fn: Handle<T>) => void;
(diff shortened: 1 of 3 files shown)


skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -831,7 +831,7 @@ class LinksImpl implements Links {
   env: Environment;
   handles: Handles;
   skjson?: SKJSON;
-  timedQueue?: TimedQueue;
+  timedQueue?: ReturnType<typeof setInterval>[];
```
mbouaziz, line 834:
```suggestion
  timedQueue?: NodeJS.Timeout[];
```
~~You want the return type, not the function itself~~ Hence `ReturnType`

Ignore me

mbouaziz (commented):
:smiling_face_with_tear:

skipruntime-ts/core/src/internals/skipruntime_module.ts
beauby: Hmm the `ReturnType<T>` should give the return type of the function type `T`. The reason I didn't go with `NodeJS.Timeout` is to preserve browser compatibility.

PR #357, opened 2024-10-01
[skipruntime/core] Simplify tests.

beauby:
Remove unnecessary `check()` function.


mbouaziz:
I inferred that the goal of `check` is to show the name of the test when an equality is broken.

beauby:
I suspect Playwright wasn't gathering/displaying that info correctly, but this shouldn't be an issue anymore (Mocha gives the file:lineno of failed expectations).

PR #360, opened 2024-10-03
[runtime] Switch from steady_clock to system_clock

jberdine:
- system_clock is not monotonic (bad) (NTP adjustments, etc.) but counts from the Unix epoch (good)
  - steady_clock is monotonic (good) but counts from system boot (bad!)


beauby (approved):

gregsexton:
Ah yeah this is bad. Steady clock is usually the cycle counter register. And I see that this is used in the uuids :D Good catch.

PR #362, opened 2024-10-03
SkipRuntime cleanups and doc comments

bennostein:
Some cleanups and improved inline documentation for Skip runtime, focusing on `skipruntime_api.ts` to make sure everything is understandable and well-documented.

### DIFF
--- a/skipruntime-ts/core/src/internals/skipruntime_impl.ts
+++ b/skipruntime-ts/core/src/internals/skipruntime_impl.ts
@@ -1,5 +1,5 @@
 import { type int, type ptr, type Opt, cloneIfProxy } from "std";
-import type { Context } from "./skipruntime_types.js";
+import type { Context as InternalContext } from "./skipruntime_types.js";
 import type * as Internal from "./skipruntime_internal_types.js";
 import type {
   Accumulator,
@@ -7,7 +7,7 @@ import type {
   LazyCollection,
   AsyncLazyCollection,
   Mapper,
-  SKStore,
+  Context,
   RuntimeFactory,
   Loadable,
   JSONObject,
@@ -75,7 +75,7 @@ export class EagerCollectionImpl<K extends TJSON, V extends TJSON>
   implements EagerCollection<K, V>
 {
   constructor(
-    private context: Context,
+    private context: InternalContext,
     private eagerHdl: string,
   ) {
     super();
@@ -193,7 +193,7 @@ class LazyCollectionImpl<K extends TJSON, V extends TJSON>
   implements LazyCollection<K, V>
 {
   constructor(
-    protected context: Context,
+    protected context: InternalContext,
     protected lazyHdl: string,
   ) {
     super();
@@ -218,7 +218,7 @@ export class LSelfImpl<K extends TJSON, V extends TJSON>
   implements LazyCollection<K, V>
 {
   constructor(
-    protected context: Context,
+    protected context: InternalContext,
     protected lazyHdl: ptr<Internal.LHandle>,
   ) {
     super();
@@ -238,8 +238,8 @@ export class LSelfImpl<K extends TJSON, V extends TJSON>
   }
 }
 
-export class SKStoreImpl extends SkFrozen implements SKStore {
-  constructor(private context: Context) {
+export class ContextImpl extends SkFrozen implements Context {
+  constructor(private context: InternalContext) {
     super();
     Object.freeze(this);
   }
@@ -322,7 +322,7 @@ export class SKStoreImpl extends SkFrozen implements SKStore {
 export class RuntimeFactoryImpl implements RuntimeFactory {
   //
   constructor(
-    private context: () => Context,
+    private internalContext: () => InternalContext,
     private create: (
       init: (
         collections: Record<string, EagerCollection<TJSON, TJSON>>,
@@ -439,7 +439,7 @@ export class EagerCollectionReader<K extends TJSON, V extends TJSON>
   implements CollectionReader<K, V>
 {
   constructor(
-    private context: Context,
+    private context: InternalContext,
     private eagerHdl: string,
   ) {}
 
@@ -475,7 +475,7 @@ export class EagerCollectionWriter<K extends TJSON, V extends TJSON>
   implements CollectionWriter<K, V>
 {
   constructor(
-    private context: Context,
+    private context: InternalContext,
     private eagerHdl: string,
   ) {}
 
@@ -494,7 +494,7 @@ export class EagerCollectionWriter<K extends TJSON, V extends TJSON>
 
 export class SkipRuntimeImpl implements SkipRuntime {
   constructor(
-    private context: Context,
+    private context: InternalContext,
     private writables: Record<string, CollectionWriter<TJSON, TJSON>>,
   ) {}
 
--- a/skipruntime-ts/core/src/internals/skipruntime_impl.ts
+++ b/skipruntime-ts/core/src/internals/skipruntime_impl.ts
@@ -24,7 +24,6 @@ import type {
   CollectionWriter,
   CallResourceCompute,
   Watermarked,
-  Notifier,
   ReactiveResponse,
   SkipBuilder,
 } from "../skipruntime_api.js";
@@ -467,8 +466,11 @@ export class EagerCollectionReader<K extends TJSON, V extends TJSON>
     return this.context.getDiff(this.eagerHdl, from);
   }
 
-  subscribe(from: string, notify: Notifier<K, V>): bigint {
-    return this.context.subscribe(this.eagerHdl, from, notify);
+  subscribe(
+    since: string,
+    f: (values: Entry<K, V>[], watermark: string, update: boolean) => void,
+  ): bigint {
+    return this.context.subscribe(this.eagerHdl, since, f);
   }
 }
 
@@ -587,10 +589,10 @@ export class SkipRuntimeImpl implements SkipRuntime {
 
   subscribe<K extends TJSON, V extends TJSON>(
     collectionName: string,
-    from: string,
-    notify: Notifier<K, V>,
+    since: string,
+    f: (values: Entry<K, V>[], watermark: string, update: boolean) => void,
   ) {
-    return this.context.subscribe(collectionName, from, notify);
+    return this.context.subscribe(collectionName, since, f);
   }
 
   unsubscribe(id: bigint): void {
--- a/skipruntime-ts/core/src/internals/skipruntime_impl.ts
+++ b/skipruntime-ts/core/src/internals/skipruntime_impl.ts
@@ -16,7 +16,7 @@ import type {
   LazyCompute,
   AsyncLazyCompute,
   NonEmptyIterator,
-  EntryPoint,
+  Entrypoint,
   ExternalCall,
   CollectionReader,
   SkipRuntime,
@@ -344,7 +344,7 @@ export class SKStoreFactoryImpl implements SKStoreFactory {
       collections: Record<string, EagerCollection<TJSON, TJSON>>,
     ) => CallResourceCompute,
     inputs: Record<string, [TJSON, TJSON][]> = {},
-    remotes: Record<string, EntryPoint> = {},
+    remotes: Record<string, Entrypoint> = {},
     tokens: Record<string, number> = {},
   ): SkipBuilder {
     const context = this.context();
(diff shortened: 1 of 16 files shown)


skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -445,12 +448,6 @@ export type Watermarked<K extends TJSON, V extends TJSON> = {
   update?: boolean;
 };
 
-export type Notifier<K extends TJSON, V extends TJSON> = (
-  values: Entry<K, V>[],
-  watermark: string,
-  update: boolean,
-) => void;
```
mbouaziz, line 454: Not sure about removing this type, it seems handy to refer to it at several places

skipruntime-ts/core/src/internals/skipruntime_types.ts
```diff
@@ -395,8 +394,14 @@ export interface FromWasm {
...
-    from: bigint,
-    notifyFn: Handle<Notifier<TJSON, TJSON>>,
+    since: bigint,
+    f: Handle<
+      (
+        values: Entry<TJSON, TJSON>[],
+        watermark: string,
+        update: boolean,
+      ) => void
+    >,
```
mbouaziz, line 413: `SkipRuntime_subscribe` should be generic in `K` and `V` to avoid the `as` in `ContextImpl.subscribe`

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -160,26 +158,24 @@ export interface NonEmptyIterator<T> extends Iterable<T> {
...
 
   /**
    * Performs the specified action for each element in the iterator.
-   * @param callbackfn - A function that accepts up to thow arguments. forEach calls the callbackfn function one time for each element.
-   * @param thisArg - An object to which the this keyword can refer in the callbackfn function. If thisArg is omitted, undefined is used as the this value.
+   * @param f - A callback to invoke on each element
+   * @param thisObj - An object to bind as `this` within the `f` invocations; if omitted, `undefined` is used
    */
-  forEach(callbackfn: (value: T, index: number) => void, thisArg?: any): void;
+  forEach(f: (value: T, index: number) => void, thisObj?: any): void;
```
mbouaziz, line 170: I'm wondering if we should use this overloaded type:
```ts
forEach(f: (this: undefined, value: T, index: number) => void, thisObj?: undefined): void;
forEach<This>(f: (this: This, value: T, index: number) => void, thisObj: This): void;
```
Maybe that's too precise... the signature of `Array.map` for example is not that precise...

skipruntime-ts/core/src/internals/skipruntime_impl.ts
```diff
@@ -7,24 +7,23 @@ import type {
   LazyCollection,
   AsyncLazyCollection,
   Mapper,
-  SKStore,
-  SKStoreFactory,
+  Skip,
```
beauby: Recap of offline discussion: `Skip` might not be a convincing name here (what is "a Skip" or "the Skip"?), as it refers to the whole product. `SkipRuntime` is (currently) already taken, but we should consider renaming it.

jberdine:
I have a PR I was about to put up that renames SKStore to Context, but see you chose Skip here. Which do we want? Personally I prefer Context since it seems to me to be more informative than Skip. WDYT?

bennostein:
Yep, `Skip` here is not great -- let's go with `Context` for now. (especially good as Daniel's current PR gets rid of the other internal "Context" from the JS side) 

I went with "Skip" initially as it's the proper noun we expect people to use to refer to the whole framework, and these are the global/static operations of the system.  But I agree that it's not very descriptive

skipruntime-ts/core/src/internals/skipruntime_impl.ts
jberdine: Re `SkipRuntime`, I worry that using that name, which for external communications we want to use for a large chunk of functionality, doesn't fit well. There is a lot of functionality that in the code won't need to be directly associated with this type from the perspective of users, and we don't want to somehow distance that functionality from the "Skip Runtime" name.

jberdine:
Ah, ok, I renamed the other Context too, will rebase.

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -243,19 +243,22 @@ export interface CollectionReader<K extends TJSON, V extends TJSON> {
...
   /**
-   * Get a diff of a collection according a from watermark, if one exists.
-   * @param from The from watermark to retrieve diff from
-   * @returns {Watermarked} an array en key, values pair with the corresponding watermark.
+   * Get a diff of added/updated entries in a collection since some watermark.
+   * @param since A watermark from which to compute the diff
+   * @returns {Watermarked} An object containing updated entries and a new watermark string.
    */
-  getDiff(from: string): Watermarked<K, V>;
+  getDiff(since: string): Watermarked<K, V>;
```
jberdine: A question that occurs to me thinking as a user is: How are deletions expressed?

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -243,19 +243,22 @@ export interface CollectionReader<K extends TJSON, V extends TJSON> {
...
 
   /**
-   * Allow to subsribe the updates of the collection
-   * @param from The watermark where to start the update
-   * @param notify The function to call on collection update
-   * @returns {Watermarked} an array en key, values pair with the corresponding watermark.
+   * Subscribe to updates of the collection since some initial watermark
+   * @param since The watermark from which to start watching for updates
```
jberdine: maybe s/start watching for/receive/

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -243,19 +243,22 @@ export interface CollectionReader<K extends TJSON, V extends TJSON> {
...
+   * @param f The function to call on collection update
+   * @returns A subscription identifier, which can be passed to `unsubscribe` to cancel this subscription
    */
-  subscribe(from: string, notify: Notifier<K, V>): bigint;
+  subscribe(
+    since: string,
+    f: (values: Entry<K, V>[], watermark: string, update: boolean) => void,
+  ): bigint;
```
jberdine, line 259: Do we want to expose the concrete representation type of these subscription ids in the api?

skipruntime-ts/core/src/skipruntime_api.ts
jberdine, line 454: A nice thing about having this type declared is that here would be a good place to document the arguments.

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -1,16 +1,26 @@
...
+ */
+export type TJSON = number | boolean | string | JSONObject | (TJSON | null)[];
 export type JSONObject = { [key: string]: TJSON | null };
 
-export type TJSON = number | JSONObject | boolean | (TJSON | null)[] | string;
-
+/**
+ * A `Param` is a valid parameter to a Skip runtime mapper function: either a constant JS
+ * value or a Skip-runtime-managed value.  In either case, restricting mapper parameters to
+ * this type ensures that reactive computations can be re-evaluated as needed with consistent
```
jberdine, line 21: Not for this PR, but maybe "ensures" is too strong here, since if users try they can break it.

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -25,6 +35,11 @@ export type Param =
  * `Loading`, or `Error`
  */
 
+export type Loadable<V extends TJSON, M extends TJSON> =
+  | Success<V, M>
+  | Loading<V, M>
+  | Error<V, M>;
```
jberdine, line 41: What are others' view on single letter type parameters versus longer role-indicating names in the public api?

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -160,26 +158,24 @@ export interface NonEmptyIterator<T> extends Iterable<T> {
...
+   * Returns an array containing all values of the iterator
    */
   toArray: () => T[];
 
   /**
    * Performs the specified action for each element in the iterator.
-   * @param callbackfn - A function that accepts up to thow arguments. forEach calls the callbackfn function one time for each element.
-   * @param thisArg - An object to which the this keyword can refer in the callbackfn function. If thisArg is omitted, undefined is used as the this value.
+   * @param f - A callback to invoke on each element
+   * @param thisObj - An object to bind as `this` within the `f` invocations; if omitted, `undefined` is used
```
jberdine: Is "if omitted, `undefined` is used" exactly the behavior of optional arguments in TS? If so, if we explicitly mention it here, will users be confused thinking that since we're mentioning it there must be something nonstandard happening, but they can't see what it is? That is at least what happens for me.

jberdine (approved):
LGTM

Re s/EntryPoint/Entrypoint/, AFAIU "entrypoint" isn't a word, it is usually two words "entry point" or at best hyphenated "entry-point", both of which seem to favor `EntryPoint` over `Entrypoint`. But I don't know that motivation for this particular change.

If we want to go for the Context name instead of Skip, it would be easier for me if you dropped the last commit.

bennostein:
> Re s/EntryPoint/Entrypoint/, AFAIU "entrypoint" isn't a word, it is usually two words "entry point" or at best hyphenated "entry-point", both of which seem to favor `EntryPoint` over `Entrypoint`. But I don't know that motivation for this particular change.

Purely vibes-based decision: camel case "EntryPoint" looked odd to me, whereas "Entrypoint" feels more natural.  But obviously not a big deal, feel free to veto or revisit.

> If we want to go for the Context name instead of Skip, it would be easier for me if you dropped the last commit.

I'm making some further changes around the Factory/Builder stuff to streamline things and get rid of some other types, so will keep pressing along with revising the last commit for now, if that's okay with you.

skipruntime-ts/core/src/skipruntime_api.ts
beauby: Deletes are expressed as the key having an empty array as value.

jberdine: Ok, but my point is that to begin to see that a user would have to unpack the definition of Watermarked, and then of Entry, to see that Entry is `[K, V[]]` and then guess that a key associated to an empty array is how a delete is represented. It would probably be good to mention here on `getDiff` that removals are also represented by the `Watermarked` value, without saying how here, just leaving that to the docs of `Watermarked`. Sorry, this should be just a note to myself, I'm happy to iterate on this once this lands.

skipruntime-ts/core/src/internals/skipruntime_impl.ts
beauby: Agreed that `SkipRuntime` is not great either, I was merely mentioning it as being available. `Context` works for me.

skipruntime-ts/core/src/skipruntime_api.ts
beauby: Oh I see. Yes, you are right, this should be made explicit in the docstring. Also, I'm not a 100% convinced about having `getDiff()` return a `Watermarked<K, V>` instead of explicitly returning a `[Entry<K, V>[], boolean, Watermark]`.

jberdine: I don't know about the name `Watermarked`, but a good thing about having a type definition is that there is one canonical place to document it.

beauby: I agree in general but in this case it seems `Watermarked` is only ever used as the return type of `getDiff()`, which hints to me that the right place to document it is in `getDiff()`.

skipruntime-ts/core/src/skipruntime_api.ts
mbouaziz: I prefer longer role-indicating names. Also it makes they more easily greppable.
In #358, I `s/M/Metadata` and `= never` as default type

bennostein: Agreed (maybe with the exception of K/V as key/value or T for the members of a collection)

jberdine: Yes, I wouldn't argue over those exceptions. In cases where there become more than say 2 type variables though, it might help to make things clearer to spell out even those.

skipruntime-ts/core/src/skipruntime_api.ts
jberdine: LGTM

skipruntime-ts/core/src/skipruntime_api.ts
bennostein: I've got some more changes I'll send soon renaming `Watermarked` to `CollectionUpdate` and using it also as the argument of various subscription lambdas/handles that currently take its same fields as 3 destructured arguments.  Will leave be for now on this PR.

PR #366, opened 2024-10-07
[skc] Clean up `compile()`.

beauby:
A few cosmetic changes to separate the compute graph definition vs triggering of actual compute (by writing to input directories).

### DIFF
--- a/skiplang/compiler/src/compile.sk
+++ b/skiplang/compiler/src/compile.sk
@@ -206,7 +206,7 @@ fun compile(
   ensureCompatibleLLVMVersion();
 
   backendDirName = SKStore.DirName::create("/backend/");
-  if (context.unsafeMaybeGetDir(FileCache.fileDirName) is None()) {
+  if (context.unsafeMaybeGetEagerDir(backendDirName) is None()) {
     _ = context.mkdir(x ~> x, x ~> x, FileCache.fileDirName);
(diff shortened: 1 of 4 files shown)


skiplang/compiler/src/compile.sk
```diff
@@ -193,191 +193,200 @@ fun ensureCompatibleLLVMVersion(): void {
...
+): void {
+  if (fileNames.isEmpty() && config.dependencies.isEmpty()) {
+    print_error("fatal error: no input files\n");
+    skipExit(4); // Exit code used by g++ for no input files
   };
 
+  ensureCompatibleLLVMVersion();
+
+  // FIXME: Backend initialization should _not_ peak into the config
+  // as it is likely to change across invocations.
```
jberdine: s/peak/peek

AFAIU the only use made of the config is to init OuterIstToIR, and that just uses ptrBitSize. It seems that ptrBitSize should not change between invocations, and so a reasonable plan would be to pass only that value down to OuterIstToIR. WDYT?

jberdine (approved):

skiplang/compiler/src/compile.sk
beauby: Good catch!

Regarding `ptrBitSize`, it can change if you successively build for `amd64` then `wasm32`. Currently we use a separate `state.db` for each arch to prevent inconsistencies.
Regardless, I agree that passing only `ptrBitSize` instead of the whole `config` is a first step in limiting chaos.

PR #371, opened 2024-10-07
[skipruntime] Type cleanups and improvements

bennostein:
* Use opaque type instead of raw bigint for subscription IDs
* Use bigint instead of converting to/from string for watermarks, and introduce an opaque type there as well
* Unify treatment of collection diffs/updates: now the same type is used for the return value of `getDiff` and the callback argument in `subscribe` (which previously took the same three values, but destructured into separate arguments). I think this is an improvement on the `Notifier` function type alias I deleted in #362 in terms of clarity, but open to reverting if people prefer the previous setup.
* Some minor semantics-preserving changes, getting rid of some single-letter type variables and tweaking comment prose.

### DIFF
--- a/skipruntime-ts/core/src/skipruntime_api.ts
+++ b/skipruntime-ts/core/src/skipruntime_api.ts
@@ -18,8 +18,8 @@ export type JSONObject = { [key: string]: TJSON | null };
 /**
  * A `Param` is a valid parameter to a Skip runtime mapper function: either a constant JS
  * value or a Skip-runtime-managed value.  In either case, restricting mapper parameters to
- * this type ensures that reactive computations can be re-evaluated as needed with consistent
- * semantics.
+ * this type helps developers to ensure that reactive computations can be re-evaluated as
+ * needed with consistent semantics.
  */
 export type Param =
   | null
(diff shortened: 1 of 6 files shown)


skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -243,20 +243,17 @@ export interface CollectionReader<K extends TJSON, V extends TJSON> {
   /**
    * Get a diff of added/updated entries in a collection since some watermark.
    * @param since A watermark from which to compute the diff
-   * @returns {UpdatedValues} An object containing updated entries and a new watermark string.
+   * @returns {CollectionUpdate} An object containing updated entries and a new watermark string.
```
jberdine: "watermark _string_"

skipruntime-ts/core/src/internals/skipruntime_impl.ts
```diff
@@ -264,21 +264,23 @@ export class ContextImpl extends SkFrozen implements Context {
   asyncLazy<
     K extends TJSON,
     V extends TJSON,
-    P extends TJSON,
+    AsyncParams extends TJSON,
     Params extends Param[],
```
jberdine: Just a note for ourselves: I think we still have some work to do to explain the difference between these two types of params clearly.

jberdine (approved):

PR #373, opened 2024-10-07
[sktore] Remove sub directories on EagerDir delete.

skiplabsdaniel:


skiplang/prelude/ts/src/sk_types.ts
```diff
@@ -950,3 +950,16 @@ export function cloneIfProxy<T>(v: T): T {
   }
   return v;
 }
+
+export function errorObjectAsError(e: ErrorObject): Error {
```
beauby, line 954: What is `ErrorObject` and why not have it subclass `Error` directly?

skiplabsdaniel: Indeed, I need to check

PR #377, opened 2024-10-08
[WIP] OSX support

beauby:


skiplang/compiler/src/skipParse.sk
```diff
@@ -22,160 +22,153 @@ const fileDir: SKStore.EHandle<
...
-);
+fun pkgDelta(
+  old_srcs: Array<(String, Int)>,
+  new_srcs: Array<(String, Int)>,
+): (Array<String>, Array<String>, Array<String>) {
+  added_files = mutable Vector[];
+  modified_files = mutable Vector[];
+  deleted_files = mutable Vector[];
+
+  // FIXME: This is O(n^2).
```
jberdine, line 82: Are the srcs arrays already sorted? A linear traversal of new_srcs seems unavoidable. So perhaps the best that can be done is to sort old_srcs and use a binary search instead of Array.find. Maybe it would be better to also sort new_srcs and then be able to narrow the binary search of old_srcs at each step.

jberdine (commented):
Most of this isn't specific to macOS right? It would be better to divide into 2 PRs I think.

beauby:
> Most of this isn't specific to macOS right? It would be better to divide into 2 PRs I think.

That's correct. The aim of this branch is to get OSX support down the line, and I made it a PR for visibility. I'll split self-contained commits out into separate PRs.

mbouaziz:

PR #380, opened 2024-10-09
Rename NonEmptyIterator.convert to map

jberdine:
To do so requires allowing the mapped function to be impure.

### DIFF
--- a/skiplang/prelude/src/skstore/Handle.sk
+++ b/skiplang/prelude/src/skstore/Handle.sk
@@ -120,7 +120,7 @@ mutable class NonEmptyIterator<+T>(
     )
   }
 
-  mutable fun convert<V>(conv: T ~> V): mutable NonEmptyIterator<V> {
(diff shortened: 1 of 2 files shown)


jberdine:
I investigated this following on from the discussion here: https://github.com/SkipLabs/skdb/pull/364#discussion_r1787796683 . I'm not sure we want this change, but put this up since it is easier to reuse the `map` name than what seemed possible from previous discussion.

bennostein (approved):
Much more intuitive to me this way, unless there's some reason to require a pure argument.  Even in that case, I'd argue for something like `mapPure` over `convert`.

jberdine:
Yes, agree, and an argument could be made that all `map` functions in skip should take pure closures and a more explicitly-named version should be used when mutation is possible/needed.

mbouaziz:
This now becomes the `map` overriding `Iterator#map` so this change is not just a renaming, it may have effects on code using `map` before

PR #382, opened 2024-10-09
Refactor out some shared logic for reactive-response HTTP header extraction

bennostein:


skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -276,6 +276,17 @@ export type CollectionUpdate<K extends TJSON, V extends TJSON> = {
   isInitial?: boolean;
 };
 
+/**
+ * A `ReactiveResponse` contains metadata which can be used to initiate and manage reactive
+ * connections to/from a service.  It specifies a `collection` and a current `watermark`,
+ * and is sent by Skip services in the `"x-reactive-response"` HTTP header by default,
```
jberdine: Not specific to this PR, but my understanding is that the `X-` prefix on headers has been deprecated for a long time now: https://datatracker.ietf.org/doc/html/rfc6648. And while the headers are case insensitive, I think the most common practice is to capitalize them, such as `X-Reactive-Response`.

jberdine (approved):

skipruntime-ts/core/src/skipruntime_api.ts
bennostein: How do you feel about `Skip-Reactive-Response`?  Not specific to this PR indeed, but as good a time as any to fix it!

beauby: Yes, we should use the opportunity to find proper names for the headers (they weren't named carefully).
Since it is Skip-specific, I would suggest to name them `Skip-Something`. `Skip-Response-Token`/`Skip-Reactive-Token`/`Skip-Subscription-Token`? The name Skip is a bit confusing here, but that's life.

skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -276,6 +276,17 @@ export type CollectionUpdate<K extends TJSON, V extends TJSON> = {
   isInitial?: boolean;
 };
 
+/**
+ * A `ReactiveResponse` contains metadata which can be used to initiate and manage reactive
+ * connections to/from a service.  It specifies a `collection` and a current `watermark`,
+ * and is sent by Skip services in the `"Skip-Reactive-Response"` HTTP header by default,
```
beauby: This does not matter much for now, but I think we should refrain from stating guarantees that the `Skip-Reactive-Response` header will actually contain the collection name + watermark in an intelligible form. For instance, we could have it return a UUID that maps back to something server-side.

bennostein: I don't think this coment states such a guaranee -- even in an unintelligible form, a UUID still "specifies" a collection and watermark.  But agreed that it's not a big deal at the moment, while it's still just direct string-encoded JSON.

skipruntime-ts/core/src/skipruntime_utils.ts
```diff
@@ -36,3 +36,14 @@ export class Max implements Accumulator<number, number> {
     return value < acc ? acc : null;
   }
 }
+
+export function getReactiveResponse(
+  httpHeaders: Headers,
```
beauby: Suggestion: how about we name this `parseReactiveResponse()` and have it take the header value as parameter? It adds a small burden to the caller (who would have to call `parseReactiveResponse(headers.get("Skip-Reactive-Response"))` instead of `getReactiveResponse(headers)` but it would not limit the scope to HTTP `Headers` objects.

bennostein: I applied this renaming but, since we're in dynamic language land already, I just made overloaded the function on both input types so callers don't need to grab the header themselves.  Saves a few keystrokes, but more importantly IMO minimizes the number of places where we have that string literal.

skipruntime-ts/examples/database-client.ts
```diff
@@ -27,13 +31,7 @@ const [_e, headers] = await fetchJSON<ReactiveResponse>(
...
 );
 
-const strReactiveResponse = headers.get("x-reactive-response");
-const reactive = strReactiveResponse
-  ? (JSON.parse(strReactiveResponse, (key: string, value: string) => {
-      if (key == "watermark") return BigInt(value);
-      return value;
-    }) as ReactiveResponse)
-  : undefined;
+const reactive = parseReactiveResponse(headers);
```
beauby, line 34: Tangential to this PR, but this makes me think that deconstruction of the `Reactive-Response-Token` should happen within the skip client, and the token itself be passed transparently through to `subscribe()`, instead of `subscribe()` taking a collection name + watermark. Just something to keep in mind, no action required on your end in this PR.

skipruntime-ts/core/src/skipruntime_utils.ts
beauby: Orthogonal: We should probably define a global constant for the header name in any case.

skipruntime-ts/examples/database-client.ts
bennostein: Yeah, I had the same thought (as subscribe takes the same fields, just destructured into function arguments) but figured there may be places where you want to subscribe and have the collection name / watermark lying around not in a `ReactiveResponse` struct.

PR #383, opened 2024-10-09
[skc] Store sklib sources in `lib.skmeta`.

beauby:
Context:
The `sklib` file format is a ranlib archive that contains possible object files from C extensions, as well as skip-specific metadata. It can thus be passed straight to the linker, or be used by `skc`. Part of the required metadata is information about the defined (skip) symbols in the sklib. In practice, we pack the source code of all files from the package into the sklib.

So far, the skip source files were embedded as members of the ranlib archive, which works fine on Linux as the linker seems to ignore files that are not ELF objects, but causes issues on macOS, where the linker complains about non Mach-O files. In order to work around this, the idea is to pack the source files of the package along with the metadata inside the `lib.skmeta` member of the archive. This would still be an issue, but the next step is to embed the `lib.skmeta` into a dummy object file, leaving a clean ranlib archive that should please any linker.


skiplang/compiler/src/sklib.sk
```diff
@@ -71,75 +95,32 @@ class Metadata(
...
-      // TODO: Properly fail.
-      }),
-    );
-    file.close();
-    Success(res)
-  }
+    } match {
+    | Some(bytes) -> bytes
+    | None() -> "".bytes()
+    // TODO: Properly fail.
```
jberdine: Did you mean to do this before landing this PR, or maybe change this to FIXME?

jberdine (approved):
I edited the PR summary, but probably the commit summary should be edited instead/too.

skiplang/compiler/src/sklib.sk
beauby: This was copied over from the previous version of the code, but agreed that it should be a `FIXME`.

PR #389, opened 2024-10-09
[skipruntime] Share ReactiveResponse to header

skiplabsdaniel:


jberdine (approved):

skipruntime-ts/server/src/rest.ts
```diff
@@ -18,12 +21,8 @@ export function createRESTServer(runtime: SkipRuntime): express.Express {
...
-        "Skip-Reactive-Response-Token",
-        JSON.stringify(data, (_key: string, value: unknown) =>
-          typeof value === "bigint" ? value.toString() : value,
-        ),
-      );
+      const [name, value] = reactiveResponseHeader(data);
+      res.set(name, value);
       res.status(200).json({});
     } catch (e: unknown) {
       res.status(500).json(e instanceof Error ? e.message : e);
```
beauby, line 28: We should respond with `404 NOT FOUND` for `UnknownCollectionError`.

PR #391, opened 2024-10-09
Enable noUncheckedIndexedAccess

jberdine:
Add dev workflow automation to run the type-checker on all the ts sources, enable noUncheckedIndexedAccess, and fix all the existing violations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,13 +65,9 @@ build/init.sql: sql/privacy/init.sql
 update-js-deps:
 	find . -name node_modules -not -prune -or -name target -not -prune -or -name package.json -exec sh -c 'cd $$(dirname "$$0"); bun update --latest' {} \;
 
-CHECK_TARGETS=$(patsubst %/Skargo.toml,check-%,$(shell ls -1 */Skargo.toml))
-
-check-%:
-	cd $* && skargo check
-
 .PHONY: check
-check: $(CHECK_TARGETS)
+check:
+	find * -name Skargo.toml -exec sh -c 'bin/cd_sh $$(dirname {}) "skargo check"' \;
 
 .PHONY: clean
 clean:
@@ -233,4 +229,3 @@ test-wasm:
 
 .PHONY: test-skfs
 test-skfs: sktest-prelude
-
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,10 @@ update-js-deps:
 check:
 	find * -name Skargo.toml -exec sh -c 'bin/cd_sh $$(dirname {}) "skargo check"' \;
 
+.PHONY: check-ts
+check-ts:
+	bin/check-ts.sh
+
 .PHONY: clean
 clean:
 	rm -Rf build
--- a/bin/cd_sh
+++ b/bin/cd_sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# Script for use from Makefiles to change to a directory and execute a command
+# For example: cd_sh subdir "cmd1 arg1 && cmd2 arg2"
+
+set -e
+
+dir=$1
+shift
+cmd=( "$@" )
+
+echo "make: Entering directory '$dir'"
+cd "$dir"
+bash -c "${cmd[@]}"
+echo "make: Leaving directory '$dir'"
--- a/bin/check-ts.sh
+++ b/bin/check-ts.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Script to run the type-checker in all directories in the npm workspaces list
+
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO="$SCRIPT_DIR/.."
+
+cd "$REPO"
+DIRS=$( jq --raw-output ".workspaces[]" package.json )
+
+for dir in $DIRS
+do
+    "$SCRIPT_DIR"/cd_sh "$dir" "[ ! -f eslint.config.js ] || npx tsc"
+done
--- a/skiplang/prelude/ts/src/sk_mem_utils.ts
+++ b/skiplang/prelude/ts/src/sk_mem_utils.ts
@@ -72,6 +72,7 @@ export class MemFS implements FileSystem {
   private fileDescrs: Map<string, number>;
   private fileDescrNbr: int;
   private files: Array<File>;
+  // invariant: range(fileDescrs) ⊆ domain(files)
 
   constructor() {
     this.fileDescrs = new Map();
@@ -87,9 +88,9 @@ export class MemFS implements FileSystem {
     let existing = this.fileDescrs.get(filename);
     if (existing != undefined) {
       if (options.create && options.create_new) {
-        throw new Error("The file '" + filename + "' already exist");
+        throw new Error("The file '" + filename + "' already exists");
       }
-      this.files[existing].open(options);
+      this.files[existing]!.open(options); // invariant
       return existing;
     }
     let fd = this.fileDescrNbr;
@@ -108,14 +109,18 @@ export class MemFS implements FileSystem {
   }
 
   writeToFile(fd: int, content: string) {
-    this.files[fd].write(content);
+    let file = this.files[fd];
+    if (file == undefined) {
+      throw new Error("Invalid file descriptor " + fd);
+    }
+    file.write(content);
   }
 
   appendToFile(filename: string, content: string) {
     let fd = this.fileDescrs.get(filename);
     let file: File;
     if (fd != undefined) {
-      file = this.files[fd];
+      file = this.files[fd]!; // invariant
     } else {
       file = new File();
       let fd = this.fileDescrNbr;
@@ -129,15 +134,27 @@ export class MemFS implements FileSystem {
   }
 
   closeFile(fd: int) {
-    this.files[fd].close();
+    let file = this.files[fd];
+    if (file == undefined) {
+      throw new Error("Invalid file descriptor " + fd);
+    }
+    file.close();
   }
 
   write(fd: int, content: string) {
-    return this.files[fd].write(content);
+    let file = this.files[fd];
+    if (file == undefined) {
+      throw new Error("Invalide file descriptor " + fd);
+    }
+    return file.write(content);
   }
 
   read(fd: int, len: int): string | null {
-    return this.files[fd].read(len) ?? null;
+    let file = this.files[fd];
+    if (file == undefined) {
+      throw new Error("Invalide file descriptor " + fd);
+    }
+    return file.read(len) ?? null;
   }
 }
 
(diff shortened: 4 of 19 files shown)


Makefile
```diff
@@ -65,13 +65,13 @@ build/init.sql: sql/privacy/init.sql
...
 
-CHECK_TARGETS=$(patsubst %/Skargo.toml,check-%,$(shell ls -1 */Skargo.toml))
-
-check-%:
-	cd $* && skargo check
-
 .PHONY: check
-check: $(CHECK_TARGETS)
+check:
+	find * -name Skargo.toml -exec sh -c 'bin/cd_sh $$(dirname {}) "skargo check"' \;
```
beauby, line 70: This reminds me that we need a "workspace" notion in Skargo (like in Cargo and npm), which would allow a simple `skargo check` within the workspace root to check all packages.

bin/check-ts.sh
```diff
@@ -0,0 +1,16 @@
...
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO="$SCRIPT_DIR/.."
+
+cd "$REPO"
+DIRS=$( jq --raw-output ".workspaces[]" package.json )
+
+for dir in $DIRS
+do
+    "$SCRIPT_DIR"/cd_sh "$dir" "[ ! -f eslint.config.js ] || npx tsc"
+done
```
beauby, line 16: All this should be equivalent to calling `npm run build` from the root directory, no?

skiplang/prelude/ts/src/sk_runtime.ts
```diff
@@ -123,11 +123,11 @@ class LinksImpl implements Links {
 
     this.SKIP_js_get_argc = () => utils.args.length;
     this.SKIP_js_get_argn = (index: int) =>
-      utils.exportString(utils.args[index]);
+      utils.exportString(utils.args[index]!);
```
beauby, line 126: Is the `index` guaranteed to be within bounds here?

skiplang/prelude/ts/src/sk_runtime.ts
```diff
@@ -123,11 +123,11 @@ class LinksImpl implements Links {
...
     this.SKIP_js_get_envc = () =>
       this.env && this.env.environment ? this.env.environment.length : 0;
     this.SKIP_js_get_envn = (index: int) =>
-      utils.exportString(this.env!.environment[index]);
+      utils.exportString(this.env!.environment[index]!);
```
beauby, line 130: Same question as above.

skiplang/prelude/ts/src/sk_runtime.ts
```diff
@@ -138,7 +138,7 @@ class LinksImpl implements Links {
       return this.lineBuffer.length;
     };
     this.SKIP_read_line_get = (i: int) => {
-      return this.lineBuffer[i];
+      return this.lineBuffer[i]!;
```
beauby, line 141: Same here.

skipruntime-ts/Makefile
```diff
@@ -2,31 +2,27 @@ SHELL := /bin/bash
...
 
 .PHONY: build
 build:
-	cd .. && npm install && npm run build
+	../bin/cd_sh .. "npm install && npm run build"
 
 .PHONY: bunbuild
 bunbuild:
-	cd .. && bun install && bun run build
+	../bin/cd_sh .. "bun install && npm run build"
```
beauby, line 25: Was `bun` choking on the `bun run build` command?

bin/cd_sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Script for use from Makefiles to change to a directory and execute a command
+# For example: cd_sh subdir "cmd1 arg1 && cmd2 arg2"
```
beauby, line 4: Is there a benefit compared to doing
```
(cd subdir && cmd1 arg1 && cmd2 arg2)
```
directly?

beauby (commented):
Much appreciated!

skiplang/prelude/ts/src/sk_mem_utils.ts
```diff
@@ -129,15 +134,27 @@ export class MemFS implements FileSystem {
...
+      throw new Error("Invalid file descriptor " + fd);
+    }
+    file.close();
   }
 
   write(fd: int, content: string) {
-    return this.files[fd].write(content);
+    let file = this.files[fd];
+    if (file == undefined) {
+      throw new Error("Invalide file descriptor " + fd);
```
mbouaziz, line 147: :fr:

skiplang/prelude/ts/src/sk_runtime.ts
mbouaziz: `SKIP_read_line_get` seems to be used only in `SKIP_read_line` and `SKIP_read_to_end` which are not used...

skipruntime-ts/core/src/skipruntime_helpers.ts
```diff
@@ -144,7 +144,7 @@ function toId(
 ): string {
   const strparams: string[] = [];
   for (const key of Object.keys(params).sort()) {
-    strparams.push(`${key}:${params[key].toString()}`);
+    strparams.push(`${key}:${params[key]!.toString()}`);
```
mbouaziz, line 147: This should have more or less the same behaviour:
```ts
const strparams = Object.entries(params).map(([key, value]) => `${key}:${value.toString()}`).sort();
```

skipruntime-ts/examples/sheet.ts
```diff
@@ -85,7 +85,7 @@ class Service implements SkipService {
     context: Context,
     inputCollections: Record<string, EagerCollection<string, TJSON>>,
   ): Record<string, EagerCollection<TJSON, TJSON>> {
-    const cells = inputCollections["cells"];
+    const cells = inputCollections["cells"]!;
```
mbouaziz, line 88: Change type of parameter `inputCollections` to `{ cells: ... }`?

skipruntime-ts/examples/utils.ts
```diff
@@ -33,7 +33,11 @@ class SkipHttpAccessV1 {
     ports: number[] = [3587],
     private creds: Protocol.Creds,
   ) {
-    this.defaultPort = ports[0];
+    let defaultPort = ports[0];
+    if (defaultPort == undefined) {
+      throw new Error("At least one port is required.");
```
mbouaziz, line 38: Given that there is a default value for this parameter, should we use it instead?

Also, it seems weird to have an optional parameter before a mandatory one

sql/ts/src/skdb-cli.ts
```diff
@@ -213,21 +213,21 @@ if (values.dev) {
...
+      hostAndPort[0]!,
       3586,
       values.db as string,
     );
   } else {
     dbCreds = await getCredsFromDevServer(
-      hostAndPort[0],
-      parseInt(hostAndPort[1]),
+      hostAndPort[0]!,
+      parseInt(hostAndPort[1]!),
```
mbouaziz, line 230: Should we use comparison with `undefined` for such cases?
```ts
const [host, port] = schemeAndRest[schemeAndRest.length - 1]!.split(":", 2);
if (host === undefined) {
  // same error
} else {
  dbCreds = await getCredsFromDevServer(
    host,
    port ? parseInt(port) : 3586,
    ...
```

Makefile
jberdine: Yes! And hopefully it would know the interdependencies so that it could do each dir only once, and maybe even in parallel.

bin/check-ts.sh
jberdine: Almost, but not quite. The issue is that `npm run build` does invoke `tsc` in each directory, but then `tsc` reports errors using paths that are relative to some directory below the one in which `npm run build` was executed, making finding the error locations hard for humans like me and impossible for tools. The thing this script does that `npm run build` does not is in `cd_sh`, which reports the usual entering/exiting directory messages before and after running commands in other directories, making it possible to jump to the errors reported by those commands.

jberdine: I would be happy to have a way to get npm or tsc to itself report errors using paths relative to the root, but could not find a way. I also didn't manage to insert a script wrapping tsc with directory reporting in a way that npm would actually call it.

skipruntime-ts/Makefile
jberdine: No, it just would itself call npm, so this change from bun to npm is just removing one indirection.

bin/cd_sh
jberdine: See the comment above, the benefit is only the entering/exiting messages

skiplang/prelude/ts/src/sk_mem_utils.ts
jberdine: I'm being assimilated it seems

skiplang/prelude/ts/src/sk_runtime.ts
jberdine: Removed in https://github.com/SkipLabs/skip/pull/409

Makefile
jberdine: @beauby another thing to consider regarding a workspace feature for skargo is that this PR's shell command doesn't actually work since what can and does happen is that we `cd` to a dir, run `skargo check`, skargo realizes that there is a dep so it tries to `skargo check` it, and in the case that there are errors while checking the dep, they are reported with locations relative to the dep's own Skargo.toml file, not relative to the directory from which the user ran `skargo check`. A workspace feature for skargo will need to teach the compiler to report errors relative to the workspace root.

skiplang/prelude/ts/src/sk_runtime.ts
jberdine: Fixed in new version.

skiplang/prelude/ts/src/sk_runtime.ts
jberdine: Fixed in new version.

skiplang/prelude/ts/src/sk_runtime.ts
jberdine: Not removed, but fixed in new version.

skipruntime-ts/core/src/skipruntime_helpers.ts
jberdine: Done in new version.

skipruntime-ts/examples/sheet.ts
jberdine: Done in new version.

skipruntime-ts/examples/utils.ts
jberdine: Done in new version. And yeah, the optional arg just wasn't optional.

sql/ts/src/skdb-cli.ts
jberdine: Done in new version.

Makefile
beauby: Yeah, I have a fix for this, it should be out soon.

PR #392, opened 2024-10-09
Update .gitignore to manage changes to skiplang directory

skiplabsdaniel:


beauby:
Good catch. I suggest to move the subdirectory-specific ignore directives in their own `.gitignore` file, i.e. create a `.gitignore` in `skiplang/compiler` with contents
```
bootstrap/*.a
bootstrap/*.ll
stage*/
tests/**/*.bin
tests/**/*.bin.ll
```
amd similar for `skiplang/prelude`.

That way, we won't be bothered if we ever move directories again.

PR #396, opened 2024-10-10
Fix Dockerfile after skiplang directory change

jberdine:
And appease the Dockerfile linter.

### DIFF
--- a/sql/Dockerfile
+++ b/sql/Dockerfile
@@ -1,8 +1,8 @@
-FROM skiplabs/skdb-base as skdb
+FROM skiplabs/skdb-base AS skdb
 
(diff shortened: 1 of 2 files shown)


sql/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM skiplabs/skdb-base as skdb
+FROM skiplabs/skdb-base AS skdb
```
beauby, line 1: Tangential: We should probably introduce a base image `skiplabs/skiplang-base` in `skiplang/Dockerfile`, which contains only an installed skiplang toolchain.

beauby (approved):

PR #398, opened 2024-10-10
[skipruntime] Fix SkipRuntime_Runtime__subscribe return type

skiplabsdaniel:


beauby (approved):

skipruntime-ts/core/native/src/Extern.sk
```diff
@@ -669,8 +669,8 @@ fun subscribeOfRuntime(
       reactiveAuth,
     )
   }) match {
-  | Success(id) -> SKJSON.CJString(id.toString())
-  | Failure(err) -> SKJSON.CJFloat(getErrorHdl(err))
+  | Success(id) -> id
+  | Failure(err) -> -getErrorHdl(err).toInt()
```
beauby, line 673: Why is it negative now?

skiplabsdaniel: Before the id was converted to String so if we get a number it could only be an error.
Now the error and the session are Int so to distinguish then the error are negative.

beauby: Ok, but call sites do not seem to be modified in this PR?

skiplabsdaniel: This is why it's a fix the call sites consider already the return value is a bigint, probably something introduced during a merge.

PR #400, opened 2024-10-10
Rename ValueMapper, add ManyToOneMapper and use it to simplify examples

jberdine:


skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -69,6 +72,31 @@ export abstract class ValueMapper<
   }
 }
 
+/**
+ * A specialized form of `Mapper` which maps values to values, reusing the
+ * input collection's key structure in the output collection. Use this form
+ * when the output values for a particular key need to consider all the
+ * input values associated to the key.
```
beauby: The last part ("when the output values for a particular key need to consider all the input values associated to the key") is a bit unclear, but maybe it's just me. Maybe stress that the output values do *not* need to consider input values for other keys?

beauby (approved):

bennostein:
Looks good to me, but I notice that all of the examples using `ValuesMapper` reduce the input values to a singleton output value -- I wonder if that's the more common special case here?

My (weakly-held) intuition is that one-to-one (ValueMapper) and many-to-one (i.e. as ValuesMapper is being applied here to produce a singleton) mappers should be common, but I can't think off the top of my head of many-to-many use cases which use the `ValuesMapper` type as defined but wouldn't just want the fully-general Mapper.  WDYT?

jberdine:
@bennostein that's a good point and I agree, I will revise. @beauby interesting, I was not thinking about possibly relying on values of other keys, as that just doesn't fit my default understanding of a `map`. I will try reworking the docs and maybe class names based on Benno's "one-to-one" and "many-to-one" terminology as that seems to be short and clear.

skipruntime-ts/core/src/skipruntime_api.ts
beauby: Actually scratch that, I was confused. I guess the confusing part of the sentence is that it suggests the alternative is for the output values to need to consider only _some_ of the input values associated to the key. If this is the case, maybe mention something like ("unlike with a FooMapper that requires only some").

jberdine: Hmm, I sense that there is still some confusion here, and digging in is probably worthwhile as users will likely hit it too. The crucial distinction is that one type of mapper is meant for the case where when a collection has multiple values for a key it is just a representation for a set of key-value pairs where the keys need not be unique, and the other type of mapper is meant for the case where the value associated with a key is already a (mathematical) set. So if you have a collection mapping keys to numbers, then the second type of mapper can be used to transform a <key>-<numbers> to <key>-<the sum of the numbers associated to the key>, while the first type of mapper cannot. I need to find a better explanation.

jberdine:
@bennostein I revised this and used the one-to-one and many-to-one terminology you mentioned. Do you think the current iteration is clear?

jberdine:
This is still open despite github thinking otherwise.

bennostein:
Looks good to me!  Following this nomenclature might have us rename the fully-general `Mapper` to `ManyToManyMapper` too, but I'm not sure if that's a good idea or not

jberdine:
The current `Mapper` would still be more general than a `ManyToManyMapper` since it can add/remove/change the keys too.

jberdine:
Merged manually

PR #403, opened 2024-10-11
[skiprutime] Renaming

skiplabsdaniel:


skipruntime-ts/core/native/src/Extern.sk
```diff
@@ -140,19 +140,19 @@ class ExternLazyCompute(eptr: SKStore.ExternalPointer) extends LazyCompute {
 
 /************  ExternalSupplier ****************/
 
-@cpp_extern("SkipRuntime_ExternalSupplier__link")
+@cpp_extern("SkipRuntime_ExternalSupplier__subscribe")
 @debug
-native fun linkOfExternalSupplier(
+native fun subscribeOfExternalSupplier(
```
beauby, line 145:
```suggestion
native fun subscribeToExternalSupplier(
```

skiplabsdaniel: As it response to a convention, tt means subscribe of ExternalSupplier, and as all the other functions are named in that way. I prefer to not make that change.
If the naming convention is not convenient we can change it.

jberdine: What about just `subscribeExternalSupplier` like how `deleteExternalSupplier` is named? The thing that trips me up about `subscribeOfExternalSupplier` is that grammatically it makes it sound like `subscribe` is a noun, and a thing belonging to `ExternalSupplier`. Leaving out `of` makes it sound like an action, which agrees with returning `void`.

jberdine: (I mean, leaving out `of` as a convention, so they are all named consistently.)

PR #404, opened 2024-10-11
[skipruntime] Add some API doc comments

skiplabsdaniel:


skipruntime-ts/core/src/skipruntime_api.ts
```diff
@@ -287,72 +295,155 @@ export type ReactiveResponse = {
...
     reactiveAuth?: Uint8Array,
   ): SubscriptionID;
 
+  /**
+   * Unsubscribe to a reactive ressource according a given subcription identifier
+   * @param id - the subcription identifier
+   */
   unsubscribe(id: SubscriptionID): void;
 
   // WRITE
```
jberdine: I'd keep both READ and WRITE comments, or neither

jberdine (approved):
Nice! I can do a little copy-editing, but easier done as separate changes than through code review.

PR #405, opened 2024-10-11
[skiprutime-server] Respond with 404 error on UnknownCollectionError

### LINKED ISSUE #395: [skipruntime/server] We should respond with `404 NOT FOUND` for `UnknownCollectionError`.
We should respond with `404 NOT FOUND` for `UnknownCollectionError`.

_Originally posted by @beauby in https://github.com/SkipLabs/skdb/pull/389#discussion_r1795228445_

skiplabsdaniel:
Fix #395

### DIFF
--- a/skipruntime-ts/server/src/rest.ts
+++ b/skipruntime-ts/server/src/rest.ts
@@ -23,7 +23,7 @@ export function createRESTServer(runtime: SkipRuntime): express.Express {
       );
       const [name, value] = reactiveResponseHeader(data);
       res.set(name, value);
-      res.status(200).json({});
+      res.status(200).json();
     } catch (e: unknown) {
       res.status(500).json(e instanceof Error ? e.message : e);
     }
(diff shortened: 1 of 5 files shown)


skipruntime-ts/server/src/rest.ts
```diff
@@ -77,7 +77,7 @@ export function createRESTServer(runtime: SkipRuntime): express.Express {
       res.status(200).json({});
     } catch (e: unknown) {
       if (e instanceof UnknownCollectionError) {
-        res.status(400).json("Bad request");
+        res.status(404).json("Not Found");
```
beauby: We should probably omit the body as a `404` response status implies the "Not found" information.

skipruntime-ts/core/src/skip-runtime.ts
```diff
@@ -22,7 +22,10 @@ export type {
   SubscriptionID,
 } from "./skipruntime_api.js";
 
-export { UnknownCollectionError } from "./skipruntime_errors.js";
+export {
+  UnknownCollectionError,
+  AccessRefusedError,
```
beauby: Is this needed?

skipruntime-ts/server/src/rest.ts
```diff
@@ -23,7 +23,7 @@ export function createRESTServer(runtime: SkipRuntime): express.Express {
       );
       const [name, value] = reactiveResponseHeader(data);
       res.set(name, value);
-      res.status(200).json({});
+      res.status(200).json();
```
beauby, line 26: I think you can simply remove the .json() call

skiplabsdaniel: Yes but I need sendStatus instead of status.

PR #406, opened 2024-10-11
Renamed http into externalDepartureAPI in the departure example

pikatchu:
Also, removed the const "departure" that was only used once and which had a confusing name, since many things are also named departure in this example (including the name of the example itself).

### DIFF
--- a/skipruntime-ts/examples/departures.ts
+++ b/skipruntime-ts/examples/departures.ts
@@ -45,7 +45,7 @@ class DeparturesResource implements Resource {
     return context.useExternalResource(
       "externalDeparturesAPI",
       "departures",
-      params
+      params,
     );
   }
 }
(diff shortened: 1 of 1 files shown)


skipruntime-ts/examples/departures.ts
```diff
@@ -48,7 +42,11 @@ class DeparturesResource implements Resource {
       resettlement: get("resettlement", "NOR,USA"),
     };
 
-    return context.useExternalResource("http", "departures", params);
+    return context.useExternalResource(
+      "externalDeparturesAPI",
+      "departures",
+      params
+    );
```
jberdine: Not specific to this diff, but I wonder if `useExternalResource` would be clearer if we defined an object type for its arguments so that calls would look something like:
```suggestion
    return context.useExternalResource({
      supplier: "externalDeparturesAPI",
      resource: "departures",
      params: params
    });
```

jberdine (approved):

skipruntime-ts/examples/departures.ts
pikatchu: Good point: https://github.com/SkipLabs/skip/pull/408

PR #407, opened 2024-10-11
Adds the constructor signature to the Resource interface.

pikatchu:
This doesn't change anything at runtime, but helps with discoverability. When somebody is trying to define a resource, what will happen is that they will sometimes need parameters. So for example, if we hit the endpoint:
http://myapi.foo.com/myresource?name=julien.  In this example, we are defining the resource myresource and we have { "name": "julien" } as a arguments.

The arguments are passed to us through the constructor. But since the constructor signature was not part of the interface, it was hard to discover ... This diff fixes that.

### DIFF
--- a/skipruntime-ts/core/src/skipruntime_api.ts
+++ b/skipruntime-ts/core/src/skipruntime_api.ts
@@ -353,6 +353,8 @@ export interface ExternalSupplier {
 }
 
 export interface Resource {
+  constructor(params: Record<string, string>);
+
   reactiveCompute(
     context: Context,
     collections: Record<string, EagerCollection<TJSON, TJSON>>,


skiplabsdaniel:
A constructor signature cannot be added to an interface in TypeScript.
If you implements It, it requires a the definition of ["contructor"] = (parameter ) => {..}
We can force to manage (think about) parameters with an abstract class.

pikatchu:
Ah ok. Thanks!

beauby:
So should we make `Resource` a `class` instead of an `interface`?

mbouaziz:
Or would that work?
```ts
export type Resource = {
  new (params: Record<string, string>): Resource;
...
```

Edit: answer: no

pikatchu:
So apparently there is no way to 

> Or would that work?
> 
> ```ts
> export type Resource = {
>   new (params: Record<string, string>): Resource;
> ...
> ```
> 
> Edit: answer: no

Can you elaborate? I was about to try this too. Why doesn't it work?

pikatchu:
Ok looks like what I am trying to do is just not possible in TS ...

jberdine:
FMI, was the possibility of making `Resource` a `class` also not going to work?

PR #408, opened 2024-10-11
Pass object to use external resource

pikatchu:
Refactor useExternalResource to take an object as a parameter.
    
The idea is that this:
```
useExternalResource({ supplier: "foo", resource: "rname", params: ... })
```
    
Is more readable than:
```
useExternalResource("foo", "rname", ...)
```


jberdine (approved):

skiplabsdaniel (approved):

mbouaziz:
I feel like non-object would be much less of an issue if the supplier was a well-typed object rather than a string.
I haven't dug in the new API yet to suggest something specific

PR #409, opened 2024-10-11
Remove SKIP_read_line from ts runtime

jberdine:


beauby:
Nice. Have we reached a state where we can fully get rid of it? (I.e. in the runtime (in `stdlib.c`/`runtime64_specific.cpp`) -- it seems only `SKIP_read_line()` calls `SKIP_read_line_fill()`, and the former is not being called anymore.)

jberdine:
`SKIP_read_line` is available as `System.read_line`, and is used by skjson and skdb. Maybe those should be converted to use `IO.BufferedReader`?

beauby:
Oh, right, I forgot about the good old non-greppable `@cpp_extern`...

Yeah I think it would be worth it to use only one buffered IO system across the code base.

jberdine:
Let me postpone removing the native version to another PR since I want this one first to get out of the way of ts type-checking changes.

mbouaziz:
Ah, sorry I've been fooled by the `@cpp_extern` as well

jberdine:
The tests are unhappy with errors such as:
```
  1) [client_nodejs] › node.play.ts:6:3 › Create table if not exists ───────────────────────────────

    LinkError: WebAssembly.instantiate(): Import #27 module="env" function="SKIP_read_line_fill": function import requires a callable
```
This seems to indicate that the playwright run tests still have some reference to the removed `SKIP_read_line_fill` function, but I haven't been able to figure out how. Maybe there is something about the wasm initialization that I'm overlooking or confused about. Is it immediately obvious to anyone else? @skiplabsdaniel or @beauby maybe?

beauby:
> This seems to indicate that the playwright run tests still have some reference to the removed `SKIP_read_line_fill` function, but I haven't been able to figure out how. Maybe there is something about the wasm initialization that I'm overlooking or confused about. Is it immediately obvious to anyone else? @skiplabsdaniel or @beauby maybe?

In `sql/src/Main.sk` there are a bunch of calls to `read_line()`, I suspect that's the reason.

jberdine:
Gah, right in front of me and I didn't realize it, thanks!

PR #412, opened 2024-10-14
Improve check targets

jberdine:
Fix the native check targets, and add one for the ts sources.

### DIFF
--- a/bin/check-ts.sh
+++ b/bin/check-ts.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Script to run the type-checker in all directories in the npm workspaces list
+
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO="$SCRIPT_DIR/.."
+
+cd "$REPO"
+DIRS=$( jq --raw-output ".workspaces[]" package.json )
+
+for dir in $DIRS
+do
+    "$SCRIPT_DIR"/cd_sh "$dir" "npm run build --if-present"
+done
--- a/Makefile
+++ b/Makefile
@@ -65,13 +65,9 @@ build/init.sql: sql/privacy/init.sql
 update-js-deps:
 	find . -name node_modules -not -prune -or -name target -not -prune -or -name package.json -exec sh -c 'cd $$(dirname "$$0"); bun update --latest' {} \;
 
-CHECK_TARGETS=$(patsubst %/Skargo.toml,check-%,$(shell ls -1 */Skargo.toml))
-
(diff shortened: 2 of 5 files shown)


jberdine:
Split off from https://github.com/SkipLabs/skip/pull/391 as this is independent and I need it for several branches.

bin/check-ts.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Script to run the type-checker in all directories in the npm workspaces list
+
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO="$SCRIPT_DIR/.."
+
+cd "$REPO"
+DIRS=$( jq --raw-output ".workspaces[]" package.json )
```
beauby, line 11: A bit sad to introduce a dependency on `jq` but not the end of the world, especially since it's used only for checking.

bin/check-ts.sh
```diff
@@ -0,0 +1,16 @@
...
+
+for dir in $DIRS
+do
+    "$SCRIPT_DIR"/cd_sh "$dir" "[ ! -f eslint.config.js ] || npx tsc"
```
beauby: Could this be replaced with a simple `npm run build` in each directory?

beauby (approved):

bin/check-ts.sh
jberdine: I don't know how to do that and still get the printing before and after. Am I missing something?

beauby: No, I meant:
```suggestion
    "$SCRIPT_DIR"/cd_sh "$dir" "npm run build"
```

bin/check-ts.sh
jberdine: I guess you're already sad due to https://github.com/SkipLabs/skip/blob/a95d7d8a4fd7a53affe29da584bdd7c03253f43c/sql/test/params/json_args_to_sqlite_params.sh#L11, but also just for testing/checking.

bin/check-ts.sh
jberdine: Ah, ok. `npm run build` fails in the eslint-config and tsconfig directories. There is probably a way to change their package.json files to make it pass. Would that be better?

beauby: Either that or do `npm run build --if-present` which succeeds when the `build` script is not defined.

jberdine: Ok the --if-present approach works, updated to do that. This does result in `make check-ts` actually build stuff with skargo, which I don't know if that is necessary for just checking the ts sources, but once it is done once it isn't significantly slower.

PR #413, opened 2024-10-14
[skc] Remove dead code.

beauby:


jberdine (approved):
LGTM but I think the code you called "Hack-specific" isn't about Hack, it is just a hack, and it is now dead since `memoized` is gone.

beauby:
> LGTM but I think the code you called "Hack-specific" isn't about Hack, it is just a hack, and it is now dead since `memoized` is gone.

You are indeed right, I will update the commit message.

PR #414, opened 2024-10-14
Print error message when client cannot connect to the websocket.

pikatchu:
I also added `import { WebSocket } from 'ws'` for older version of node that don't have it by default.

### DIFF
--- a/skipruntime-ts/client/src/client.ts
+++ b/skipruntime-ts/client/src/client.ts
@@ -40,16 +40,16 @@ async function createWs(uri: string): Promise<WebSocket> {
   return new Promise((resolve, reject) => {
     const socket = new WebSocket(uri);
     socket.binaryType = "arraybuffer";
-    socket.onclose = (_event) => {
+    socket.onclose = (_event: any) => {
       reject(new Error("Socket closed before open"));
     };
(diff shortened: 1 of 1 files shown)


skipruntime-ts/client/src/client.ts
```diff
@@ -28,6 +28,8 @@ export type {
   Creds,
 };
 
+import WebSocket from "ws";
```
beauby: The reason I had removed it and relied on the (recent) node implementation available in the global scope, was so that the client could be used in a browser. If we import `ws` (which is node-specific), the client cannot be used directly in the browser anymore.

pikatchu: Is it worth it for this specific example?

jberdine: Maybe I'm confused, but I thought this was the implementation of our generic client package, not an example.

beauby:
Also: [Node 22 will become LTS in October](https://nodejs.org/fr/blog/announcements/v22-release-announce)

jberdine:
This seemed stuck so I rebased it and trimmed out the `import WebSocket`

jberdine:
merged

PR #339, opened 2024-09-25
Build insight UI

gregsexton:
This renders a debug trace given a static JSON file that is generated
by SKStore.

PR #415, opened 2024-10-14
Insight Backend

gregsexton:
This supplies the data as used by the UI (see PR #339).

There is a networking PR to connect them but the rebase is extremely painful, so I've carved this out separately.


mbouaziz:
Heads-up: I force-pushed this branch to move it off the pre-rewrite history. The repository history was rewritten (~2025-03) to purge the precompiled bootstrap binaries into submodules, which re-hashed every commit after `5cc20f746`. This PR was still based on that old history, so GitHub could not compute a real diff or merge. I replayed the branch's own commits onto the new-history twin of their original base — commit contents are byte-identical, only the base SHAs changed. It is mergeable against `main` again.

PR #416, opened 2024-10-16
Enable noUncheckedIndexedAccess

jberdine:
Enable noUncheckedIndexedAccess, and fix all the existing violations.

(Retry of https://github.com/SkipLabs/skip/pull/391 since github "helpfully"
closed it for me and refuses to reopen it.)

### DIFF
--- a/skiplang/prelude/ts/src/sk_mem_utils.ts
+++ b/skiplang/prelude/ts/src/sk_mem_utils.ts
@@ -72,6 +72,7 @@ export class MemFS implements FileSystem {
   private fileDescrs: Map<string, number>;
   private fileDescrNbr: int;
   private files: Array<File>;
+  // invariant: range(fileDescrs) ⊆ domain(files)
 
   constructor() {
     this.fileDescrs = new Map();
@@ -87,9 +88,9 @@ export class MemFS implements FileSystem {
     let existing = this.fileDescrs.get(filename);
     if (existing != undefined) {
       if (options.create && options.create_new) {
-        throw new Error("The file '" + filename + "' already exist");
+        throw new Error("The file '" + filename + "' already exists");
       }
-      this.files[existing].open(options);
+      this.files[existing]!.open(options); // invariant
       return existing;
     }
     let fd = this.fileDescrNbr;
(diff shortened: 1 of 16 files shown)


jberdine:
In the previous version I misunderstood the typescript docs: when they state that the `!` operator _asserts_ that the argument is non-null, what they actually mean is _assumes_, not _asserts_. Thanks TS. So this version changes the uses of `!` to actually check for errors.

skiplang/prelude/ts/src/sk_mem_utils.ts
```diff
@@ -108,14 +109,18 @@ export class MemFS implements FileSystem {
   }
 
   writeToFile(fd: int, content: string) {
-    this.files[fd].write(content);
+    let file = this.files[fd];
+    if (file == undefined) {
```
beauby: Probably `===` instead of `==` is better here but it shouldn't matter much.

beauby (approved):

skiplang/prelude/ts/src/sk_mem_utils.ts
jberdine: done now

jberdine: This made me think a bit more about the various notions of equality, and I changed this to use `??` instead of either `== undefined` or `=== undefined`. My understanding is that `e ?? fail(msg)` evaluates to `e` if it is neither `null` nor `undefined` and otherwise evaluates `fail(msg)`. Do you think that is an appropriate semantics here? If not, it is probably not appropriate in all the other places where `?? fail` is used in this PR.

skiplang/prelude/ts/src/sk_types.ts
```diff
@@ -52,6 +52,10 @@ class SkError extends Error {
   }
 }
 
+export function fail(msg: string): never {
+  throw new Error(msg);
+}
+
```
beauby: This looks suspicious: `Error` is a native JS class, so the `fail()` function doesn't have much to do with skiplang's standard library.

skiplang/prelude/ts/src/sk_mem_utils.ts
beauby: Yes this works as well (only caveat would be for places where `null` is a legit value).

skiplang/prelude/ts/src/sk_types.ts
jberdine: Yeah, fair enough. I'll look for a better place. I use this in the prelude, the ts examples, and skdb ts code, so it isn't obvious to me where it should go. Really it is just a work-around for the fact that `throw new Error(...)` is not allowed as the RHS of `??` while a function call is. I think I could instead replace every call to `fail` with something like `(() => throw new Error(...))()`. WDYT? (Grr, languages that don't get expressions right.)

skiplang/prelude/ts/src/sk_mem_utils.ts
jberdine: Yes. I was concerned about arrays where the element type admits `null` and checked a bunch of them, but I should probably make a pass over them all.

jberdine:
Ok, the TS `??` operator is not very usable and now gone. I looked at all the occurrences and think the checks are good wrt undefined vs null.

beauby:
Looks good!

PR #421, opened 2024-10-18
Minor possibly-functional cleanups

jberdine:
To help code review, I have extracted a few small changes that might be
functional from upcoming changes to appease the type-checker and
linter. Some individual commit summaries have more info.

### DIFF
--- a/skiplang/prelude/ts/src/sk_worker.ts
+++ b/skiplang/prelude/ts/src/sk_worker.ts
@@ -7,7 +7,7 @@ export class Wrappable {
 }
 
 class UnmanagedMessage extends Error {
-  constructor(msg: any) {
+  constructor(msg?: string) {
     super(msg);
   }
 }
@@ -241,7 +241,7 @@ export class PromiseWorker {
     this.onMessage = (message: MessageEvent) => {
       let data = Message.asReturn(message.data ?? message);
       if (!data) {
-        throw new UnmanagedMessage(message);
+        throw new UnmanagedMessage(JSON.stringify(message));
       } else {
         let result = data.payload as Return;
         let callback = this.callbacks.get(asKey(data.id));
--- a/skiplang/prelude/ts/src/sk_worker.ts
+++ b/skiplang/prelude/ts/src/sk_worker.ts
@@ -208,7 +208,7 @@ export class PromiseWorker {
       const parameters = fn.parameters.map((p) => {
         if (typeof p == "function") {
           let subscriptionId = new MessageId(this.source, ++this.lastId);
-          let wfn = (result: Return) => p.apply(null, result.value);
+          let wfn = (result: Return) => void p.apply(null, result.value);
           let key = asKey(subscriptionId);
           this.subscriptions.set(key, wfn);
           subscribed.add(key);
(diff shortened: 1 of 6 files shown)


skiplang/prelude/ts/src/sk_node.ts
```diff
@@ -61,11 +61,16 @@ class Env implements Environment {
...
+      filename = url;
+    } else if (url && url instanceof URL && url.pathname) {
       filename = "./" + path.relative(cwd, url.pathname);
-      // @ts-ignore
-    } else if (url && url.default) {
-      // @ts-ignore
+    } else if (
+      url &&
+      // @ts-expect-error: Property 'default' does not exist on type 'URL'.
+      url.default
```
jberdine: It seems that `default` isn't in `string` or `URL`, so I don't know when this case is ever used.

skiplang/prelude/ts/src/sk_worker.ts
```diff
@@ -241,7 +241,7 @@ export class PromiseWorker {
     this.onMessage = (message: MessageEvent) => {
       let data = Message.asReturn(message.data ?? message);
       if (!data) {
-        throw new UnmanagedMessage(message);
+        throw new UnmanagedMessage(JSON.stringify(message));
```
jberdine, line 244: I'm not sure this is the best way to create a string from a MessageEvent. Any suggestions?

sql/ts/src/skdb_database.ts
```diff
@@ -13,6 +13,10 @@ import { SKDBTable } from "./skdb_util.js";
 import { SKDBGroupImpl } from "./skdb_group.js";
 import { connect } from "./skdb_orchestration.js";
 
+type DiffAcc = {
+  [key: string]: { since: bigint };
+};
```
jberdine: AFAIU this needs to be a named type definition. I'd be happy if anyone knows the syntax to inline this into the one occurrence.

skiplang/prelude/ts/src/sk_node.ts
beauby: It was apparently introduced in [this diff](https://github.com/SkipLabs/skip/commit/8867b216da79c46b254746dc9b10b20a6be944c1#diff-7eb38b23d135400fced06531156114887376359756631d803d3ece6e79e2cabeR70). @skiplabsdaniel any reasons?

sql/ts/src/skdb_database.ts
beauby: Does `const acc: { [key: string]: { since: bigint } } = {};` not work? What is the error?

beauby (approved):

skiplang/prelude/ts/src/sk_node.ts
skiplabsdaniel: It's because for vite we use the `await import ` pattern that returns a module with a default property that contains the URL in string format.
It's was a fix made in a bad place to manage that.
The import for Url pattern shall be correctly manage each time it's used. In `skipruntime_init.ts` it's badly managed the wasmURL function shall returns a `Promise<URL | string>` and the the `default` property as string of `await import("./libskip-runtime-ts.wasm?url")` that should be returned and not the await directly.

jberdine: Ok, thanks a lot for this context. I will not make these functional changes to this stuff in this PR as there are broader and more delicate interactions than I realized.

sql/ts/src/skdb_database.ts
jberdine: You're right, it works, my failure was a missing set of braces. :-/

PR #425, opened 2024-10-23
[skipruntime] Allow to wait for external resources

skiplabsdaniel:


skipruntime-ts/core/native/src/BaseTypes.sk
```diff
@@ -91,4 +91,20 @@ base class Service(
...
+value class ReactiveResponse(id: String, watermark: SKStore.Tick)
+
+value class Values(
+  values: Array<(SKJSON.CJSON, Array<SKJSON.CJSON>)>,
+  reactive: ReactiveResponse,
+)
+
+base class Request {
+  children =
+  | Identified(id: String)
```
bennostein:
```suggestion
  | Identifier(id: String)
```

skipruntime-ts/core/native/src/Extern.sk
```diff
@@ -295,6 +313,32 @@ class ExternResourceBuilder(
...
+native fun deleteChecker(mapper: UInt32): void;
+
+@export("SkipRuntime_createChecker")
+fun createChecker(executor: UInt32): ExternChecker {
+  ExternChecker(SKStore.ExternalPointer::create(executor, deleteChecker))
+}
+
+@export("SkipRuntime_createIdentified")
+fun createIdentified(request: String): Identified {
+  Identified(request)
```
bennostein:
```suggestion
@export("SkipRuntime_createIdentifier")
fun createIdentifier(request: String): Identifier {
  Identifier(request)
```

skipruntime-ts/core/src/skipruntime_remote.ts
```diff
@@ -115,19 +121,24 @@ export class ExternalSkipService implements ExternalSupplier {
...
+      update: (updates: Entry<TJSON, TJSON>[], isInit: boolean) => void;
+      error: (error: TJSON) => void;
+      loading: () => void;
+    },
     reactiveAuth?: Uint8Array,
   ): Promise<void> {
     const [client, creds] = await this.client!;
     const publicKey = new Uint8Array(await Protocol.exportKey(creds.publicKey));
     const reactive = await this.auth(resource, params, publicKey);
+    // TODO Manage Status
```
bennostein, line 137: What's the plan here?  Currently the added callbacks to handle error or loading status are just ignored?

skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -994,6 +1037,37 @@ class CollectionWriter<K extends TJSON, V extends TJSON> {
       throw this.refs.handles.deleteAsError(result);
     }
   }
+
+  loading(): void {
+    const todo = () => {
```
bennostein: rename this `todo` to `loading` or similar?

skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -994,6 +1037,37 @@ class CollectionWriter<K extends TJSON, V extends TJSON> {
...
+    let result: Handle<ErrorObject>;
+    if (needGC) result = this.refs.skjson.runWithGC(todo);
+    else result = todo();
+    if (result != 0) {
+      throw this.refs.handles.deleteAsError(result);
+    }
+  }
+
+  error(error: TJSON): void {
+    const todo = () => {
```
bennostein: rename this `todo` to `errorObj` or similar?

skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -994,6 +1037,37 @@ class CollectionWriter<K extends TJSON, V extends TJSON> {
...
+      return this.refs.fromWasm.SkipRuntime_CollectionWriter__loading(
+        this.refs.skjson.exportString(this.collection),
+      );
+    };
+    const needGC = this.refs.needGC();
+    let result: Handle<ErrorObject>;
+    if (needGC) result = this.refs.skjson.runWithGC(todo);
+    else result = todo();
+    if (result != 0) {
```
bennostein:
```suggestion
    const result: Handle<ErrorObject> = needGC ? this.refs.skjson.runWithGC(todo) : todo();
    if (result != 0) {
```

skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -994,6 +1037,37 @@ class CollectionWriter<K extends TJSON, V extends TJSON> {
...
+      return this.refs.fromWasm.SkipRuntime_CollectionWriter__error(
+        this.refs.skjson.exportString(this.collection),
+        this.refs.skjson.exportJSON(error),
+      );
+    };
+    const needGC = this.refs.needGC();
+    let result: Handle<ErrorObject>;
+    if (needGC) result = this.refs.skjson.runWithGC(todo);
+    else result = todo();
+    if (result != 0) {
```
bennostein:
```suggestion
    const result: Handle<ErrorObject> = needGC ? this.refs.skjson.runWithGC(todo) : todo();
    if (result != 0) {
```

skipruntime-ts/core/native/src/Extern.sk
```diff
@@ -295,6 +313,32 @@ class ExternResourceBuilder(
   }
 }
 
+/************  Checker ****************/
```
bennostein, line 316: I'm a bit confused about what these are doing and why -- would you explain briefly?

I see the OneChecker and AllChecker over in skipruntime_module.ts which make queries and check reject/resolve depending on the number of errors.  But this seems very heavy weight to set up all these interfaces and pass things across the FFI to do so, right?

And, with all this indirection I'm not positive, but it seems to me that the getOne call within OneChecker will itself create a new OneChecker -- where/how does that get deduplicated?

skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -1087,9 +1161,90 @@ export class SkipRuntimeFactory implements Shared {
...
 
-class SkipRuntimeImpl implements SkipRuntime {
+export type Values<K extends TJSON, V extends TJSON> = {
+  values: Entry<K, V>[];
+  reactive?: ReactiveResponse;
+};
+
+export type GetResult<T> = {
+  request?: string;
+  values: T;
```
bennostein:
```suggestion
  result: T;
```

The `.values.values` in the `getAll` helper in `runtime.spec.ts` seems likely a common pattern; this makes it more clear what those fields actually mean.

skipruntime-ts/core/test/runtime.spec.ts
```diff
@@ -13,10 +13,30 @@ import type {
...
+function getOne<V extends TJSON>(
+  runtime: SkipRuntime,
+  resource: string,
+  params: Record<string, string>,
+  key: string | number,
+  reactiveAuth?: Uint8Array,
+): V[] {
+  return runtime.getOne<V>(resource, params, key, reactiveAuth).values;
+}
+
```
bennostein: I'd prefer to drop these helper functions -- people may look at these as examples, and it'd be better to just use the actual API throughout

skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -1242,6 +1479,9 @@ class SkipRuntimeImpl implements SkipRuntime {
     }
   }
 
+  /**
+   * Close all the resource and shutdown the SkipService
+   */
   close(): void {
     const result = this.refs.skjson.runWithGC(() => {
       return this.refs.fromWasm.SkipRuntime_closeService();
```
bennostein, line 1538: What's the behaviour if I call other runtime functions after `close`?  That would be a bug of course but I don't see any explicit logic about it -- would be good to at least make a note in this comment.

skipruntime-ts/core/src/internals/skipruntime_module.ts
```diff
@@ -1189,6 +1409,14 @@ class SkipRuntimeImpl implements SkipRuntime {
...
+   * Subscribe to a reactive ressource according a given reactive response
+   * @param reactiveId - the reactive response collection
+   * @param since - the reactive response watermark
+   * @param f - the callback called on collection updates
+   * @param reactiveAuth The client user Skip session authentification corresponding to the reactive response
+   * @returns The subcription identifier
+   */
   subscribe<K extends TJSON, V extends TJSON>(
     reactiveId: string,
     since: Watermark,
```
bennostein: These two params are the unpacked "reactive response" referenced in the doc comment, right?

Why unpack them here?  Or, if it's better to unpack for some reason, why even bother with the `ReactiveResponse` type at all?

bennostein (commented):
First batch of comments.  Mostly little things, but a handful of questions/comments that could use responses.

I don't see any tests about the main bit of behaviour that I thought this was designed to address: the initial GET request always returning an empty collection when it depends on something external.  Can you explain what observable changes this PR makes from the user's perspective?  What guidance should we give on how to use this / handle loading of external resources?

I'll follow up with a commit to clean up the doc comments' English, but appreciate all of those!

skipruntime-ts/core/native/src/Extern.sk
skiplabsdaniel: No, the “OneChecker” doesn't itself create a “OneChecker”, the “Checker” calls the corresponding “get” with a request identifier in order to retrieve the corresponding state.
Perhaps this could be simplified. I did it this way because I didn't want to be too closely modeled on the way promises work. I'd rather wait until I've integrated other languages before optimizing this.
I started by managing resolve/reject directly, but in terms of type and FFI management, it was heavier than this version.

skipruntime-ts/core/src/skipruntime_remote.ts
skiplabsdaniel: There are still some unanswered questions:
- What does this mean for a Skip service chain (status propagation, error handling, etc.)?

skipruntime-ts/core/test/runtime.spec.ts
skiplabsdaniel: The tests essentially use internal mechanics, so if people can look at this as an example we'll certainly have to rewrite them to avoid misuses.

skiplabsdaniel (commented):
The general principle is to add a collection of statuses to each resource, with each external service involved in the resource adding its status to this table, along with “creation” and “modification” times.
 When a request is made, a “request” (for want of a better word) is created to mark that a request is being made at that time.
The various statuses are aggregated relative to the time of this request (all this to avoid unfortunate time sequences that could lead to infinite waiting).
A request can be made with a “Checker” (too, for want of a better word).
If a “Checker” is provided, if will be called at status change for the requester to check and do what is necessary.
If there is no “Checker” provided and there is a dependency being loaded a “request” identifier is returned to allow subsequent status request relative to that request.
This is the general case for having a notion of status for the resource throughout its life in the service.
 There are still some unanswered questions:
- What does this mean for a Skip service chain (status propagation, error handling, etc.)?

skipruntime-ts/core/src/internals/skipruntime_module.ts
skiplabsdaniel: Indeed, I asked myself the same question. No real reason to unpack.

PR #426, opened 2024-10-23
Reshuffle/rename things around SkipRuntime/Services

bennostein:
The main change here is that we're pulling the "SkipRuntime" interface out of the public API, as the intention from Daniel/Lucas is that those operations on resources are to be managed under the hood by the framework.

I also renamed it to `ServiceInstance` but am open to changing that -- my thinking is that it represents a running service, i.e. an instance created by passing a `SkipService` (which specifies the logic/structure of a reactive service) to `initService` (which actually spins the thing up)


skipruntime-ts/core/src/skipruntime_init.ts
```diff
@@ -31,62 +34,113 @@ async function wasmUrl(): Promise<URL> {
...
- * other references.
- *
- * The primary use for this function is to satisfy the requirement that all
- * parameters to Skip `Mapper` constructors must be deep-frozen: objects
- * that have not been constructed by Skip can be passed to `freeze()` before
- * passing them to a `Mapper` constructor.
- *
- * @param value - The object to deep-freeze.
- * @returns The same object that was passed in.
+ * An `ServiceInstance` is the result of `initService`
```
jberdine:
```suggestion
 * A `ServiceInstance` is the result of `initService`
```

jberdine (commented):
I like the `ServiceInstance` naming, much more informative than `SkipRuntime`.

I think that `freeze` should appear in the API next to the `Constant` interface and the `Param` type. If someone wants a general `Param`, then they will need to use the `Constant` alternative of `Param`, and _the_ way to get such a value is with the `freeze` function. So I think those all need to appear together in the API. Also, `freeze` is not really a "utility" since apart from reverse-engineering internal representation details, it is the only way to produce a `Constant` `Param`, while the other utils are non-essential but convenient functionality.

jberdine (approved):
Nice

PR #430, opened 2024-10-25
suppress typescript-eslint/no-this-alias

jberdine:


skiplang/prelude/ts/src/sk_worker.ts
```diff
@@ -201,6 +201,7 @@ export class PromiseWorker {
     this.source = ++sourcesLastId;
     this.callbacks = new Map();
     this.subscriptions = new Map();
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
```
jberdine, line 204: For my understanding, if the occurrences of `self` on lines 228 and 238 below were replaced with `this`, what would `this` resolve to? Would it be equivalent, or would `this` refer to one of the nested closures, or...?

skiplabsdaniel: It's here because later in the function uses to build the Promise this does not refers the the good object, but it's it's true that the function can probably be binded to this.

jberdine:
Included in https://github.com/SkipLabs/skip/pull/435

PR #434, opened 2024-10-28
Not strictly NFC cleanups toward being lint-clean

jberdine:


sql/ts/src/skdb_skdb.ts
```diff
@@ -318,7 +318,15 @@ class LinksImpl implements Links, ToWasm {
...
-      this.stream = stream as 0 | 1 | 2;
+      switch (stream) {
+        case 0:
+        case 1:
+        case 2:
+          this.stream = stream;
+          break;
+        default:
+          throw new Error(`Invalid stream ${stream}`);
+      }
```
jberdine: This breaks the tests. I'm investigating. `as` is really evil.

jberdine: Fixed now

sql/ts/src/skdb_skdb.ts
```diff
@@ -318,7 +318,16 @@ class LinksImpl implements Links, ToWasm {
...
+      switch (stream) {
+        case 0:
+        case 1:
+        case 2:
+        case 3:
+          this.stream = stream;
+          break;
+        default:
+          throw new Error(`Invalid stream ${stream}`);
+      }
```
jberdine, line 330: I made this change because the `as 0 | 1 | 2` was not ensured just from local code, and in fact the tests showed that `3` is also needed.

sql/ts/src/skdb_skdb.ts
```diff
@@ -207,13 +207,13 @@ class LinksImpl implements Links, ToWasm {
...
   private objectIdx: number;
-  private stream: 0 | 1 | 2;
+  private stream: 0 | 1 | 2 | 3;
   private object: Record<string, any>;
   private stdout_objects: [
     Record<string, any>[],
     Record<string, any>[],
     Record<string, any>[],
-    [] | [{ tick: number }],
+    Record<string, any>[],
```
jberdine, line 216: This change is needed for `SKIP_push_object` to type-check.

sql/ts/src/skdb_skdb.ts
```diff
@@ -472,7 +481,9 @@ class LinksImpl implements Links, ToWasm {
           const rows = new SKDBTable(...this.stdout_objects[0]);
           const added = new SKDBTable(...this.stdout_objects[1]);
           const removed = new SKDBTable(...this.stdout_objects[2]);
-          const tick = this.stdout_objects[3][0]?.tick;
+          const tick = (
+            this.stdout_objects[3][0] as { tick: number } | undefined
+          )?.tick;
```
jberdine, line 486: This change is required by the change to the type of `stdout_objects[3]` above. I'm not sure there isn't a better way to resolve the non-uniformity of the types of the different streams, but it was more involved to move the non-uniform code from here to `SKIP_push_object`.

jberdine (commented):
There are several commits in this PR that remove tests that the linter thinks are constant due to various types not allowing null/undefined/any other falsy value. I have looked at each case, but it would be good if someone was to make a second sanity check of those commits.

bennostein:
Only weirdness that jumps out from all of the removed tests for types that don't admit falsy values: `this.env` is treated as nullable in [686c13c](https://github.com/SkipLabs/skip/pull/434/commits/686c13cbc5bc3bb361d8e63123eac7a3c7b4844c) but non-nullable in [0468491](https://github.com/SkipLabs/skip/pull/434/commits/04684919c04e2d5d8f367306db334e9d78f40630) -- should maybe keep the nullability check in [0468491](https://github.com/SkipLabs/skip/pull/434/commits/04684919c04e2d5d8f367306db334e9d78f40630)?

All looks good otherwise!

jberdine:
Thanks for looking!

> Only weirdness that jumps out from all of the removed tests for types that don't admit falsy values: `this.env` is treated as nullable in [686c13c](https://github.com/SkipLabs/skip/pull/434/commits/686c13cbc5bc3bb361d8e63123eac7a3c7b4844c) but non-nullable in [0468491](https://github.com/SkipLabs/skip/pull/434/commits/04684919c04e2d5d8f367306db334e9d78f40630) -- should maybe keep the nullability check in [0468491](https://github.com/SkipLabs/skip/pull/434/commits/04684919c04e2d5d8f367306db334e9d78f40630)?

In sk_runtime ([686c13c](https://github.com/SkipLabs/skip/pull/434/commits/686c13cbc5bc3bb361d8e63123eac7a3c7b4844c)) `env` is declared:
```
  env: Environment | undefined;
```
while in sk_types ([0468491](https://github.com/SkipLabs/skip/pull/434/commits/04684919c04e2d5d8f367306db334e9d78f40630)) `env` is declared:
```
  private env: Environment;
```
which is the origin of that oddity AFAIU.

PR #438, opened 2024-10-29
Reorder args of reactiveCompute to allow omitting context

jberdine:
It seems to be common for code to not need the context argument of
reactiveCompute. This PR reorders the args so that when it is not needed, it
can be omitted.


bennostein (approved):
looks great!

jberdine:
Merged manually

PR #439, opened 2024-10-29
[nfc] Rename Skip REST service and fix getOne/getArray there

bennostein:
following up on convo with Julien earlier today


pikatchu:
Thanks!

PR #441, opened 2024-10-30
[docs] Update functions.md with some examples; add a generic CountMapper utility

bennostein:
I added the CountMapper utility as it came up while writing examples, and I figured it's a common enough thing that a prefab util version may be useful.


jberdine (commented):
Text LGTM but I wonder about defining count as a mapper rather than an accumulator. Will we lead people to write the less incremental code by doing that? And this makes me wonder about the naming convention, why have Sum, Min, and Max accumulators and then Count*Mapper*?

PR #443, opened 2024-10-30
[skiprutime] Prepare npm packages

skiplabsdaniel:


examples/hackernews/back-end/reactive_service/package.json
```diff
@@ -10,7 +10,7 @@
   "license": "MIT",
   "description": "",
   "dependencies": {
-    "@skipruntime/core": "^1.0.0",
+    "skip-wasm": "^1.0.0",
```
beauby, line 13:
```suggestion
    "@skipruntime/core": "npm:skip-wasm@^1.0.0",
```

This way, we don't need to change the downstream code.

skipruntime-ts/native/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@skipruntime/wasm",
+  "version": "1.0.0",
+  "type": "module",
+  "scripts": {
+    "build": "skargo build -r --target wasm32-unknown-unknown --lib --manifest-path=./Skargo.toml --out-dir=../wasm/dist/src/",
```
beauby, line 6: We could copy relevant `.ts` files from `std`/`skjson`/etc. here to avoid publishing a lot of broken packages (the `std`/`skjson`/etc. packages will be broken because they won't ship the wasm definitions, as everything gets compiled only into the "skipruntime" wasm files).

skipruntime-ts/wasm/package.json
```diff
@@ -1,28 +1,28 @@
...
     "clean": "rm -rf dist",
-    "lint": "eslint src/",
+    "lint": "eslint",
     "test": "mocha"
   },
   "devDependencies": {
     "@types/express": "^4.17.21",
     "@types/mocha": "^10.0.8",
     "@types/ws": "^8.5.12",
+    "@skipruntime/tests": "^1.0.0",
```
beauby, line 18: Why does this need to be a separate package?

skipruntime-ts/native/package.json
skiplabsdaniel: For me the package are not broken, they are just library packages with a bunch of definitions.
Copying the file implies lot of works in imports process, so, or we copy the files in each package they are used, so when a modification is done in std / skjson / etc need to be reported in each copy, or we have a sknpm like post process that manage all imports correctly for publishing. In both cases I don't see the win, in the current state all the chain works well and packages can be publish as is after tsc build.

skipruntime-ts/wasm/package.json
skiplabsdaniel: It's to allow to run the tests with skip-wasm and with skip-native.

examples/hackernews/back-end/reactive_service/package.json
skiplabsdaniel: This does not seem to work in the current development chain.

beauby: What error(s) are you getting?

beauby: Oh yeah, this probably clashes with the workspace thing. There might be a way to make it work though, did you try removing the ˋnpm:` prefix?

skiplabsdaniel: The skip-wasm@^1.0.0 does not exists.

skiplabsdaniel: Removing `npm:` `skip-wasm@^1.0.0 ` considered as version number

skiplabsdaniel: pnpm seems to manage it with `workspace:` prefix, but not npm.

PR #445, opened 2024-10-31
Rename SkipService inputCollections to initialData

jberdine:
I am finding the name of the inputCollections field of SkipService
confusing, and so propose a renaming. There are two aspects: the type of the
field is not a Collection in the Skip sense, and the name inputCollections
is also used for that actual collections argument of reactiveCompute. The
reuse of the name is confusing, I think, as the field is the initial data
while the argument is the working data of the service. Users should not ever
read the inputCollections field, only provide it for runService (and
eventually initService) to read. But users should read from the
inputCollections argument. I think that renaming to emphasize that the field
is only the initial data will help.


bennostein:
Sound good to me.  Agree that this could be confusing.

PR #448, opened 2024-11-01
Rename TJSON to Json and JSONObject to JsonObject

jberdine:
RFC PR to understand what the problem is with renaming TJSON to JSON.


jberdine:
I seem to recall someone saying that there was a problem with this renaming in a past discussion, @skiplabsdaniel maybe? Things seem ok as far as what our CI can check. Any pointers?

skiplabsdaniel:
The problem with renaming TJSON to JSON is as soon you import JSON type you cannot anymore use JSON.parse or JSON.stringify functions.

jberdine:
Ok, thanks. What do people think about `Json` instead? We are even not using "JSON" with it's acronym meaning anyhow, so just capitalizing it might even be more correct. My experimentation hasn't run into any issues like `JSON.stringify` being unavailable in this case.

bennostein:
I am happy with any of these options, no strong feelings.

But, one point of clarification, maybe I'm missing something: what exactly breaks if we have a type shadowing a value?  seems like that shouldn't be a problem with type erasure.  e.g. https://typescript-eslint.io/rules/no-shadow/#ignoretypevalueshadow

jberdine:
> But, one point of clarification, maybe I'm missing something: what exactly breaks if we have a type shadowing a value? seems like that shouldn't be a problem with type erasure. e.g. https://typescript-eslint.io/rules/no-shadow/#ignoretypevalueshadow

Discussed offline, summary: there is [lib.es5](https://github.com/microsoft/TypeScript/blob/main/src/lib/es5.d.ts) that also defines a JSON type and that conflicts.

PR #449, opened 2024-11-01
Simplify nomenclature around external services

bennostein:
Writing docs about external services/resources/suppliers, I found the namings to be quite confusing and at times conflicting or redundant.

The previous situatiuon was as follows: `ExternalSupplier` is an umbrella interface for all kinds of external data sources, whether skip or not.  It had two main implementations: `ExternalService` for generic non-skip externals, and `ExternalSkipService` for skip externals.  `ExternalService`s were composed of multiple `ExternalResource`s and `ExternalSkipService`s also had an internal notion of "resource"; skip services had a field "externalServices" which held a collection of "ExternalSupplier"s, etc. etc.

This diff attempts to unify/streamline the naming scheme.  

The umbrella interface is now `ExternalService`, and it is implemented by `SkipExternalService` and `GenericSkipService`.  Both forms have a notion of resource still, but they are analogous and I don't think will cause too much trouble.  There may still be some places where parameter/variable names are conflicting as that's tricky to grep for, but I tried to find/fix the ones I could.

I'm continuing to work on further in-code and markdown docs around this stuff, but thought I'd put this up to get feedback and minimize rebase pain.

### DIFF
--- a/skipruntime-ts/api/src/api.ts
+++ b/skipruntime-ts/api/src/api.ts
@@ -320,9 +320,20 @@ export type ReactiveResponse = {
 };
 
 /**
- * ExternalSupplier allows to have access to external ressources
+ * External services must be carefully managed in reactive Skip services to make sure that
+ * dependencies are properly tracked and data from external systems is kept up to date.
+ *
+ * There are two main implementations of `ExternalService`, depending on whether the external
+ * data source is a reactive Skip service or not.
+ *
+ * If it _is_ a Skip service, then use `SkipExternalService`, specifying an entrypoint and/or
+ * auth handler.  The external Skip service will update its resources reactively and those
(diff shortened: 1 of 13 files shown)


jberdine (approved):
LGTM. I do wonder if there might be a better name for GenericExternalService. Are they all going to be polled? Maybe call them PolledExternalService?

bennostein:
> LGTM. I do wonder if there might be a better name for GenericExternalService. Are they all going to be polled? Maybe call them PolledExternalService?

They are not all polled!  E.g. `TimeCollection` in `external.ts` exposes essentially our old refresh token system as an external resource which can be used in a generic external service.  My grepping missed that misnomer, though -- I've just renamed it to `TimerResource`

PR #498, opened 2024-11-08
Renamings/refactorings from group discussion

bennostein:
Mostly just renamings through the first few commits, but the final commit's `getUnique` changes the semantics for that operation and modifies the tests accordingly.


bennostein:
Landing now so that others can rebase, but will keep an eye here if there are any comments following on

PR #499, opened 2024-11-12
Server-sent events

### LINKED ISSUE #472: Swap out websockets for Server-Sent Events

beauby:
Closes #472.


bennostein (approved):
+158/-1573, hell yeah.

The new versions of clients in the examples are much simpler/nicer, too!

I guess the only question is whether we should be worried about the connection limit on HTTP/1.   But that's a problem for the future -- if I understand correctly the limit is per client/domain, so most use cases we have at this point will be just fine.

jberdine:
Is this waiting on the example, or could it be merged?

beauby:
> Is this waiting on the example, or could it be merged?

There are 2 remaining things I'd like to get in:
1. leverage the SSE protocol's "event type" feature to distinguish between init/update, instead of sending it as a boolean field of the messages
2. decide on whether to expose a read-only streaming endpoint vs having the web app stream the data through.

I'll wrap up 1. now, and maybe we can slate 2. for later as it does not prevent playing with skip at the moment.

PR #500, opened 2024-11-12
Fixes #496: support for joinCollections

### LINKED ISSUE #496: Join helpers for eager collections
this can be done with merge now, but it would be nice to provide helpers to do left/outer/inner/right joins on different collections

pikatchu:
This diff implements the 'join' operation of two collections. It is assumed that:
- both collection values are objects (keys can be of any type)
- there never is more than 1 object on boths sides at any time unless the option allowMultipleValuesOnBothSides is passed (which should be rare).

If both those conditions are met, given a key, if the first collection defines an object of type V1 and the second collection defines an object V2 for the same key, the resulting collection will associate {...V1, ...V2} for that key. No value is associated if either side is missing.

### DIFF
--- a/skipruntime-ts/helpers/src/utils.ts
+++ b/skipruntime-ts/helpers/src/utils.ts
@@ -5,6 +5,8 @@ import type {
   NonEmptyIterator,
   ReactiveResponse,
   Json,
+  EagerCollection,
+  Mapper,
 } from "@skipruntime/api";
 
 export class Sum implements Reducer<number, number> {
@@ -67,4 +69,133 @@ export function reactiveResponseHeader(
   reactiveResponse: ReactiveResponse,
 ): [string, string] {
   return ["Skip-Reactive-Response-Token", JSON.stringify(reactiveResponse)];
+}
+
+/*****************************************************************************/
+// Unique collection
+/*****************************************************************************/
+
+export type UniqueEagerCollection<
+  K extends Json,
+  V extends Json,
+> = EagerCollection<K, V> & { unique: true };
+
+class UniqueMapper<K extends Json, V extends Json>
+  implements Mapper<K, V, K, V>
+{
+  mapEntry(key: K, values: NonEmptyIterator<V>): Iterable<[K, V]> {
+    return [[key, values.getUnique()]];
+  }
+}
+
+export function makeUniqueCollection<K extends Json, V extends Json>(
+  col: EagerCollection<K, V>,
+): UniqueEagerCollection<K, V> {
+  return { ...col.map(UniqueMapper<K, V>), unique: true };
+}
+
+/*****************************************************************************/
+// Joins
+/*****************************************************************************/
+
+type JoinObject = { value: Json; side: "left" | "right" };
+
+class AddJoinSideField implements Mapper<Json, Json, Json, JoinObject> {
+  constructor(private joinSide: "left" | "right") {}
+  mapEntry(
+    key: Json,
+    values: NonEmptyIterator<Json>,
+  ): Iterable<[Json, JoinObject]> {
+    let result: Array<[Json, JoinObject]> = [];
+    for (const v of values) {
+      if (typeof v !== "object") {
+        throw new Error(
+          "joinCollection only works on objects, not: " + JSON.stringify(v),
+        );
+      }
+      if (v === null) {
+        throw new Error("joinCollection does not accept null");
+      }
+      result.push([key, { value: v, side: this.joinSide }]);
+    }
+    return result;
+  }
+}
+
+function mergeObjects(object1: JoinObject, object2: JoinObject): Json {
+  const v1 = object1.value;
+  const v2 = object2.value;
+  if (typeof v1 !== "object" || v1 === null) {
+    throw new Error(
+      "mergeObjects only works with objects, not: " + JSON.stringify(v1),
+    );
+  }
+  if (typeof v2 !== "object" || v1 === null) {
+    throw new Error(
+      "mergeObjects only works with objects, not: " + JSON.stringify(v2),
+    );
+  }
+  const keys1 = Object.keys(v1);
+  if (keys1.some((key) => key in v2)) {
+    throw new Error(
+      "Objects don't have distinct fields: " +
+        JSON.stringify(v1) +
+        " " +
+        JSON.stringify(v2),
+    );
+  }
+  return { ...v1, ...v2 };
+}
+
+class MergeJoinFields implements Mapper<JoinObject, JoinObject, Json, Json> {
+  constructor(private allowNN: boolean) {}
+  mapEntry(
+    key: Json,
+    values: NonEmptyIterator<JoinObject>,
+  ): Iterable<[Json, Json]> {
+    const result: Array<[Json, Json]> = [];
+    let countLeft = 0;
+    let countRight = 0;
+    for (const value of values) {
+      if (value.side === "left") {
+        countLeft++;
+      }
+      if (value.side === "right") {
+        countRight++;
+      }
+    }
+    if (countLeft > 1 && countRight > 1) {
+      if (!this.allowNN) {
+        throw new Error(
+          "More than one value detected on both sides for key: " +
+            JSON.stringify(key),
+        );
+      }
+    }
+    for (const value1 of values) {
+      if (value1.side === "left") {
+        for (const value2 of values) {
+          if (value2.side === "right") {
+            result.push([key, mergeObjects(value1, value2)]);
+          }
+        }
+      }
+    }
+    return result;
+  }
+}
+
+export function joinCollections<
+  K extends Json,
+  V1 extends Json,
+  V2 extends Json,
+>(
+  col1: EagerCollection<K, V1>,
+  col2: EagerCollection<K, V2>,
+  allowNN: boolean = false,
+): EagerCollection<K, V1 & V2> {
+  return col1
+    .map(AddJoinSideField, "left")
+    .merge(col2.map(AddJoinSideField, "right"))
+    .map(MergeJoinFields, allowNN) as EagerCollection<K, V1 & V2>;
 }
--- a/skipruntime-ts/helpers/src/utils.ts
+++ b/skipruntime-ts/helpers/src/utils.ts
@@ -123,7 +123,7 @@ function mergeObjects(object1: JoinObject, object2: JoinObject): Json {
 }
 
 class MergeJoinFields implements Mapper<JoinObject, JoinObject, Json, Json> {
-  constructor(private allowNN: boolean) {}
+  constructor(private allowMultipleValuesOnBothSides: boolean) {}
 
   mapEntry(
     key: Json,
@@ -141,7 +141,7 @@ class MergeJoinFields implements Mapper<JoinObject, JoinObject, Json, Json> {
       }
     }
     if (countLeft > 1 && countRight > 1) {
-      if (!this.allowNN) {
+      if (!this.allowMultipleValuesOnBothSides) {
         throw new Error(
           "More than one value detected on both sides for key: " +
             JSON.stringify(key),
@@ -168,10 +168,13 @@ export function joinCollections<
 >(
   col1: EagerCollection<K, V1>,
   col2: EagerCollection<K, V2>,
-  allowNN: boolean = false,
+  allowMultipleValuesOnBothSides: boolean = false,
 ): EagerCollection<K, V1 & V2> {
   return col1
     .map(AddJoinSideField, "left")
     .merge(col2.map(AddJoinSideField, "right"))
-    .map(MergeJoinFields, allowNN) as EagerCollection<K, V1 & V2>;
+    .map(MergeJoinFields, allowMultipleValuesOnBothSides) as EagerCollection<
+    K,
+    V1 & V2
+  >;
 }
--- a/skipruntime-ts/helpers/src/utils.ts
+++ b/skipruntime-ts/helpers/src/utils.ts
@@ -152,8 +152,9 @@ class MergeJoinFields implements Mapper<JoinObject, JoinObject, Json, Json> {
     if (countLeft > 1 && countRight > 1) {
       if (!this.allowMultipleValuesOnBothSides) {
         throw new Error(
-          "More than one value detected on both sides for key: " +
-            JSON.stringify(key),
+          `Potentially expensive join of ${countLeft} x ${countRight} \
+           values for key: ${JSON.stringify(key)}. \
+           If this is intended, pass allowMultipleValuesOnBothSides = true.`,
         );
       }
     }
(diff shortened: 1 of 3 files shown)


skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +70,132 @@ export function reactiveResponseHeader(
...
+        JSON.stringify(v1) +
+        " " +
+        JSON.stringify(v2),
+    );
+  }
+  return { ...v1, ...v2 };
+}
+
+class MergeJoinFields implements Mapper<JoinObject, JoinObject, Json, Json> {
+  constructor(private allowNN: boolean) {}
```
bennostein: Hmm, I can't figure out what "NN" stands for here -- if it's meant to be rare, shall we just expand that abbreviation?

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +70,132 @@ export function reactiveResponseHeader(
...
+export type UniqueEagerCollection<
+  K extends Json,
+  V extends Json,
+> = EagerCollection<K, V> & { unique: true };
+
+class UniqueMapper<K extends Json, V extends Json>
+  implements Mapper<K, V, K, V>
+{
+  mapEntry(key: K, values: NonEmptyIterator<V>): Iterable<[K, V]> {
+    return [[key, values.getUnique()]];
```
bennostein: Is this ever _not_ either a no-op or an exception?

I'm not sure that the benefit of having a special `UniqueEagerCollection` type is worth the extra API surface and copying/allocation (though I suppose that will be minimal here due to sharing).

I'd propose either (a) checking uniqueness during the actual join operation and dropping this stuff or (b) making UniqueEagerCollection a more full-fledged component of the API, with its own getter/mapper types that take advantage of the uniqueness.  Probably slightly prefer (a) for simplicity's sake, but open to arguments / the possibility that I'm misunderstanding something here.

skipruntime-ts/helpers/src/utils.ts
pikatchu: It's for N to N relations. But I renamed it to allowMultipleValuesOnBothSides.

skipruntime-ts/helpers/src/utils.ts
pikatchu: Ok. I removed it. The join doesn't need it to operate, so no change needed there.

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +70,111 @@ export function reactiveResponseHeader(
 ): [string, string] {
   return ["Skip-Reactive-Response-Token", JSON.stringify(reactiveResponse)];
 }
+
+/*****************************************************************************/
+// Joins
+/*****************************************************************************/
+
+type JoinObject = { value: Json; side: "left" | "right" };
```
jberdine, line 81: Would `JsonObject` instead of `Json` here be more accurate and simplify the code and interface?

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +70,111 @@ export function reactiveResponseHeader(
...
+      if (typeof v !== "object") {
+        throw new Error(
+          "joinCollection only works on objects, not: " + JSON.stringify(v),
+        );
+      }
+      result.push([key, { value: v, side: this.joinSide }]);
+    }
+    return result;
+  }
+}
```
jberdine: Am I missing something or could this be simplified by extending OneToOneMapper instead of implementing Mapper?

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +70,111 @@ export function reactiveResponseHeader(
...
+      "mergeObjects only works with objects, not: " + JSON.stringify(v1),
+    );
+  }
+  if (typeof v2 !== "object") {
+    throw new Error(
+      "mergeObjects only works with objects, not: " + JSON.stringify(v2),
+    );
+  }
+  const keys1 = Object.keys(v1);
+  if (keys1.some((key) => key in v2)) {
```
jberdine: Are these objects expected to be small, or might they be dictionaries or arrays? Asking since this is quadratic and doesn't strictly need to be.

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +70,111 @@ export function reactiveResponseHeader(
...
+}
+
+export function joinCollections<
+  K extends Json,
+  V1 extends Json,
+  V2 extends Json,
+>(
+  col1: EagerCollection<K, V1>,
+  col2: EagerCollection<K, V2>,
+  allowMultipleValuesOnBothSides: boolean = false,
```
jberdine, line 182: The expectation is that the fields of the values in the left and right value types are disjoint, right? Or is it expected that idiomatic use cases will have non-disjoint types, but only per-key will disjointness will hold?

My understanding is that the simplest case is where both the left and right argument collections map keys to unique values. On the other hand, if the left and right argument collections map keys to potentially multiple values, then the result will map each key to the merge of each pair in the cartesian product of the left values and right values for that key.

Maybe this is just me, but I don't see why the intermediate cases, where only one of the argument collections is multi-valued, are grouped with the simple case rather than the general one. Would it be a simpler interface if there were two booleans, one stating that the left collection is allowed to be multi-valued and the other stating that the right collection is allowed to be multi-valued?

skipruntime-ts/tests/src/tests.ts
```diff
@@ -378,6 +382,29 @@ class MergeReduceService implements SkipService {
...
 }
 
+// Test join
+
+class JoinResource implements Resource {
+  instantiate(cs: {
+    input1: UniqueEagerCollection<number, JsonObject>;
+    input2: EagerCollection<number, JsonObject>;
+  }): EagerCollection<number, JsonObject> {
+    return joinCollections(cs.input1, cs.input2);
```
jberdine, line 388: Add a TODO to add tests for the 1_to_N, N_to_1, and N_to_N cases?

skipruntime-ts/tests/src/tests.ts
```diff
@@ -16,7 +16,11 @@ import type {
   ReactiveResponse,
 } from "@skipruntime/api";
 import { NonUniqueValueException, OneToOneMapper } from "@skipruntime/api";
-import { Sum } from "@skipruntime/helpers";
+import {
+  Sum,
+  joinCollections,
+  type UniqueEagerCollection,
```
jberdine: gone now

skipruntime-ts/helpers/src/utils.ts
pikatchu: I tried. But I didn't manage to make it work. Typescript is complaining with an error that I don't understand ...

skipruntime-ts/helpers/src/utils.ts
pikatchu: Ah. That's a very good point. Let me revisit.

skipruntime-ts/helpers/src/utils.ts
pikatchu: Good point.

skipruntime-ts/helpers/src/utils.ts
pikatchu: Let me try (I thought I tried to add JsonObject and it didn't work, but it might have been somewhere else).

pikatchu: So yeah, typescript refuses it. I don't understand why ...

skipruntime-ts/helpers/src/utils.ts
pikatchu: But the objects are indeed expected to be small ... still, no need to keep something O(n^2) for no reason.

skipruntime-ts/helpers/src/utils.ts
pikatchu: The reason is my experience with the SQL engine (which might be flawed). I have seen multiple cases where either side had more than one value, but never something that required a cartesian product that was actually useful. Let's review them, let's assume we have 2 large collections (large in the sense that N*M is too large to be computed).

Example, one / many: A table of users, a table of addresses, a user can have multiple addresses => maximum size of the able = number of address
Example, one / one: A table of users, a table of social security numbers => maximum size of the table = min(users, ssn)
Example, many / many: ??? Usually a contrived example that is only useful to explain the behavior of a join 

I understand the cartesian product in the many/many case is the natural extension of what the join should do. But I have yet to see anything in practice using it. In fact, more often than not, it's a mistake, the query explodes and the user does not know why.

When you think about it, it makes sense. A cartesian product is only going to work for very small sets ...

So I would rather make it explicit. It's not banned. You just have to be explicit about it.

skipruntime-ts/helpers/src/utils.ts
jberdine: I tried and pushed the result to your branch. Seems to work, or am I missing something?

skipruntime-ts/helpers/src/utils.ts
jberdine: Annoyingly, Object.keys is not sorted, so they would have to be first sorted and then zipped together a la mergesort.

skipruntime-ts/helpers/src/utils.ts
pikatchu: So I see you pushed a diff that makes JsonObject bubble up all the way to joinCollections. I don't know that this is a good idea.

The reason is that when I tried to use joinCollections on inputs, Typescript did not complain even when the input collections where not objects. That is a pretty subtle "gotcha" that I would like to avoid.

jberdine:
Summarizing offline discussion: It would be possible to give `joinCollections` the type
```
export function joinCollections<
  K extends Json,
  V1 extends JsonObject,
  V2 extends JsonObject,
>(
  col1: EagerCollection<K, V1>,
  col2: EagerCollection<K, V2>,
  allowMultipleValuesOnBothSides: boolean = false,
): EagerCollection<K, V1 & V2>
```
and propagate `JsonObject` down, eliminating the checks that each value is an object.

BUT, since typescript unsoundly treats method parameters as bivariant, calling `joinCollections` e.g. from `createGraph` where `inputCollections: Record<string, EagerCollection<Json, Json>>` will not detect mistakes where general `Json` values are provided where `JsonObject`s are required. So this PR's dynamic checks on each value are still needed. Thanks for nothing typescript.

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +71,96 @@ export function reactiveResponseHeader(
...
+  const v1 = object1.value;
+  const v2 = object2.value;
+  const keys1 = Object.keys(v1);
+  if (keys1.some((key) => key in v2)) {
+    throw new Error(
+      "Objects don't have distinct fields: " +
+        JSON.stringify(v1) +
+        " " +
+        JSON.stringify(v2),
+    );
```
jberdine, line 130: A question on the error behavior: join in relational algebra drops pairs that do not agree on common fields, while here it is an error for there to be common fields (other than the key). No strong position, but it isn't immediately obvious to me that this choice is the one users will expect, or be the most useful.

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +71,96 @@ export function reactiveResponseHeader(
...
+            result.push([key, mergeObjects(value1, value2)]);
+          }
+        }
+      }
+    }
+    return result;
+  }
+}
+
+export function joinCollections<
```
jberdine, line 175: From an implementation point of view, I understand why this function is here. But from the API point of view, I think it would be better if this was a `join` operation in the `EagerCollection` interface. Also, if/when we reimplement this in skip, putting it in helpers will not be the right place anymore.

skipruntime-ts/helpers/src/utils.ts
jberdine: To clarify, I wonder about just dropping the pairs with common fields other than the key, rather than throwing, not taking pairs that happen to have matching values for common fields.

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +72,118 @@ export function reactiveResponseHeader(
...
+    for (const value of values) {
+      if (value.side === "left") {
+        countLeft++;
+      }
+      if (value.side === "right") {
+        countRight++;
+      }
+    }
+    if (countLeft > 1 && countRight > 1) {
+      if (!this.allowMultipleValuesOnBothSides) {
```
jberdine, line 154: This could move up to before `let countLeft...` to avoid the counting loop in the allow case.

skipruntime-ts/helpers/src/utils.ts
pikatchu: I don't know people know relational algebra well enough to expect that. I suspect they don't.
I am happy to add an option (dropOnCommonFields for example), but making that the default is not a good idea IMO. Imagine that is not the behavior you expect, now imagine you end up inserting new kinds of objects in a collection with a field that collides with another collection by mistake. The debugging is going to be pretty brutal ...

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +72,118 @@ export function reactiveResponseHeader(
...
+        throw new Error(
+          "More than one value detected on both sides for key: " +
+            JSON.stringify(key),
```
jberdine:
```suggestion
          `Potentially expensive join of ${countLeft} x ${countRight} \
           values for key: ${JSON.stringify(key)}. \
           If this is intended, pass allowMultipleValuesOnBothSides = true.`,
```

skipruntime-ts/helpers/src/utils.ts
```diff
@@ -68,3 +72,118 @@ export function reactiveResponseHeader(
...
+      }
+    }
+    return result;
+  }
+}
+
+export function joinCollections<
+  K extends Json,
+  V1 extends Json,
+  V2 extends Json,
```
jberdine: Even if TS doesn't check it, I think it would be a documentation improvement to use JsonObject for the value types here.

skipruntime-ts/helpers/src/utils.ts
pikatchu: It's not the common case, so I think it's fine. If allowMultipleValuesOnBothSides is true, your set better be very small to begin with ...

skipruntime-ts/helpers/src/utils.ts
pikatchu: You are right. I wish there was a clean way of extending a class in another file. You know something like:
```
extension class EagerCollection {
  join(...): ...
}
```

skipruntime-ts/helpers/src/utils.ts
pikatchu: Agreed. Done.

skipruntime-ts/helpers/src/utils.ts
jberdine: Ok makes sense, the debugging angle is compelling. And we can wait to see if anyone asks to drop instead of error.

PR #501, opened 2024-11-13
Convert SkipService-implementing classes to object literals in examples

bennostein:
This is a change we discussed at the offsite and mostly agreed was an improvement, since it makes more clear that the object is an "initial config" / "definition of the service" and not an OO representation that could keep being updated/interacted with over the life of the program.

What do you guys think, seeing it in better context here?  I also inlined the object literal to some `runService` calls where it made sense -- I expect that will be a common way of using these interfaces, and it looks pretty good/ergonomic to me FWIW.

### DIFF
--- a/examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
+++ b/examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
@@ -26,38 +26,28 @@ type Upvote = {
 
 type Upvoted = Post & { upvotes: number; author: User };
 
-export default class HackerNewsService implements SkipService {
-  initialData: {
-    posts: Entry<string, Post>[];
-    users: Entry<string, User>[];
-    upvotes: Entry<string, Upvote>[];
+export function serviceWithInitialData(
+  posts: Entry<string, Post>[],
+  users: Entry<string, User>[],
+  upvotes: Entry<string, Upvote>[],
+): SkipService {
+  return {
+    initialData: { posts, users, upvotes },
+    resources: { posts: PostsResource },
+    createGraph: (inputCollections: {
+      posts: EagerCollection<number, Post>;
+      users: EagerCollection<number, User>;
+      upvotes: EagerCollection<number, Upvote>;
+    }) => {
+      return {
+        postsWithUpvotes: inputCollections.posts.map(
+          PostsMapper,
+          inputCollections.users,
+          inputCollections.upvotes.map(UpvotesMapper),
+        ),
+      };
+    },
   };
-  resources = { posts: PostsResource };
-
-  constructor(
-    posts: Entry<string, Post>[],
-    users: Entry<string, User>[],
-    upvotes: Entry<string, Upvote>[],
-  ) {
-    this.initialData = { posts, users, upvotes };
-  }
-
-  createGraph(inputCollections: {
-    posts: EagerCollection<number, Post>;
-    users: EagerCollection<number, User>;
-    upvotes: EagerCollection<number, Upvote>;
-  }): Record<string, EagerCollection<Json, Json>> {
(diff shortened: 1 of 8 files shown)


examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
```diff
@@ -26,38 +25,28 @@ type Upvote = {
...
+  upvotes: Entry<string, Upvote>[],
+): SkipService {
+  return {
+    initialData: { posts, users, upvotes },
+    resources: { posts: PostsResource },
+    createGraph: (inputCollections: {
+      posts: EagerCollection<number, Post>;
+      users: EagerCollection<number, User>;
+      upvotes: EagerCollection<number, Upvote>;
+    }) => {
```
jberdine: Can the return type be specified here? It could be tightened to `Record<string, EagerCollection<number, Upvoted>>`, but just because there is only one collection here, so doing that in the examples is probably misleading. So actually nevermind.

jberdine: One thing that might be good in examples though is to define a type alias for e.g. the type of `PostsResource.instantiate.collection` and use it both there and as the return type of `createGraph` to emphasize the value flow.

jberdine (approved):
Overall I think this is an improvement, making the example user code feel lighter-weight.

examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
bennostein: Something like https://github.com/SkipLabs/skip/pull/501/commits/b8ddc8bf950d546afff3f33fc63a0a44193b0a84 ?

That gives me the idea that we can type-parameterize services more generally, using these `Record<string, EagerCollection<...>>` things as bounds instead of directly in types... will give that a shot, see how it looks.

examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
```diff
@@ -26,38 +25,33 @@ type Upvote = {
...
-      upvotes,
-    );
+type Inputs = {
+  posts: EagerCollection<number, Post>;
+  users: EagerCollection<number, User>;
+  upvotes: EagerCollection<number, Upvote>;
+};
+type Outputs = {
+  postsWithUpvotes: EagerCollection<number, Upvoted>;
+};
```
jberdine: This is nice, but the name `Outputs` emphasizes the role as the result of `SkipService.createGraph` at the expense of the role as the argument of `Resource.instantiate`. It would be better to use a name that is good for both I think, and if anything, the resource role might be the more important one.

Maybe rename `Inputs` to `InitialInputs` and `Outputs` to `ResourceInputs`.

examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
```diff
@@ -26,38 +25,33 @@ type Upvote = {
...
-  }
+export function serviceWithInitialData(
+  posts: Entry<string, Post>[],
+  users: Entry<string, User>[],
+  upvotes: Entry<string, Upvote>[],
+): SkipService {
+  return {
+    initialData: { posts, users, upvotes },
+    resources: { posts: PostsResource },
+    createGraph: (inputCollections: Inputs) => {
```
jberdine: I think the return type annotation is important to make the connection with the argument of `instantiate`
```suggestion
    createGraph: (inputCollections: Inputs): Outputs => {
```

bennostein:
Accidentally merged this before squashing the fixups, but just went in and cleaned it up on `main`, fyi.  Would be nice if github would check/automatically do that squash when you merge through the UI

PR #503, opened 2024-11-14
Strengthen typing of SkipService.initialData vs createGraph

jberdine:
This diff changes the type parameters of SkipService so that they only
specify corresponding collection names, key types and value types, without
already instantiating the EagerCollection type. This type "signature" can
then be used to construct the instantiations of Entry for the type of
initialData as well as the instantiations of EagerCollection for the type of
e.g. createGraph.

This uses typescript's "mapped" and "indexed access" types but does not need
the machinery of "conditional" or "infer" types.

### DIFF
--- a/skipruntime-ts/tests/src/tests.ts
+++ b/skipruntime-ts/tests/src/tests.ts
@@ -73,7 +73,7 @@ const map1Service: SkipService = {
   initialData: { input: [] },
   resources: { map1: Map1Resource },
 
-  createGraph(inputCollections: { input: EagerCollection<number, number> }) {
+  createGraph(inputCollections: { input: EagerCollection<string, number> }) {
     return inputCollections;
   },
 };
--- a/examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
+++ b/examples/hackernews/back-end/reactive_service/src/hackernews.service.ts
@@ -2,6 +2,7 @@ import type {
   Entry,
   EagerCollection,
   NonEmptyIterator,
+  CollectionsOf,
   SkipService,
   Resource,
 } from "skip-wasm";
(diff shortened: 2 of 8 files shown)


skipruntime-ts/tests/src/tests.ts
```diff
@@ -73,7 +73,7 @@ const map1Service: SkipService = {
   initialData: { input: [] },
   resources: { map1: Map1Resource },
 
-  createGraph(inputCollections: { input: EagerCollection<number, number> }) {
+  createGraph(inputCollections: { input: EagerCollection<string, number> }) {
```
jberdine, line 76: another type error that had slipped through

jberdine:
based on https://github.com/SkipLabs/skip/pull/502

bennostein (commented):
I like this change -- less conditional/infer stuff is good, and the boilerplate changes seem to me to be basically a wash

bennostein (commented):
Would you elaborate on what you mean by "more invasive" (in your comment on the other PR) though? I assume you're referring to the increase in type definitions in the examples but maybe there's something I'm missing.

jberdine:
Basically what you mentioned about more type definitions in the examples, but also the type functions added to the api take up some namespace. At the core the approach here that builds the two instantiations, Entrys and EagerCollections, from something, rather than destructing one and building the other from it, leads to needing 2 types rather than 1 in the code that uses SkipService. Also, in the destruct-and-rebuild approach, the primary thing that user code defines and talks about is the actual eager collection types, and it seems right for that to be the main focus for the user code, since those are the core around which the service is built. In this PR, the focus is more on the name-key-value association, which isn't too bad, but is something auxiliary. From that perspective I am not completely sure it is better to go with this approach versus pushing more in the conditional types with infer direction. I expect that often it would work to not name the argument to CollectionsOf and to not explicitly instantiate SkipService, allowing only 1 type in user code, but I don't trust that doing so wouldn't essentially just turn off type-checking.

Thoughts @skiplabsdaniel? And ping @pikatchu who has had opinions on conditional types and infer previously.

bennostein:
Yeah, I agree that the user code before this PR is a bit more natural in the sense that the type your write down is the type of values your code operates over.  Maybe the thing to do is shelf this as a possible future change if/when `infer` and conditional types cause problems?

jberdine:
Yeah, ok, I'll rebase and close for now.

PR #505, opened 2024-11-15
Prefer indexed object types over Record

jberdine:
The type `Record<K,V>` can be helpful when `K` is a finite union, especially
when named and used elsewhere, to deduplicate definitions, improve
exhaustivity checking, etc. But for cases such as `Record<string,V>`, it is
just an indirect way (using a mapped type) to express an indexed object
type. This PR switches to indexed types, which seem to be more idiomatic,
and subjectively clearer.


bennostein (commented):
This looks like an improvement to my eyes!  Especially for things like ports, resources, etc. where it's not immediately obvious what the record is without naming the keys.

PR #506, opened 2024-11-15
Improve freeze

jberdine:
- Rename freeze to deepFreeze

  To emphasize that it modifies the entire object graph reachable from a
  value, unlike `Object.freeze`, and it is not expected to be used very
  frequently.

- Remove redundancy from return type.

- Remove assignments that are superfluous since deepFreeze returns its
  argument.

- Move `check` near to other operations related to frozen

- Improve `check` error messages
 
- Do not support manually deep-frozen values

  Supporting values that are manually deep-frozen rather than using
  `deepFreeze` is a complication, though slight, and a hidden latent
  performance foot-gun. Without use cases that need it, this PR simplifies the
  API and implementation by removing this support, requiring users to call
  `deepFreeze` instead.

- Treat bigint and symbol types as primitives in deepFreeze

  Also clarify that undefined and function values are not supported.

### DIFF
--- a/skipruntime-ts/wasm/src/internals/skipruntime_module.ts
+++ b/skipruntime-ts/wasm/src/internals/skipruntime_module.ts
@@ -77,13 +77,13 @@ abstract class SkFrozen extends Frozen {
  *
  * The primary use for this function is to satisfy the requirement that all
  * parameters to Skip `Mapper` constructors must be deep-frozen: objects
- * that have not been constructed by Skip can be passed to `freeze()` before
- * passing them to a `Mapper` constructor.
+ * that have not been constructed by Skip can be passed to `deepFreeze()`
+ * before passing them to a `Mapper` constructor.
  *
  * @param value - The object to deep-freeze.
  * @returns The same object that was passed in.
  */
-export function freeze<T>(value: T): (T & Param) | (T & Constant) {
+export function deepFreeze<T>(value: T): (T & Param) | (T & Constant) {
   if (
     typeof value == "string" ||
     typeof value == "number" ||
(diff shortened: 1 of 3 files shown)


skipruntime-ts/wasm/src/internals/skipruntime_module.ts
```diff
@@ -63,6 +63,31 @@ abstract class SkFrozen extends Frozen {
...
+      if (Array.isArray(value)) {
+        value.forEach(check);
+      } else {
+        Object.values(value).forEach(check);
+      }
+    } else {
+      throw new Error("Invalid object: must be deep-frozen.");
+    }
+  } else {
+    throw new Error(`'${typeof value}' cannot be deep-frozen.`);
```
beauby, line 81: Not specific to this PR but: this else branch corresponds to `bigint`, `undefined`, `null`, and `symbol`, or is there an other type? Maybe here (or in `deepFreeze()` directly) is a good place to add a comment explaining why those can't be frozen.

skipruntime-ts/wasm/src/internals/skipruntime_module.ts
```diff
@@ -94,9 +94,6 @@ export function deepFreeze<T>(value: T): T & Param {
       return value;
     } else if (isSkFrozen(value)) {
       return value;
-    } else if (Object.isFrozen(value)) {
-      check(value);
-      return value as T & Param;
```
mbouaziz: Won't `sk_freeze` fail on an `Object.`-frozen object?

jberdine: You're right. I updated to instead give a better error message instead.

skipruntime-ts/wasm/src/internals/skipruntime_module.ts
jberdine: I think, but am not sure, that we can treat `bigint`~, `undefined`,~ and `symbol` as primitives that are already deep-frozen. Typeof never returns `null`, since `typeof null === "object"`. The only case left is `function`, which we can't deep-freeze. I added a comment.

mbouaziz (approved):

PR #510, opened 2024-11-19
Make `slice` variadic instead of taking an array argument

bennostein:
Pretty small change, but I noticed while writing out some examples that this could be made a bit lighter-weight syntactically, particularly noticeable when you're slicing out one range, i.e. `foo.slice([x, y])` instead of `foo.slice([[x, y]])`


beauby (approved):
LGTM

PR #511, opened 2024-11-19
Manage resource instance identifier to manage URL REWRITE With SSE

skiplabsdaniel:


skipruntime-ts/server/src/rest.ts
```diff
@@ -67,7 +64,7 @@ export function createRESTServer(service: ServiceInstance): express.Express {
             req.params.resource,
             req.query as Record<string, string>,
             {
-              resolve: (data: Values<Json, Json>) => {
+              resolve: (data: Entry<Json, Json>[]) => {
                 res.status(200).json(data.values);
```
beauby:
```suggestion
                res.status(200).json(data);
```

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -9,6 +9,12 @@ class ReadInCreatorException() extends Exception {
   }
 }
 
+class ExistingResourceException() extends Exception {
+  fun getMessage(): String {
+    "A resource instance with specified identifier already exist."
```
bennostein, line 14:
```suggestion
    "A resource instance with specified identifier already exists."
```

skipruntime-ts/tests/src/tests.ts
```diff
@@ -9,24 +9,22 @@ import type {
...
+/*
 import {
   TimerResource,
   GenericExternalService,
 } from "@skipruntime/helpers/external.js";
-
-type Values<K extends Json, V extends Json> = {
-  values: Entry<K, V>[];
-  reactive?: ReactiveResponse;
-};
```
bennostein, line 29: Can just delete instead of commenting out

PR #513, opened 2024-11-21
Update api docs workflow following packaging changes

jberdine:
and update docusaurus


jberdine:
This results in the separate packages having separate API documentation trees, and the sidebar now looks like:
image
I don't claim this is ideal. There might be some alternative incantations to get the API docs for multiple packages.

beauby:
I think it makes things clear for the user. On the other hand, we should probably rethink a bit the split into packages (bring back `@skipruntime/core` containing both native and wasm code, merge some stuff from `@skipruntime/helpers` into `@skipruntime/core`, etc.) but that is irrelevant to this PR.

jberdine:
Also, I didn't add an entry for skipruntime/wasm since when I checked it had not been published.

PR #515, opened 2024-11-21
[skip/server] Split control service and streaming service.

beauby:
This commit changes the REST API of the skip server, making it listen
on two separate ports, one for the public facing streaming API, and
one for the backend-only control API (sync reads/writes and stream
creation/deletion).

The control API (by default exposed on port `8081`) is now:
- `GET /v1/resources/:resource`: Sync read of a whole resource.
- `GET /v1/resources/:resource/:key`: Sync read of a single key from a
resource.
- `PATCH /v1/inputs/:collection`: Partial write (update only the
specified keys) of an input collection.
- `PUT /v1/inputs/:collection/:key`: Update of a single key of an
input collection.
- `POST /v1/streams`: Instantiate a resource using the provided
`resource` and `params` parameters (as a JSON object in the body),
and return an UUID to subscribe to
updates.
- `DELETE /v1/streams/:uuid`: Destroy a resource instance.

The streaming API (by default exposed on port `8080`) is now:
- `GET /v1/streams/:uuid`: SSE endpoint to subscribe to updates of the
resource instance represented by the UUID.

Splitting the two has several benefits:
- Avoiding the foot-gun of mistakenly exposing control routes to the
outside world.
- Making it easier to support a redirection-based mechanism where
clients connect directly to the streaming endpoint (vs deploying a
more complicated infra to efficiently stream the updates directly
through the web server).
- Not per se about splitting the APIs, but having an explicit stream
instantiation endpoint allows for explicit stream destruction (for
terminating currently subscribed streams of malicious actors
identified by an external mechanism for instance).

### DIFF
--- a/skipruntime-ts/examples/database-client.ts
+++ b/skipruntime-ts/examples/database-client.ts
@@ -10,15 +10,12 @@ import { fetchJSON } from "@skipruntime/helpers/rest.js";
 async function sleep(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
-const replication = 8081;
 const port = 8082;
 const url = `http://localhost:${port.toString()}`;
 
 console.log("Connect to replication server for resource /users");
 
-const evSource = new EventSource(
-  `http://localhost:${replication.toString()}/v1/users`,
-);
+const evSource = new EventSource(`${url}/users`);
 evSource.addEventListener("init", (e: MessageEvent<string>) => {
   const updates = JSON.parse(e.data);
   console.log("Init", updates);
--- a/skipruntime-ts/examples/database-server.ts
+++ b/skipruntime-ts/examples/database-server.ts
@@ -8,7 +8,8 @@ import express from "express";
 
 const service = new RESTWrapperOfSkipService({
   host: "localhost",
-  port: 8081,
+  control_port: 8081,
+  streaming_port: 8080,
 });
 
 import sqlite3 from "sqlite3";
@@ -35,6 +36,27 @@ const run = function (
     });
   });
 };
+
+app.get("/users", (_req, res) => {
+  fetch("http://localhost:8081/v1/streams", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({
+      resource: "users",
+      params: {},
+    }),
+  })
+    .then((rres) => {
+      return rres.text();
+    })
+    .then((uuid) => {
+      res.redirect(301, "http://localhost:8080/v1/streams/" + uuid);
+    })
+    .catch((e: unknown) => {
+      console.log(e);
+      res.status(500).json("Internal error");
+    });
+});
 
 app.get("/user/:id", (_req, res) => {
   service
--- a/skipruntime-ts/helpers/src/rest.ts
+++ b/skipruntime-ts/helpers/src/rest.ts
@@ -2,14 +2,15 @@ import type { Json, Entry } from "@skipruntime/api";
 
 export type Entrypoint = {
   host: string;
-  port: number;
+  streaming_port: number;
+  control_port: number;
   secured?: boolean;
 };
 
 function toHttp(entrypoint: Entrypoint) {
   if (entrypoint.secured)
-    return `https://${entrypoint.host}:${entrypoint.port}`;
-  return `http://${entrypoint.host}:${entrypoint.port}`;
+    return `https://${entrypoint.host}:${entrypoint.control_port}`;
+  return `http://${entrypoint.host}:${entrypoint.control_port}`;
 }
 
 export async function fetchJSON<V>(
@@ -46,7 +47,8 @@ export class RESTWrapperOfSkipService {
   constructor(
     entrypoint: Entrypoint = {
       host: "localhost",
-      port: 3587,
+      streaming_port: 8080,
+      control_port: 8081,
     },
   ) {
     this.entrypoint = toHttp(entrypoint);
@@ -58,7 +60,7 @@ export class RESTWrapperOfSkipService {
   ): Promise<Entry<K, V>[]> {
     const qParams = new URLSearchParams(params).toString();
     const [optValues, _headers] = await fetchJSON<Entry<K, V>[]>(
-      `${this.entrypoint}/v1/${resource}?${qParams}`,
+      `${this.entrypoint}/v1/resources/${resource}?${qParams}`,
       "GET",
     );
     const values = optValues ?? [];
@@ -72,7 +74,7 @@ export class RESTWrapperOfSkipService {
   ): Promise<V[]> {
     const qParams = new URLSearchParams(params).toString();
     const [data, _headers] = await fetchJSON<V[]>(
-      `${this.entrypoint}/v1/${resource}/${key}?${qParams}`,
+      `${this.entrypoint}/v1/resources/${resource}/${key}?${qParams}`,
       "GET",
     );
     return data ?? [];
@@ -90,7 +92,12 @@ export class RESTWrapperOfSkipService {
     collection: string,
     values: Entry<K, V>[],
   ): Promise<void> {
-    await fetchJSON(`${this.entrypoint}/v1/${collection}`, "PATCH", {}, values);
+    await fetchJSON(
+      `${this.entrypoint}/v1/inputs/${collection}`,
+      "PATCH",
+      {},
+      values,
+    );
   }
 
   async deleteKey<K extends Json>(collection: string, key: K): Promise<void> {
--- a/skipruntime-ts/native/src/Runtime.sk
+++ b/skipruntime-ts/native/src/Runtime.sk
@@ -9,6 +9,12 @@ class ReadInCreatorException() extends Exception {
   }
 }
 
+class ExistingResourceException() extends Exception {
+  fun getMessage(): String {
+    "A resource instance with specified identifier already exists."
+  }
+}
+
 class Params(value: Map<String, PValue>) extends SKStore.File uses Orderable {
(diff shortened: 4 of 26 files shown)


skipruntime-ts/server/src/rest.ts
```diff
@@ -1,18 +1,49 @@
...
-export function createRESTServer(service: ServiceInstance): express.Express {
+export function controlService(service: ServiceInstance): express.Express {
   const app = express();
   app.use(express.json());
+
+  // DEBUG
+  app.use((req, _res, next) => {
+    console.log("Control:", req.url, req.params);
+    next();
+  });
```
jberdine: This should be removed before landing right?

skipruntime-ts/server/src/server.ts
```diff
@@ -1,28 +1,44 @@
...
+  const streamingHttpServer = streamingService(runtime).listen(
+    options.port,
+    () => {
+      console.log(
+        `Skip streaming service listening on port ${options.port.toString()}`,
+      );
+    },
+  );
 
   // TODO: Return a proper object.
```
jberdine, line 36: Resolving this TODO will change the public API IIUC. I'm not sure what the comment suggests, is it hard?

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -1126,34 +1094,24 @@ class CollectionWriter(dirName: SKStore.DirName) {
...
-  ): Bool {
+  ): ?String {
     dir.creator match {
     | Some(arrow) ->
-      if (this.isResourceDir(arrow.parentName)) {
-        true
-      } else {
+      this.isResourceDir(arrow.parentName) match {
+      | true -> Some(base64(SKStore.SID::keyType(arrow.key).value))
+      | _ ->
```
jberdine: You don't like `if`?

skipruntime-ts/server/src/server.ts
```diff
@@ -1,28 +1,44 @@
...
   Inputs extends NamedCollections,
   Outputs extends NamedCollections,
 >(
   service: SkipService<Inputs, Outputs>,
-  port: number = 443,
+  options: {
+    port: number;
+    control_port: number;
+  } = {
+    port: 80,
```
jberdine: Maybe rename port to streaming_port for consistency with control_port.

skipruntime-ts/server/src/rest.ts
```diff
@@ -1,13 +1,41 @@
...
+    try {
+      service.closeResourceInstance(req.params.uuid);
+      res.sendStatus(200);
+    } catch (e: unknown) {
+      res.status(500).json(e instanceof Error ? e.message : e);
+    }
+  });
+
+  app.get("/v1/streams/:uuid", (_req, res) => {
+    res.status(200).json("TODO");
```
jberdine: ?

skipruntime-ts/server/src/rest.ts
```diff
@@ -97,5 +87,62 @@ export function createRESTServer(service: ServiceInstance): express.Express {
...
+  app.use(express.json());
+
+  app.get("/v1/streams/:uuid", (req, res) => {
+    if (!req.accepts("text/event-stream")) {
+      res.sendStatus(406);
+      return;
+    }
+    res.writeHead(200, {
+      "Content-Type": "text/event-stream",
+      Connection: "keep-alive",
```
jberdine, line 126: no quotes, intended, why?

jberdine (commented):
I don't understand the description regarding "public facing streaming API" versus "backend-only control API". User clients will need to use both in order to set up, pull from, and close a stream, right? It is "public-facing" versus "backend-only" that is confusing me. Any hints?

skipruntime-ts/server/src/server.ts
beauby: This is an old TODO meaning we could declare a proper `type` alias for `Promise<{ close: () => void }>`.

skipruntime-ts/server/src/rest.ts
beauby: Yes correct. Proper logging of requests should be implemented in a subsequent PR.

skipruntime-ts/native/src/Runtime.sk
beauby: That would be a question for @skiplabsdaniel :D

skipruntime-ts/server/src/rest.ts
beauby: Yes, IIRC the linter complains about unnecessary quotes.

skipruntime-ts/server/src/rest.ts
beauby: This is supposed to be a "maintenance"/"debug" endpoint, responding with data about the streaming session itself. As it is not necessary for now, I'll remove the endpoint from this PR altogether.

beauby:
> I don't understand the description regarding "public facing streaming API" versus "backend-only control API". User clients will need to use both in order to set up, pull from, and close a stream, right?

No, only the streaming endpoint will be exposed to the outside world. The resource instantiation (`POST /v1/streams`) itself will be handled by the web app (by making a call to the control API on the backend's local network), and responding to the client with a redirect to the streaming endpoint.

Basically you can see the control API as analogous to a Postgres database connection, in that it is never exposed to the outside world directly.

> It is "public-facing" versus "backend-only" that is confusing me. Any hints?

Does the above help?

skipruntime-ts/examples/database-client.ts
```diff
@@ -10,15 +10,12 @@ import { fetchJSON } from "@skipruntime/helpers/rest.js";
...
-const replication = 8081;
 const port = 8082;
 const url = `http://localhost:${port.toString()}`;
 
 console.log("Connect to replication server for resource /users");
 
-const evSource = new EventSource(
-  `http://localhost:${replication.toString()}/v1/users`,
-);
+const evSource = new EventSource(`${url}/users`);
```
jberdine, line 18: Do I understand correctly that we are assuming that buried in the implementation of `EventSource` is effectively an HTTP client, that in particular knows about the 301 redirect responses and silently makes another request? I haven't been able to find documentation for this. The [mdn http docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#principle) just state "When browsers receive a redirect, they immediately load the new URL" but the [docs of `EventSource`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/EventSource) don't say anything about how the url is treated, just that it is a string. Is there something I'm missing?

skipruntime-ts/examples/database-server.ts
```diff
@@ -36,6 +37,27 @@ const run = function (
...
+      return rres.text();
+    })
+    .then((uuid) => {
+      res.redirect(301, "http://localhost:8081/v1/streams/" + uuid);
+    })
+    .catch((e: unknown) => {
+      console.log(e);
+      res.status(500).json("Internal error");
+    });
+});
```
jberdine: Am I confused, or does this swap the port numbers? I expected the POST to the control port 8081 and the redirect to the streaming port 8080.

skipruntime-ts/helpers/src/rest.ts
```diff
@@ -2,14 +2,15 @@ import type { Json, Entry } from "@skipruntime/api";
 
 export type Entrypoint = {
   host: string;
-  port: number;
+  streaming_port: number;
```
jberdine, line 5: The streaming_port is not used. I think it would be clearer if the constructor of RESTWrapperOfSkipService did not accept it, to emphasize that the REST wrapper only operates over the control port.

skipruntime-ts/tests/src/tests.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from "earl";
-
+// Temporary comment external related tests
```
jberdine, line 2: still necessary?

jberdine (commented):
LGTM

skipruntime-ts/examples/database-client.ts
beauby: Correct, see [the WHATWG spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events-intro).

skipruntime-ts/tests/src/tests.ts
beauby: cc @skiplabsdaniel

skipruntime-ts/examples/database-server.ts
beauby: Good catch, I swapped the default ports at some point, will update.

skipruntime-ts/helpers/src/rest.ts
beauby: It is used by `SkipExternalService` (in `helpers/src/remote.ts`). Those APIs need to be simplified though (`RESTWrapperOfSkipService`, `SkipHttpAccessV1`, `SkipExternalService`), and using the `Endpoint` type everywhere doesn't seem right. Do you mind if we address that in a separate PR?

jberdine: Leaving for later is no problem. FWIW I only noticed this because reading the code using this I got confused about why it was receiving the streaming port.

skipruntime-ts/examples/database-client.ts
jberdine: Thanks for the pointer. For (my) future reference, the EventSource constructor is [specified](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface) at step 15 to perform a "fetch", and it is step 6 of the [fetch spec](https://fetch.spec.whatwg.org/#http-fetch) that states that such redirects are implicitly refetched.

PR #519, opened 2024-11-22
Clone object proxies in mapper/resource params

bennostein:
This was causing some issues in one of my examples, as pointers copied over through proxy objects were outliving the actual object in the skip heap. Thanks Daniel for the help debugging!

### DIFF
--- a/skipruntime-ts/wasm/src/internals/skipruntime_module.ts
+++ b/skipruntime-ts/wasm/src/internals/skipruntime_module.ts
@@ -37,6 +37,7 @@ import { NonUniqueValueException } from "@skipruntime/api";
 import { Frozen, type Constant } from "@skipruntime/api/internals.js";
 
 import type { Exportable, SKJSON } from "@skip-wasm/json";
+import { isObjectProxy } from "@skip-wasm/json";
 import { UnknownCollectionError } from "@skipruntime/helpers/errors.js";
 
 export type Handle<T> = Internal.Opaque<int, { handle_for: T }>;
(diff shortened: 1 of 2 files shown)


skipruntime-ts/wasm/src/internals/skipruntime_module.ts
```diff
@@ -64,22 +65,20 @@ abstract class SkFrozen extends Frozen {
...
-    } else {
-      throw new Error("Invalid object: must be deep-frozen.");
-    }
-  } else {
-    throw new Error(`'${typeof value}' cannot be deep-frozen.`);
+  )
+    return value;
+  if (typeof value == "object") {
+    if (value === null) return value;
+    if (isObjectProxy(value)) return value.clone() as T;
```
jberdine, line 77: Why call `clone` here instead of in `deepFreeze`? Do we need a distinct clone for each time it gets `check`ed? If it needs to be here, then `check` is confusingly misnamed.

bennostein: My understanding (@skiplabsdaniel is the authority) is that this is a distinct issue from deep-freezing: these proxy objects are frozen, but when they're used as parameters to mappers and resources, their skip-heap pointers can outlive the lifetime of the objects they point to.  So it's not always necessary to clone when freezing: only at this point when the proxy is used as a parameter.

Agreed on the naming point, will change.

jberdine: Ok, but I don't understand. :-) Is it a pointer stored in the skip heap to a proxy object in the js heap that is the problem? Is the proxy in the js heap really just supposed to be a wrapper for a skip object? So would it make sense to have the skip heap hold a pointer the the proxied skip object instead?

It isn't clear how calling `clone` here will make a difference to the relative lifetimes. I worry that this is a bug waiting to happen since the `clone` call looks like a nop and someone could easily remove it later.

skiplabsdaniel (approved):

PR #527, opened 2024-11-22
Add helper function to get stream UUIDs from wrapped Skip service

bennostein:
See second commit for payoff in client code

### DIFF
--- a/skipruntime-ts/examples/utils.ts
+++ b/skipruntime-ts/examples/utils.ts
@@ -25,7 +25,7 @@ export class SkipHttpAccessV1 {
 
   constructor(
     private streaming_port: number = 8080,
-    private control_port: number = 8081,
+    control_port: number = 8081,
   ) {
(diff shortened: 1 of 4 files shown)


skipruntime-ts/examples/utils.ts
```diff
@@ -25,7 +25,7 @@ export class SkipHttpAccessV1 {
 
   constructor(
     private streaming_port: number = 8080,
-    private control_port: number = 8081,
+    control_port: number = 8081,
```
beauby, line 28: Is this change needed?

beauby (approved):
LGTM

skipruntime-ts/examples/utils.ts
bennostein: Yes, otherwise that's recognized as an unused field by our JS tooling.  Without `private`, it's just a local variable argument of the constructor.

PR #533, opened 2024-11-25
[skipruntime] Manage resource sharing

### LINKED ISSUE #522: Re-enable tests for external services
_Originally posted by @jberdine in https://github.com/SkipLabs/skip/pull/515#discussion_r1853970844_

skiplabsdaniel:
Close #522

### DIFF
--- a/skipruntime-ts/native/src/Runtime.sk
+++ b/skipruntime-ts/native/src/Runtime.sk
@@ -71,7 +71,15 @@ class ConvReducer(
 
 class ResourceCollection(value: Collection) extends SKStore.File
 
-class ResourceDef(name: String, params: Params) extends SKStore.File
+class ResourceDef(
+  name: String,
+  params: Params,
+) extends SKStore.File, SKStore.Key {
+  //
+  fun toString(): String {
+    `${this.name}:${paramsToString(this.params.value)}`
+  }
+}
 
 class ResourceStatus(
   loadable: SKStore.DirName,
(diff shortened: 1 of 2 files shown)


skipruntime-ts/native/src/Extern.sk
```diff
@@ -607,7 +607,9 @@ fun getAllOfRuntime(
     }
   | _ ->
     SKStore.runWithResult(context ~> {
-      getAll(context, resource, params(jsonParams), optRequest)
+      res = getAll(context, resource, params(jsonParams), optRequest);
+      context.update();
```
beauby: Why is this needed now and not before?

skipruntime-ts/native/src/Extern.sk
```diff
@@ -648,7 +650,9 @@ fun getForKeyOfRuntime(
     }
   | _ ->
     SKStore.runWithResult(context ~> {
-      getForKey(context, resource, params(jsonParams), key, optRequest)
+      res = getForKey(context, resource, params(jsonParams), key, optRequest);
+      context.update();
```
beauby: Same question as above.

skipruntime-ts/wasm/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "devDependencies": {
     "@types/mocha": "^10.0.8",
-    "@skipruntime/tests": "^0.0.3",
+    "@skipruntime/tests": "^0.0.4",
```
beauby: This should go in a separate commit.

skipruntime-ts/wasm/src/internals/skipruntime_module.ts
```diff
@@ -1243,7 +1243,7 @@ export class ServiceInstance {
   instantiateResource(
     identifier: string,
     resource: string,
-    params: { [param: string]: string },
+    params: { [param: string]: string } = {},
```
beauby: Is this related to resource sharing? If not let's put it in a different commit.

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -239,16 +247,35 @@ fun buildResourcesGraph(
...
     ResourceInfo::type,
     context,
     kResourceCollectionsDir,
-    (context, writer, key, it) ~> {
+    (context, writer, key, _it) ~> {
       pushContext(context);
       try {
-        resourceId = base64(key.value);
-        def = it.first;
+        resourceId = toResoureId(key.name, key.params.value);
```
beauby:
```suggestion
        resourceId = toResourceId(key.name, key.params.value);
```

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -239,16 +247,35 @@ fun buildResourcesGraph(
...
+    SKStore.IntFile::type,
+    context,
+    dirname.sub("availables"),
+    (_c, writer, key, _it) ~> {
+      // map to ignore array changes
+      // to prevent recompute on each add/remove session
+      writer.set(key, SKStore.IntFile(1))
+    },
+  );
+  _ = availablesHdl.map(
```
beauby, line 270: Why not directly map on `resourcesHdl` and use `it.first`?

skipruntime-ts/native/src/Extern.sk
skiplabsdaniel: I realized that this was missing since the status management, it works without but it would be cleaner with. But it's true that it's not part of this commit.

skipruntime-ts/native/src/Extern.sk
skiplabsdaniel: Same

skipruntime-ts/native/src/Runtime.sk
skiplabsdaniel: Because each source will compute its own graph, it's allow to have have one source force a resource reactive graph.

PR #534, opened 2024-11-25
Add type for running server returned by runService

jberdine:


mbouaziz (approved):

skipruntime-ts/server/src/server.ts
```diff
@@ -2,6 +2,16 @@ import { initService } from "skip-wasm";
 import type { SkipService, NamedCollections } from "skip-wasm";
 import { controlService, streamingService } from "./rest.js";
 
+/**
+ * A running server.
+ */
+export type Server = {
```
jberdine: As part of the exposed API, I wonder if anyone thinks this name is overly general?

beauby: Maybe SkipServer?

beauby (approved):
LGTM

PR #536, opened 2024-11-25
ts: add readonly modifier

mbouaziz:
I realized we don't use `readonly` much, though it's pretty helpful in understanding code.

Like a child who discovers a new word I decided to put it everywhere (it made sense).

Note on style: I turned class properties into parameter properties only when they were all initialized with the same visibility and readonly-ness.

### DIFF
--- a/skiplang/prelude/ts/src/sk_browser.ts
+++ b/skiplang/prelude/ts/src/sk_browser.ts
@@ -2,11 +2,7 @@ import type { float, int, Environment, Wrk, Shared } from "./sk_types.js";
 import { MemFS, MemSys } from "./sk_mem_utils.js";
 
 class WrkImpl implements Wrk {
-  worker: Worker;
-
-  constructor(worker: Worker) {
-    this.worker = worker;
-  }
+  constructor(private readonly worker: Worker) {}
 
   static fromPath(url: URL, options?: WorkerOptions) {
     return new this(new Worker(url, options));
--- a/skipruntime-ts/examples/utils.ts
+++ b/skipruntime-ts/examples/utils.ts
@@ -24,7 +24,7 @@ export class SkipHttpAccessV1 {
   private service: SkipServiceBroker;
 
   constructor(
-    private streaming_port: number = 8080,
+    private readonly streaming_port: number = 8080,
     control_port: number = 8081,
   ) {
     this.service = new SkipServiceBroker({
@@ -118,21 +118,14 @@ interface DeleteQuery {
 export type Step = RequestQuery | LogQuery | WriteQuery | DeleteQuery;
 
 class Session {
-  scenario: Step[];
-  perform: (l: Step) => void;
-  error: (e: string) => void;
   current = 0;
   on = false;
 
   constructor(
-    scenario: Step[],
-    perform: (l: Step) => void,
-    error: (e: string) => void,
-  ) {
-    this.scenario = scenario;
-    this.perform = perform;
-    this.error = error;
-  }
+    private readonly scenario: Step[],
+    private readonly perform: (l: Step) => void,
+    private readonly error: (e: string) => void,
+  ) {}
 
   next(): boolean {
     if (this.current >= this.scenario.length) {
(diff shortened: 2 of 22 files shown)


skiplang/prelude/ts/src/sk_browser.ts
```diff
@@ -2,11 +2,7 @@ import type { float, int, Environment, Wrk, Shared } from "./sk_types.js";
 import { MemFS, MemSys } from "./sk_mem_utils.js";
 
 class WrkImpl implements Wrk {
-  worker: Worker;
-
-  constructor(worker: Worker) {
-    this.worker = worker;
-  }
+  constructor(private readonly worker: Worker) {}
```
jberdine, line 5: FMI, is this just a code style change, or more? If just style, is it preferred?

skipruntime-ts/api/src/internals.ts
```diff
@@ -11,7 +11,7 @@ export interface Constant {
 
 export abstract class Frozen implements Constant {
   // tsc misses that Object.defineProperty in the constructor inits this
-  [sk_frozen]!: true;
+  readonly [sk_frozen]!: true;
```
jberdine: Now this is both readonly and has a type inhabited by only one value. :-)

skipruntime-ts/examples/utils.ts
```diff
@@ -21,10 +21,10 @@ interface Delete {
 }
 
 export class SkipHttpAccessV1 {
-  private service: RESTWrapperOfSkipService;
+  private readonly service: RESTWrapperOfSkipService;
 
   constructor(
-    private streaming_port: number = 8080,
+    private readonly streaming_port: number = 8080,
     control_port: number = 8081,
```
jberdine, line 28: Why not this too?

jberdine (approved):
Seems fine to me. The meaning of `readonly` is shallow, only saying that a local variable is not assigned to, right? In the skipruntime-ts/examples, it might be better to not tag things like `EagerCollection`s as `readonly`, since I don't think it adds anything, and unless users are already very familiar with the shallow semantics, it might cause confusion by hinting at a read-only collection, when in fact they commonly can be written to from elsewhere.

I wonder, how trustworthy is `readonly`? Do you have experience failing to break it?

skiplang/prelude/ts/src/sk_browser.ts
mbouaziz: Two things:
- code style change to _parameter properties_ which I find simpler (the type is written once only) and clearer, it resembles Skip constructors btw;
- added `private` and `readonly`

skipruntime-ts/api/src/internals.ts
mbouaziz: We're never too sure with Typescript :grin:

skipruntime-ts/examples/utils.ts
mbouaziz: Because of `private`, the parameter `streaming_port` is also a property. `control_port` is not

mbouaziz:
> The meaning of readonly is shallow, only saying that a local variable is not assigned to, right?

Yes, only saying that a property is not assigned to after the object is constructed.

> In the skipruntime-ts/examples, it might be better to not tag things like EagerCollections as readonly, since I don't think it adds anything, and unless users are already very familiar with the shallow semantics, it might cause confusion by hinting at a read-only collection, when in fact they commonly can be written to from elsewhere.

Right, I'll revert this part.

> I wonder, how trustworthy is readonly? Do you have experience failing to break it?

It's only a typing thing and it disappears at runtime. It doesn't bring guarantees outside Typescript. I got warned a few times while writing the diff.
Actually I should also add `readonly` to `type`s and `interface`s because otherwise it's too easy to break:
```typescript
class A {
    constructor (readonly f: number) {}
};
const x = new A(0);
x.f = 1; // type error
const y : { f : number} = x;
y.f = 2; // totally fine 😢
```

jberdine:
> Actually I should also add `readonly` to `type`s and `interface`s because otherwise it's too easy to break:

You feel that it's all worthwhile? I don't have a strong view either way, but it's not obvious to me.

mbouaziz:
To avoid any fake readonly guarantees, I kept it only for private fields.

It could still make sense to add more `readonly` for parameters as a way to document we're not mutating them, but that could be too verbose

jberdine (approved):

PR #537, opened 2024-11-25
Add streamURL helper function

jberdine:
To better encapsulate details of the HTTP interface in the wrapper class.

### DIFF
--- a/skipruntime-ts/examples/database-server.ts
+++ b/skipruntime-ts/examples/database-server.ts
@@ -41,7 +41,7 @@ app.get("/users", (_req, res) => {
   service
     .getStreamUUID("users")
     .then((uuid) => {
-      res.redirect(301, `http://localhost:8080/v1/streams/${uuid}`);
+      res.redirect(301, service.streamURL(uuid));
     })
     .catch((e: unknown) => {
       console.log(e);
--- a/skipruntime-ts/examples/groups-server.ts
+++ b/skipruntime-ts/examples/groups-server.ts
@@ -16,7 +16,7 @@ app.get("/active_friends/:uid", (req, res) => {
   service
     .getStreamUUID("active_friends", { uid: req.params.uid })
     .then((uuid) => {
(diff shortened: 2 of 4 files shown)


bennostein (approved):
Nice, looks good to me

skipruntime-ts/examples/database-server.ts
```diff
@@ -41,7 +41,7 @@ app.get("/users", (_req, res) => {
   service
     .getStreamUUID("users")
     .then((uuid) => {
-      res.redirect(301, `http://localhost:8080/v1/streams/${uuid}`);
+      res.redirect(301, service.streamURL(uuid));
```
beauby, line 44: In a real world scenario, the `streamURL()` itself will not be exposed as is to the outside world (usually some URL rewriting will take place in the reverse proxy, for instance mapping `https://mywebsite.com/api/sse/{uuid}` to `http://192.168.42.37:8081/v1/streams/{uuid}`. Redirecting to `streamURL()` will issue an HTTP redirect to the latter, which may not be accessible from the client.

Edit: it is indeed possible to have the reverse proxy rewrite the `Location` header in the response, but it seems more error prone than having the web app directly deal with public URLs.

jberdine: To check my understanding: So you mean that when people write their client-facing server code, such as this database-server.ts example, then they will want to send redirect URLs to the client code that will then be rewritten by the reverse proxy to the actual URL of the reactive skip service? So it is the reverse proxy configuration that will need to talk about the actual URL of the stream. In which case encapsulating these details just can't work.

beauby: > So you mean that when people write their client-facing server code, such as this database-server.ts example, then they will want to send redirect URLs to the client code that will then be rewritten by the reverse proxy to the actual URL of the reactive skip service?

That's one option. Basically, the reactive service itself does not know its own public url, which makes it unable to produce public URLs for streams.

PR #538, opened 2024-11-25
Change EagerCollection `slice` behavior to match JS `Array.prototype.slice`

bennostein:
Julien mentioned running into some user confusion on this, so this renames our previous `slice` (which offered general skstore power of multiple key ranges) to `slices` and gives `slice` the simpler semantics of JS arrays' `slice` method.


mbouaziz (approved):
Makes sense, lgtm

Issue #535, opened 2024-11-25
Consider using POST instead of GET for resource reads


jberdine:
To expand on my previous reply, we could replace `GET /v1/resource/:resource` with a `POST /v1/snapshot` passing the `resource` and `params` (and possibly a set of `keys`) as a JSON object in the request body. This would also stress the fact that doing a sync read does indeed involve instantiating a resource.

_Originally posted by @beauby in https://github.com/SkipLabs/skip/pull/514#discussion_r1855200771_

PR #539, opened 2024-11-25
Restrict `Resource` output to string-keyed collections

bennostein:
The default REST server implementation passes HTTP path parameters straight through to Skip, which caused me a bit of confusion when resources have non-string keys.

We may avoid putting key parameters straight into the path in the future, but for now this prevents the confusion I ran into (i.e. requesting a path like `/users/1` and getting nothing because the user is keyed on the number `1` and not the string `"1"`)

### DIFF
--- a/examples/hackernews/reactive_service/src/hackernews.service.ts
+++ b/examples/hackernews/reactive_service/src/hackernews.service.ts
@@ -86,11 +86,11 @@ class SortingMapper {
   mapEntry(
     key: number,
     values: NonEmptyIterator<Upvoted>,
-  ): Iterable<[number, Upvoted]> {
+  ): Iterable<[string, Upvoted]> {
     const posts = values.toArray();
     // Sorting in descending order of upvotes.
     posts.sort((a, b) => b.upvotes - a.upvotes);
-    return posts.map((p) => [key, p]);
+    return posts.map((p) => [key.toString(), p]);
   }
 }
 
@@ -101,7 +101,7 @@ class PostsResource implements Resource<ResourceInputs> {
     this.limit = Number(params["limit"]);
   }
 
-  instantiate(collections: ResourceInputs): EagerCollection<number, Upvoted> {
+  instantiate(collections: ResourceInputs): EagerCollection<string, Upvoted> {
     return collections.postsWithUpvotes.take(this.limit).map(SortingMapper);
   }
 }
--- a/skipruntime-ts/examples/departures.ts
+++ b/skipruntime-ts/examples/departures.ts
@@ -23,7 +23,7 @@ class DeparturesResource implements Resource<ResourceInputs> {
   instantiate(
     cs: ResourceInputs,
     context: Context,
-  ): EagerCollection<number, Departure> {
+  ): EagerCollection<string, Departure> {
     const get = (name: string, def: string) => {
       try {
         return cs.config.getUnique(name).join(",");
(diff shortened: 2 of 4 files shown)


skipruntime-ts/tests/src/tests.ts
```diff
@@ -341,17 +357,17 @@ const mergeReduceService: SkipService<Input_NN_NN, Input_NN_NN> = {
...
 
   mapEntry(
     key: number,
     values: NonEmptyIterator<{ value: JsonObject; pattern: string }>,
-  ): Iterable<[number, Json[]]> {
+  ): Iterable<[string, Json[]]> {
     const value = values.getUnique();
     const result = this.context.jsonExtract(value.value, value.pattern);
-    return Array([key, result]);
+    return Array([JSON.stringify(key), result]);
```
beauby:
```suggestion
    return Array([key.toString(), result]);
```

examples/hackernews/reactive_service/src/hackernews.service.ts
```diff
@@ -86,11 +86,11 @@ class SortingMapper {
   mapEntry(
     key: number,
     values: NonEmptyIterator<Upvoted>,
-  ): Iterable<[number, Upvoted]> {
+  ): Iterable<[string, Upvoted]> {
     const posts = values.toArray();
     // Sorting in descending order of upvotes.
     posts.sort((a, b) => b.upvotes - a.upvotes);
-    return posts.map((p) => [key, p]);
+    return posts.map((p) => [JSON.stringify(key), p]);
```
beauby:
```suggestion
    return posts.map((p) => [key.toString(), p]);
```

skipruntime-ts/examples/departures.ts
```diff
@@ -58,7 +58,8 @@ const service = await runService(
         departuresFromAPI: new Polled(
           "https://api.unhcr.org/rsq/v1/departures",
           10000,
-          (data: Result) => data.results.map((v, idx) => [idx, [v]]),
+          (data: Result) =>
+            data.results.map((v, idx) => [JSON.stringify(idx), [v]]),
```
beauby: `idx` here is guaranteed to be a `number`.
```suggestion
            data.results.map((v, idx) => [idx.toString(), [v]]),
```

beauby (approved):

jberdine:
Question: would changing from GET to POST a la https://github.com/SkipLabs/skip/issues/535, which would put the resources and params into the request body as json instead of being encoded in the URL, have avoided the original confusion?

bennostein:
> Question: would changing from GET to POST a la #535, which would put the resources and params into the request body as json instead of being encoded in the URL, have avoided the original confusion?

As I understand it, yes!  Lucas mentioned (offline) wanting to do that at some point too, and in particular emphasizing that a sync read instantiates a resource, which could be misunderstood under the current scheme where the key is a path parameter.

jberdine:
> > Question: would changing from GET to POST a la #535, which would put the resources and params into the request body as json instead of being encoded in the URL, have avoided the original confusion?
> 
> As I understand it, yes! Lucas mentioned (offline) wanting to do that at some point too, and in particular emphasizing that a sync read instantiates a resource, which could be misunderstood under the current scheme where the key is a path parameter.

Ok good. This makes me wonder if we should make the POST change instead of reducing the generality of the key type.

bennostein:
Seems reasonable to me.  I'll table this for the time being, but still going to keep working on docs-writing, not implement that POST change.

PR #542, opened 2024-11-26
Add deleteUUID helper function

jberdine:


jberdine:
I'm working on docs for RESTWrapperOfSkipService and this function seems to be missing compared to the http interface of the service started by runService. But thinking about adding calls in the examples, I'm not sure when this function should be called. Is there an automatic mechanism where e.g. when EventSource.close is called, the close of the connection triggers a delete of the resource instance, or should user code be triggering this explicitly?

bennostein:
I think Daniel has been working on a garbage collection mechanism to clean up resource instances, but that may be out-of-date or misremembered...

In any case, no problem to have this in the API as far as I can see.

bennostein (approved):

beauby:
> I'm working on docs for RESTWrapperOfSkipService and this function seems to be missing compared to the http interface of the service started by runService. But thinking about adding calls in the examples, I'm not sure when this function should be called. Is there an automatic mechanism where e.g. when EventSource.close is called, the close of the connection triggers a delete of the resource instance, or should user code be triggering this explicitly?

Explicit destruction of a live stream is expected to happen in exceptional situations (an example would be having an out-of-band mechanism to detect malicious users, and stop feeding them data immediately). In the general case, a stream will not get destroyed immediately upon disconnection to allow fast reconnect after transient network errors. Instead, it will be garbage collected after some inactivity (Daniel is working on this as Benno mentioned).

PR #543, opened 2024-11-26
Rename RESTWrapperOfSkipService to SkipServiceBroker and document it

jberdine:
- Rename RESTWrapperOfSkipService to SkipServiceBroker

This class provides methods that mirror and have thin implementations over
the skip service HTTP interface. This class is providing a method-call
interface for the HTTP interface. In this context, the term "broker" seems
more appropriate and clear.

- Document SkipServiceBroker

### DIFF
--- a/skipruntime-ts/helpers/src/rest.ts
+++ b/skipruntime-ts/helpers/src/rest.ts
@@ -41,7 +41,7 @@ export async function fetchJSON<V>(
   return [responseJSON, response.headers];
 }
 
-export class RESTWrapperOfSkipService {
+export class SkipServiceBroker {
   private entrypoint: string;
 
   constructor(
--- a/skipruntime-ts/helpers/src/rest.ts
+++ b/skipruntime-ts/helpers/src/rest.ts
@@ -41,9 +41,20 @@ export async function fetchJSON<V>(
   return [responseJSON, response.headers];
 }
 
+/**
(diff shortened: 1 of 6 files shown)


skipruntime-ts/helpers/src/rest.ts
```diff
@@ -79,30 +104,55 @@ export class RESTWrapperOfSkipService {
...
+   * @returns {void}
+   */
   async put<K extends Json, V extends Json>(
     collection: string,
     key: K,
-    value: V[],
+    values: V[],
   ): Promise<void> {
-    return await this.patch(collection, [[key, value]]);
+    return await this.patch(collection, [[key, values]]);
```
jberdine, line 120: The HTTP interface supports PUT, so I don't know why this is not implemented using it. Also, deleteKey could be implemented using PUT as well. Is there any reason to prefer the PATCH implementations? @beauby ?

beauby: It might be to circumvent the key type issue. @skiplabsdaniel?

jberdine: Ah, that looks related. I guess we should remove PUT when switching GET to POST.

jberdine:
Does anyone have comments on renaming RESTWrapperOfSkipService to SkipServiceBroker?

bennostein:
Sounds good to me!

bennostein (approved):

PR #544, opened 2024-11-26
[www/docs] External services

bennostein:
* Add some prose documentation of external services
 * Update markdown file names and skeletons to match current state of affairs


www/docs/externals.md
```diff
@@ -0,0 +1,113 @@
...
+However, when those inputs come from external systems or services it is crucial to ensure that your reactive service still produces up-to-date results.
+
+Skip provides mechanisms to do so easily, whether those external systems are other reactive Skip services or non-reactive systems like databases or external APIs.
+
+The core of these mechanisms is the [`ExternalService` type](todo: API docs link), which provides a generic interface which can be implemented to wrap arbitrary external systems for use in your Skip service.
+Each Skip reactive service can specify any number of `ExternalService`s, which can then be brought into the reactive computation graph as an `EagerCollection` with [`Context#useExternalResource`](todo: API docs link).
+
+Skip provides two `ExternalService` implementations -- one which polls external HTTP services, and one which subscribes to external Skip services.
+If your use case falls outside of these defaults, you can define your own custom external service by providing another `ExternalService` implementation with whatever arbitrary behavior is required.
```
jberdine:
```suggestion
If your use case falls outside of these defaults, you can define your own custom external service by providing another `ExternalService` implementation with the required behavior.
```

www/docs/externals.md
```diff
@@ -0,0 +1,113 @@
...
+
+## External Skip Services
+
+Skip reactive services are designed to interoperate simply and easily: services push their changes reactively from the bottom-up, eagerly propagating updates through a distributed system.
```
jberdine: not used as an adjective or adverb here
```suggestion
Skip reactive services are designed to interoperate simply and easily: services push their changes reactively from the bottom up, eagerly propagating updates through a distributed system.
```

www/docs/externals.md
```diff
@@ -0,0 +1,113 @@
...
+externalData.map(...);
+externalData.getArray(...);
+```
+
+When the external resource is "used" in this fashion, Skip queries the HTTP endpoint with the given `params` periodically, updating `externalData` with the results and allowing it to be used as if it were a reactive resource.
+The frequency of those requests is an important consideration, depending on the latency requirements of the application as well as the load capacity of the external system.
+
+### Custom External Resources/Services
+
+If more control is required or your external system does not fit into this paradigm, then the `ExternalService` and `ExternalResource` interfaces both support extensions to define arbitrary logic for `subscribe`/`unsubscribe`-ing from services and `open`/`close`-ing resource.
```
jberdine: "does not fit into this paradigm" seems too strong to me:
```suggestion
If more control is required or your external system does not fit this form, then the `ExternalService` and `ExternalResource` interfaces both support extensions to define arbitrary logic for `subscribe`/`unsubscribe`-ing from services and `open`/`close`-ing resource.
```

www/docs/externals.md
```diff
@@ -0,0 +1,113 @@
...
+
+For example, your service can keep track of open resources and combine requests to the external service if it supports batched requests, lessening request load and potentially allowing more optimized execution in the non-reactive system.
```
jberdine:
```suggestion
For example, your service can keep track of open resources and combine requests to the external service if it supports batched requests, reducing request load and potentially allowing more optimized execution in the non-reactive system.
```

jberdine (commented):
LGTM, a couple comments inline. I'm not sure about the changes to reference.md and resources.md, I think they are assuming different meanings of "reference" and "resources" than originally intended.

One point about the docs themselves is that I have not yet fully understood how params work with externals. If you understand that, adding something here would be great. I'm also not sure if the example code snippets here are self-consistent regarding the example params and where they get passed to.

jberdine (approved):

bennostein:
> LGTM, a couple comments inline.

Thanks for the prose comments!  all look good to me

> I'm not sure about the changes to reference.md and resources.md, I think they are assuming different meanings of "reference" and "resources" than originally intended.

Yeah, that part of this diff is trying to address the fact that we had `resources.md` referring to "resources" in the generic sense of reference materials, but now we also have a technical sense of "resource" which needs documenting.

So I moved what was previously called `resources.md` into `reference.md` and set up an outline in `resources.md` for documenting resources (in that technical sense).
 
> One point about the docs themselves is that I have not yet fully understood how params work with externals. If you understand that, adding something here would be great. I'm also not sure if the example code snippets here are self-consistent regarding the example params and where they get passed to.

I'll add in a bit more detail about params. But, I'm not sure what the self-consistency issue is?  maybe I'm missing something but this is how I understand params to work (modulo the elided bits and surrounding context).

That is, they're passed as part of the call to `useExternalResource`, so each such call to that can have different parameters.  I suppose we could also accept some parameters in the service definition, to be merged with the per-instantiation params, but I don't get the sense that's what you mean.

jberdine:
> > I'm not sure about the changes to reference.md and resources.md, I think they are assuming different meanings of "reference" and "resources" than originally intended.
> 
> Yeah, that part of this diff is trying to address the fact that we had `resources.md` referring to "resources" in the generic sense of reference materials, but now we also have a technical sense of "resource" which needs documenting.
> 
> So I moved what was previously called `resources.md` into `reference.md` and set up an outline in `resources.md` for documenting resources (in that technical sense).

Ok. Keeping "resources" for the notion in the code makes sense. I'm unsure of using the term "reference" for other info e.g. tutorials since such material is not "reference" material.

> > One point about the docs themselves is that I have not yet fully understood how params work with externals. If you understand that, adding something here would be great. I'm also not sure if the example code snippets here are self-consistent regarding the example params and where they get passed to.
> 
> I'll add in a bit more detail about params. But, I'm not sure what the self-consistency issue is? maybe I'm missing something but this is how I understand params to work (modulo the elided bits and surrounding context).
> 
> That is, they're passed as part of the call to `useExternalResource`, so each call to that can have different parameters. I suppose we could also accept some parameters in the service definition, to be merged with the per-instantiation params, but I don't get the sense that's what you mean.

As I mentioned, I don't think I understand params of externals yet, so I could be just confused. But what I meant is that each external will, from its definition, expect some params, and each instantiation will provide some, but it isn't clear to me that in the examples the code for the expecting and providing places always agrees. I'm not sure though.

PR #558, opened 2024-11-28
[docs] Client markdown documentation

bennostein:
Not a lot to say here, really -- using SSEs means that there's not a ton of Skip-specific work to do on the client.

So, I dive quickly into the actual event-stream data sent over the wire, give some examples of how to work with it in client code, and also show how synchronous reads work.

### DIFF
--- a/www/docs/client.md
+++ b/www/docs/client.md
@@ -74,6 +74,15 @@ useEffect(() => {
 
 ## Synchronous HTTP interface
 
+Skip reactive services also support synchronous (i.e. non-reactive) reads of resources, either in their entirety or at a specific key.
+Note that this requires instantiation of the resource, just the same as reactive streaming would.
+As such, there is some overhead to synchronous reads and they are generally not needed in systems built using reactive services.
+They are useful, however, for debugging the state of reactive systems and for maintaining compatibility with legacy systems and non-reactive clients.
+
+To make a synchronous read, call either `getAll` or `getArray` on the reactive service handle in your web service to query the corresponding routes on the reactive service; an example using `getArray` is given in the example Express web service below.
+
+Then, from your client, issue HTTP `GET` requests to e.g. `http://reactive.service.hostname/my_resource/foo/key1` to make a synchronous read of data in the `foo` resource instance associated with key `key1`.
+
 # Example web service configuration
 
 Skip reactive services instantiate resources on request, generating a unique UUID for each unique query and serving the resulting stream of updates over HTTP.
@@ -96,17 +105,26 @@ const reactive_service = new SkipServiceBroker({
   control_port,
 });
 
-// On receiving a GET request for a resource instance,
+// On receiving a GET request for a reactive resource instance,
 //  1. Parse out any params from the request
 //  2. Request a stream identifier with those parameters from the reactive service
 //  3. Redirect the client to the corresponding stream address
 app.get("/my_resource/:id", (req, res) => {
-  const params = { id: req.params.id, foo: req.params.bar, ...};
+  const params = { id: req.params.id, foo: req.params.bar, ... };
   reactive_service
     .getStreamUUID("my_resource", params)
     .then((uuid) => {
       res.redirect(301, `http://${reactive_host}:${streaming_port}/v1/streams/${uuid}`);
     })
 });
-```
 
+// Synchronous read of a specific key in a resource instance
+app.get("/my_resource/:id/:key", (req, res) => {
+  const params = { id: req.params.id, foo: req.params.bar, ... };
+  reactive_service
+    .getArray("my_resource", params, req.params.key)
+    .then((data) => {
+      res.status(200).json(data);
+    });
+});
+```
--- a/www/docs/client.md
+++ b/www/docs/client.md
@@ -1,17 +1,17 @@
 # Client connections
 
 Clients can subscribe to updates from a reactive service using the widely-available [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) protocol, or make synchronous read requests for specific data.
-This means that Skip clients are thin and lightweight: all major browsers implement the requisite `EventSource` interface, and other types of clients can easily implement parsers for the simple text-based protocol.
+This means that Skip clients are thin and lightweight: the JavaScript `EventSource` interface is widely implemented, and other types of clients can easily implement parsers for the simple text-based protocol.
 
-Note that this page describes mechanisms for front-end clients or non-Skip systems to access the outputs of a Skip service; to communicate among Skip services it is both simpler and more efficient to use [external dependencies](todo: docusaurus link).
+Note that this page describes mechanisms for front-end clients or non-Skip systems to access the outputs of a Skip service; to communicate among Skip services it is both simpler and more efficient to use [external dependencies](externals.md).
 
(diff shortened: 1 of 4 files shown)


www/docs/client.md
```diff
@@ -1,9 +1,130 @@
 # Client connections
 
-## Overview
+Clients can subscribe to updates from a reactive service using the widely-available [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) protocol, or make synchronous read requests for specific data.
+This means that Skip clients are thin and lightweight: all major browsers implement the requisite `EventSource` interface, and other types of clients can easily implement parsers for the simple text-based protocol.
```
jberdine: I would avoid "requisite" since AFAIU it is not _required_, and it is implemented not just in browsers:
```suggestion
This means that Skip clients are thin and lightweight: the `EventSource` interface is widely implemented, and other types of clients can easily implement parsers for the simple text-based protocol.
```

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
 
-## Skip Listener
+Note that this page describes mechanisms for front-end clients or non-Skip systems to access the outputs of a Skip service; to communicate among Skip services it is both simpler and more efficient to use [external dependencies](todo: docusaurus link).
```
jberdine: I guess we should settle on a term for "page" / "section" / "guide" for these docs.
```suggestion
Note that this page describes mechanisms for front-end clients or non-Skip systems to access the outputs of a Skip service; to communicate among Skip services it is both simpler and more efficient to use [external dependencies](externals.md).
```

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
 
-## Websockets
+By design, client code does not need to import or reason about the internal details of Skip reactive services; this page abstracts over those details, but for completeness's sake we provide a simple explanatory example [below](todo: intrapage link to "Example web service configuration") of a backend service which can support such clients.
```
jberdine:
```suggestion
By design, client code does not need to import or reason about the internal details of Skip reactive services; this page abstracts over those details, but for completeness's sake we provide a simple explanatory example [below](client#example-web-service-configuration) of a backend service which can support such clients.
```

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
 
-## Run Skip client-side
+## Event Streams
+
+Clients of reactive services can opt to receive pushed data updates using HTTP server-sent events.
+
+The simplest way to see the data that your client will receive is to use `curl`, e.g. `curl -LN http://reactive.service.hostname/my_resource/foo` to stream from `my_resource` with parameter `foo`, which will produce output like the following:
```
jberdine: I don't understand how this URL is correct. Maybe for brevity just remove mention of the parameter. The stumbling point is, I guess, that parameters are specified when obtaining the uuid of a stream, but not when opening the sse stream for that uuid.

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+data: [["key2",["value2"]]]
+
+event: update
+id: 1065371604
+data: [["key3",["value4"]]]
+```
+
+The `init` event contains all available data for the resource when the connection is created, and can be used to set up initial client state.
+
+Subsequent `update` events indicate changes _only_ at the included keys: the first update here removes `"value3"` from `"key2"`, while the second removes adds a new binding from `"key3"` to `"value4"`.
```
jberdine:
```suggestion
Subsequent `update` events indicate changes _only_ at the included keys: the first update here removes `"value3"` from `"key2"`, while the second adds a new binding from `"key3"` to `"value4"`.
```
Note that in the API, the term "entry" is used instead of "binding" for "associating a key to one or more values".

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+
+All events include an `id` metadata which should mostly be ignored, but can be useful for debugging, replay, and similar.
```
jberdine:
```suggestion
All events include an `id` metadata which should mostly be ignored, but can be useful for debugging, replay, and similar purposes.
```

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+    stream.close();
+  };
+}, [])
+```
+
+## Synchronous HTTP interface
+
+Skip reactive services also support synchronous (i.e. non-reactive) reads of resources, either in their entirety or at a specific key.
+Note that this requires instantiation of the resource, just the same as reactive streaming would.
+As such, there is some overhead to synchronous reads and they are generally not needed in systems built using reactive services.
```
jberdine: Re overhead, yes, although reading from a resource with the same params will reuse a recent instance, so is hopefully not too bad. Not sure we want to call out the overhead.

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+They are useful, however, for debugging the state of reactive systems and for maintaining compatibility with legacy systems and non-reactive clients.
+
+To make a synchronous read, call either `getAll` or `getArray` on the reactive service handle in your web service to query the corresponding routes on the reactive service; an example using `getArray` is given in the example Express web service below.
```
jberdine: Question: Do you prefer "handle" over "broker"? Should we rename `SkipServiceBroker`?

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+
+Then, from your client, issue HTTP `GET` requests to e.g. `http://reactive.service.hostname/my_resource/foo/key1` to make a synchronous read of data in the `foo` resource instance associated with key `key1`.
```
jberdine, line 84: I don't know how understandable this sentence is without first seeing the `app.get` code below for handling it. But I don't have a good suggested change right now.

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+
+# Example web service configuration
+
+Skip reactive services instantiate resources on request, generating a unique UUID for each unique query and serving the resulting stream of updates over HTTP.
```
jberdine:
```suggestion
Skip reactive services instantiate resources on request, generating a UUID for each distinct query and serving the resulting stream of updates over HTTP.
```

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+  host: reactive_host,
+  streaming_port,
+  control_port,
+});
+
+// On receiving a GET request for a reactive resource instance,
+//  1. Parse out any params from the request
+//  2. Request a stream identifier with those parameters from the reactive service
+//  3. Redirect the client to the corresponding stream address
+app.get("/my_resource/:id", (req, res) => {
```
jberdine, line 112: Ok, now I understand the things I commented on earlier in the file. I guess I am just in a definition-before-use mental state.

www/docs/client.md
```diff
@@ -1,9 +1,130 @@
...
+// Synchronous read of a specific key in a resource instance
+app.get("/my_resource/:id/:key", (req, res) => {
+  const params = { id: req.params.id, foo: req.params.bar, ... };
+  reactive_service
+    .getArray("my_resource", params, req.params.key)
+    .then((data) => {
+      res.status(200).json(data);
+    });
+});
+```
```
jberdine, line 130: I think it would make sense to expand this example and explanation to cover sync writes to a collection.

jberdine (approved):
LGTM, some comments but nothing blocking

www/docs/client.md
bennostein: No strong feeling! I used "handle" here as it felt more natural than "broker", but "broker" does seem more precise/correct for the actual class in the code.

jberdine: Yeah, I see what you mean about "natural". I think it will help people to always use the same terms though.

PR #560, opened 2024-11-28
Random minor cleanups

mbouaziz:


jberdine (approved):
LGTM. My only question is whether Uint8Array and ArrayBuffer are mutually interchangeable enough that we don't care about just unioning their types.

mbouaziz:
>  My only question is whether Uint8Array and ArrayBuffer are mutually interchangeable enough that we don't care about just unioning their types.

`Uint8Array` is a view on an `ArrayBuffer` but I don't think the types are mutually interchangeable

PR #561, opened 2024-11-28
Make `resources` field non-optional on SkipService

bennostein:
As it says on the tin.

I noticed this while writing up resource/service docs, and couldn't think of any case where it makes sense to have a service with no output resources.  Nor does any such case exist in our examples.


jberdine (approved):
A service with no resources would have to be write-only, which I guess is enough of a special case to make users pass an empty dict or resources explicitly. :-)

beauby (approved):

PR #562, opened 2024-11-28
[RFC] Starting the debugging API

mbouaziz:
Here is a start for a debugging API. You can test it like that:
```sh
cd skipruntime-ts
bun run build
bun run examples/groups.ts
```
```sh
curl http://localhost:8081/v1/debug/raw/dirs | jq
```
which gives me [before.json](https://github.com/user-attachments/files/17949298/before.json)

After running the example server and client, I get: [after.json](https://github.com/user-attachments/files/17949305/after.json)

Design discussion points:
- URL: I plan on having SKStore-level stuff under `/debug/raw/` and SkipRuntime-level stuff under `/debug/inputs/`, `/debug/resources/`, `/debug/streams/`, what do you think?
- Security: the debugging entrypoints should be disabled by default (where to put this?), forthcoming non-read-only ones will even be under another flag
- Code separation: we could even put the code under another package so that it doesn't pollute production code, maybe as an optional node package, for now the code should be easy to find by grepping debug
- Data usage: for now the API is non-reactive but part of it can be made reactive later, under a feature flag as well because it could be costly


mbouaziz:
Probably subsumed by what Daniel is doing

PR #569, opened 2024-12-02
[skipruntime-ts] Take control over reducer instantiation to avoid capturing/side effects

bennostein:
It came up in conversation today that reducers ought to be instantiated under skipruntime control like mappers, to prevent accidental capture of mutable state or other side effects in reducers. (users may be particularly likely to capture state there, instead of implementing the purely-functional fold using the accumulator arguments) So, this PR does just that.

The main interesting bit is what to do about the two variable-sized parameter lists (i.e. params to the mapper constructor and the reducer constructor).  The obvious thing to do is to have mapReduce take two explicit arrays which default to empty
`(mapper: new(...mapperParams)=>Mapper<...>, reducer: new(...reducerParams)=>Reducer<...>, mapperParams: MapperParams = [], reducerParams: ReducerParams = [])` but that (rightly) doesn't typecheck, since it would allow to elide params even when MapperParams/ReducerParams are non-empty.  So, we can use overloads to only allow eliding those parameters when the respective constructors take no arguments. But, that's a bit clunky, and leads to the same problem in `map`/`createLazyCollection` if we also take array arguments there for consistency's sake.

Long-winded way of saying: I decided to just curry `mapReduce` so that it can splat out arguments to its mapper and reducer separately.  Seems a nice clean interface to me, but open to other suggestions!

### DIFF
--- a/skipruntime-ts/api/src/api.ts
+++ b/skipruntime-ts/api/src/api.ts
@@ -207,13 +207,32 @@ export interface EagerCollection<K extends Json, V extends Json>
   mapReduce<
     K2 extends Json,
     V2 extends Json,
-    Acc extends Json,
-    Params extends Param[],
+    Accum extends Json,
+    MapperParams extends Param[],
+    ReducerParams extends Param[],
   >(
-    mapper: new (...params: Params) => Mapper<K, V, K2, V2>,
-    reducer: Reducer<V2, Acc>,
-    ...params: Params
-  ): EagerCollection<K2, Acc>;
+    mapper: new (...params: MapperParams) => Mapper<K, V, K2, V2>,
+    reducer: new (...params: ReducerParams) => Reducer<V2, Accum>,
+    mapperParams: MapperParams,
+    reducerParams: ReducerParams,
+  ): EagerCollection<K2, Accum>;
+
+  // Type overloads for `mapReduce`, allowing callers to not pass mapper/reducer parameters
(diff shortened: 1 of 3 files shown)


mbouaziz:
Should we just have a `reduce` function rather than `mapReduce`?

In the longer-term we can optimize such sequences: `map` followed by `reduce`, `slices` followed by `map`, etc.

bennostein:
> Should we just have a `reduce` function rather than `mapReduce`?

I have argued for that in the past as it makes for nicer user code.  But doing so directly comes with a real perf cost, materializing a lot more intermediate collections, so was rejected.  May be worth revisiting and doing some optimizations like you suggest for such sequences of operations!

bennostein:
Lots of offline conversation atm w/ Josh/Lucas/Daniel about adding an abstraction layer between the TS APIs and the actual skstore computation graph, some kind of builder-pattern operating over a graph IR.

No unanimous conclusions, but to summarize:
(1) fusing operations like this and separating our API design from implementation is a nice medium/long term thing to work on which would enable all kinds of cool things
(2) it's probably not worth the effort in the immediate term, there are lots of complexities particularly around dynamic changes to the graph and conditional operations in mappers
(3) we should probably just expose `reduce` now, position `mapReduce` as a special case with better perf, and not worry too much about the extra materializations

jberdine (approved):

PR #571, opened 2024-12-03
Add OneToManyMapper and tweak groups example to use it

bennostein:
Updating the getting-started examples to use the now-available array operations (filter, includes, etc.) now that arrays aren't proxied.  Also tightened up syntax a bit using OneToOneMapper

### DIFF
--- a/skipruntime-ts/examples/groups.ts
+++ b/skipruntime-ts/examples/groups.ts
@@ -2,7 +2,7 @@ import {
   type EagerCollection,
   type InitialData,
   type Resource,
-  OneToOneMapper,
+  OneToManyMapper,
 } from "@skipruntime/api";
 
 import { runService } from "@skipruntime/server";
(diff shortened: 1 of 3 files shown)


skipruntime-ts/examples/groups.ts
```diff
@@ -36,37 +35,28 @@ type ServiceInputs = {
 // Type alias for inputs to the active friends resource
 type ResourceInputs = {
   users: EagerCollection<UserID, User>;
-  actives: EagerCollection<GroupID, UserID>;
+  actives: EagerCollection<GroupID, UserID[]>;
```
jberdine: So each `GroupID` is associated with an array of arrays of `UserID`s, why?

bennostein: It lets the example use OneToOneMapper, cutting down on boilerplate.  Charles was looking for some small example showing the feel of writing reactive code, so I tried to tighten this stuff up as much as possible.

One option that could keep the types as before is to define a `OneToManyMapper` and use that here instead.
I just tried that out and it seems useful in general and a bit nicer here in particular -- wdyt?

jberdine (approved):
I like OneToManyMapper, it fits right into the pattern, and the example is tidy.

PR #572, opened 2024-12-04
[skipruntime] Manage Common Wasm/Native binding

### LINKED ISSUE #471: Node addon for native skip runtime

skiplabsdaniel:
- Split core from wasm runtime specific code
- Add node native addon
- Move some code from helper to core
- Close #471


beauby (approved):
Let's merge this as is, and do some clean up/reorg later.

PR #574, opened 2024-12-04
Expose and document types used by public api

jberdine:


jberdine:
actually wait, I have an update for this one

skipruntime-ts/api/src/internals.ts
```diff
@@ -1,13 +1,34 @@
...
-export interface Constant {
+/**
+ * Values that are either unmodifiable or managed by the Skip Runtime.
+ *
+ * A `Constant` value is either managed by the Skip Runtime, in which case its modifications are carefully tracked by the reactive computation system, or it is deep-frozen, meaning that it cannot be modified and neither can its sub-objects, recursively.
+ * See `deepFreeze` to make an object `Constant`.
+ *
+ * `Constant` values are important because they can be used in code that will be executed by the reactive computation system without introducing the possibility of producing stale or unreproducible results.
+ */
+export type Constant = {
```
jberdine, line 28: To call this change out: I think it is clearer to use `type` over `interface` in cases where user code is not expected to / should not / cannot define its own subtypes.

jberdine: However, the doc generator doesn't do well for `type`s that have methods, so probably the net result is worse. :-(

bennostein (approved):
LGTM across the board, one question about an inline comment.

skipruntime-ts/api/src/internals.ts
bennostein: No methods here though, right? Or am I misunderstunding the issue with the doc generator?

jberdine: Ah, yes, here it is ok, I was getting confused with `Context` which can't be implemented by users, but does have methods.

Issue #578, opened 2024-12-05
Consider renaming the Constant type, and relationship with Param


jberdine:
Documenting the Constant type indicated that it is misnamed, as some inhabitants are in fact non-constant, it is just that their modifications can be tracked by skip.

Additionally, there is little conceptual difference for users between Constant and Param. It seems likely that unifying them in the exposed interface would be an improvement.

bennostein:
Support on both fronts!  I've thought about this for a while but haven't been able to come up with a good name for this type, though...

jberdine:
Also, `NonEmptyIterator` should be renamed to some plural form of whatever `Constant` and `Param` get renamed to. See also https://github.com/SkipLabs/skip/pull/582#discussion_r1876054925.

Issue #577, opened 2024-12-05
Consider whether the `Param` type is too lenient


jberdine:
Currently `Param` accepts
```
  | readonly Param[]
  | { readonly [k: string]: Param };
```
without deep-freezing them, but it is not clear that this is safe. It would be better IMO to initially remove them.

PR #582, opened 2024-12-06
Tighten Param type to enforce deep-freezing

bennostein:
See #577: `readonly` modifiers on some Params only ensures that those arrays/objects aren't mutated _within_ mappers; they can still be mutated outside. This PR tightens the Param type, requiring that any objects/arrays are `sk_frozen`, allowing better runtime enforcement of immutability.

@jberdine and I spoke a bunch about this offline and agreed about the general problem/fix.  @mbouaziz , do you agree?  And @skiplabsdaniel , does this seem reasonable to you? Anything we're failing to consider? 

This also makes #578 more urgent -- I'll try to improve the naming in another PR, but thought it'd be better to keep this separate for now.

### DIFF
--- a/skiplang/skjson/ts/src/skjson.ts
+++ b/skiplang/skjson/ts/src/skjson.ts
@@ -12,7 +12,17 @@ import type {
 import { sk_isObjectProxy } from "@skip-wasm/std";
 import type * as Internal from "./skjson_internal_types.js";
 
+/* Redefining Skip runtime types/symbols to avoid circular module dependency */
 const sk_frozen: unique symbol = Symbol.for("Skip.frozen");
+type Constant = { [sk_frozen]: true };
+function sk_freeze<T extends object>(x: T): T & Constant {
+  return Object.defineProperty(x, sk_frozen, {
+    enumerable: false,
+    writable: false,
+    value: true,
+  }) as T & Constant;
+}
+/* END Skip runtime redefinitions */
 
 export enum Type {
   Undefined,
@@ -110,13 +120,7 @@ function interpretPointer<T extends Internal.CJSON>(
       const array = Array.from({ length }, (_, idx) =>
         interpretPointer(hdl, hdl.access.SKIP_SKJSON_at(aPtr, idx)),
       );
-      return Object.freeze(
-        Object.defineProperty(array, sk_frozen, {
-          enumerable: false,
-          writable: false,
-          value: true,
-        }),
-      ) as Exportable;
+      return sk_freeze(array);
     }
     case Type.Object: {
       const oPtr = hdl.access.SKIP_SKJSON_asObject(ptr);
@@ -279,7 +283,7 @@ export type Exportable =
   | null
   | undefined
   | ObjectProxy<{ [k: string]: Exportable }>
-  | (readonly Exportable[] & { [sk_frozen]: true });
+  | (readonly Exportable[] & Constant);
 
 export interface SKJSON extends Shared {
   importJSON(value: ptr<Internal.CJSON>, copy?: boolean): Exportable;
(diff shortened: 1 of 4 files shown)


skiplabsdaniel (approved):

skiplang/skjson/ts/src/skjson.ts
```diff
@@ -107,9 +107,16 @@ function interpretPointer<T extends Internal.CJSON>(
...
+      const array = Array.from({ length }, (_, idx) =>
         interpretPointer(hdl, hdl.access.SKIP_SKJSON_at(aPtr, idx)),
       );
+      return Object.freeze(
+        Object.defineProperty(array, sk_frozen, {
+          enumerable: false,
+          writable: false,
+          value: true,
+        }),
+      ) as Exportable;
```
jberdine: There is redundancy between this code and `sk_freeze`, which would be good to de-dup. @skiplabsdaniel do you know if this duplication is necessary due to package dependencies?

skiplang/skjson/ts/src/skjson.ts
```diff
@@ -272,7 +279,7 @@ export type Exportable =
   | null
   | undefined
   | ObjectProxy<{ [k: string]: Exportable }>
-  | Exportable[];
+  | (readonly Exportable[] & { [sk_frozen]: true });
```
jberdine: Similarly here, we can't use `Constant` here right?

skipruntime-ts/api/src/api.ts
```diff
@@ -152,31 +150,31 @@ export class NonUniqueValueException extends Error {}
 /**
  * A mutable iterator with at least one element
  */
-export interface NonEmptyIterator<T> extends Iterable<T> {
+export interface NonEmptyIterator<T> extends Iterable<T & Param> {
```
jberdine, line 153: I wasn't expecting this. Doesn't the `T` being a subtype of `Param` in the output iterable come from the type param of `NonEmptyIterator`? This makes it look like `NonEmptyIterator` internally deep-freezes the elements, rather than requiring that the type parameter is already a subtype of `Param`.

skiplang/skjson/ts/src/skjson.ts
bennostein: It is -- that's why I wrote it out here.  Will make a note of that, though

skipruntime-ts/api/src/api.ts
bennostein: My logic throughout this diff is to inform typescript that values coming _out_ of Skip are always safely frozen or primitive -- the outputs of getters, the arguments of mapper functions, etc.
So, since `NonEmptyIterator`s is only ever instantiated with such values, its contents are `Param` -- but, we don't want to enforce that its type param is frozen, since that constraint would propagate through basically every type parameter in the system, even in places where they're not necessarily frozen.

jberdine: Ok, I see that every occurrence of `NonEmptyIterator` in the api is in contravariant position. So while the name gives the impression that this is a slight specialization of standard iterators, it is actually a completely different thing. In fact it is an iterable view of a skip-managed list of skip-managed values. Makes me think we should rename it, but don't have good ideas at the moment.

I worry that since this is just a TS interface, people can define their own implementation of it, and then use it to cast anything to `Param` but just putting it in and taking it back out. But then using `as` people can already do any cast they want.

I wonder if there is some way to write this in TS such that user code cannot define their own implementations, if only to reinforce that these can only be created by skip.

skiplang/skjson/ts/src/skjson.ts
```diff
@@ -12,7 +12,17 @@ import type {
...
 const sk_frozen: unique symbol = Symbol.for("Skip.frozen");
+type Constant = { [sk_frozen]: true };
+function sk_freeze<T extends object>(x: T): T & Constant {
+  return Object.defineProperty(x, sk_frozen, {
+    enumerable: false,
+    writable: false,
+    value: true,
+  }) as T & Constant;
+}
+/* END Skip runtime redefinitions */
```
jberdine, line 25: This is nice, thanks. When doing such things, I also put a comment on the other copy of the definition, so that people changing it have a chance to notice that they should change the copy as well.

jberdine (approved):

PR #583, opened 2024-12-06
Remove docs site landing page

jberdine:
- Remove placeholder content from docs site landing page
- Remove broken and redundant links from header and footer
- Update icon etc. images


jberdine:
I don't really know if this works. The behavior of the preview is to land on the url without "/docs" at the end, but is blank in the preview. The hope is that in production that would end up hitting http://skiplabs.io, but I don't know how to test. @00-pbrstreetgang-00 do you know?

jberdine:
New preview with updated logo: https://docs-site-an45iy2d9-skiplabs-website.vercel.app/docs

jberdine:
I still need to figure out how to set the color scheme

www/src/pages/index.tsx
```diff
@@ -1,45 +1,16 @@
...
-    </Layout>
-  );
-}
+export default () => (
+  <BrowserOnly>
+    {() => {
+      <Route
+        path="/"
+        component={() => {
+          window.location.href = "https://skiplabs.io/";
```
bennostein: This is the key change for the bug we noticed on Friday, right?  i.e. it'll act like a regular non-SPA link to the home page, jumping outside of docusaurus

bennostein (approved):
Looks good to me (though I can't access the vercel preview, just reading through changes)

As for how to test, since it's specifically about the prod deployment, I'd propose just pushing and being ready to roll back if anything looks off...

www/src/pages/index.tsx
jberdine: AFAIU, yes, to get an actual redirect, that does a genuine http fetch, without being hijacked by whatever this SPA business is, one has to resort to something like this that just assigns to a global, but doubly-delayed by both being in a closure and under a BrowserOnly tag in order to hide it well enough that docusaurus can't break it.

jberdine:
No, this doesn't work. I pushed it to prod momentarily to test, but the links don't go back to the main site, but just show a blank page. I'll have to find something else.

jberdine:
Due to failure to remove the landing page, I updated to a hopefully-temporary state that isn't as bad as the current one. With this PR the landing page looks like:
image
The links on the docs still point to this one though, and there is no link back to the main site from anywhere in the docs.

PR #586, opened 2024-12-10
Constant/Param/NonEmptyIterator type renamings

### LINKED ISSUE #578: Consider renaming the Constant type, and relationship with Param

bennostein:
See #578 -- this PR aims to make the names of our types and symbols for frozen/skip-managed values a bit more accurate and descriptive. Doc comments are updated in the PR, but high-level is:

 * `Constant` becomes `DepSafeObj`: these objects are _either_ deep-frozen or managed by the skip runtime (in which case they're not constant) but the key feature is that they can safely be used within reactive computations without incurring untracked dependencies
 * `Param` becomes `DepSafeValue`: these values are similarly dependency-safe; this type is relevant in other contexts than mapper/reducer params, and `Param` doesn't tell you much about what the type actually means in any case.
 * `NonEmptyIterator` becomes `Values` -- see https://github.com/SkipLabs/skip/pull/582#discussion_r1876054925 for some context.  This is the name I'm least sure about; `Values` makes clear that it's not meant as a general-use data structure but instead for use in our API, but isn't super descriptive otherwise.  I didn't want to use `DepSafeValues` as the "dependency-safe-ness" is not necessarily the primary characteristic -- they're just the values you get in a mapper! (which so happen to be dependency safe)

Of course, these are just what I was able to come up with -- if people are opposed or have better options, happy to change things around!

### DIFF
--- a/examples/hackernews/reactive_service/src/hackernews.service.ts
+++ b/examples/hackernews/reactive_service/src/hackernews.service.ts
@@ -2,7 +2,7 @@ import type {
   EagerCollection,
   Entry,
   Json,
-  NonEmptyIterator,
+  Values,
   Resource,
   SkipService,
 } from "@skipruntime/runtime";
@@ -56,10 +56,7 @@ export function serviceWithInitialData(
 }
 
 class UpvotesMapper {
-  mapEntry(
-    key: number,
-    values: NonEmptyIterator<Upvote>,
-  ): Iterable<[number, number]> {
+  mapEntry(key: number, values: Values<Upvote>): Iterable<[number, number]> {
     const value = values.getUnique().post_id;
     return [[value, key]];
   }
@@ -71,10 +68,7 @@ class PostsMapper {
     private upvotes: EagerCollection<number, number>,
   ) {}
 
-  mapEntry(
-    key: number,
-    values: NonEmptyIterator<Post>,
-  ): Iterable<[number, Upvoted]> {
+  mapEntry(key: number, values: Values<Post>): Iterable<[number, Upvoted]> {
     const post: Post = values.getUnique();
     const upvotes = this.upvotes.getArray(key).length;
     const author = this.users.getUnique(post.author_id);
@@ -84,10 +78,7 @@ class PostsMapper {
 }
 
 class SortingMapper {
-  mapEntry(
-    key: number,
-    values: NonEmptyIterator<Upvoted>,
-  ): Iterable<[number, Upvoted]> {
+  mapEntry(key: number, values: Values<Upvoted>): Iterable<[number, Upvoted]> {
     const posts = values.toArray();
     // Sorting in descending order of upvotes.
     posts.sort((a, b) => b.upvotes - a.upvotes);
--- a/skiplang/skjson/ts/binding/src/index.ts
+++ b/skiplang/skjson/ts/binding/src/index.ts
@@ -5,21 +5,21 @@ export type { Pointer, Nullable, Binding };
 export type { Type };
 
 export const sk_isObjectProxy: unique symbol = Symbol();
-export const sk_frozen: unique symbol = Symbol.for("Skip.frozen");
+export const sk_managed: unique symbol = Symbol.for("Skip.managed");
 
 /**
- * Values that are either unmodifiable or managed by the Skip Runtime.
+ * Values that are either unmodifiable or tracked by the Skip Runtime.
  *
- * A `Constant` value is either managed by the Skip Runtime, in which case its modifications are carefully tracked by the reactive computation system, or it is deep-frozen, meaning that it cannot be modified and neither can its sub-objects, recursively.
- * See `deepFreeze` to make an object `Constant`.
+ * A `Managed` value is either managed by the Skip Runtime, in which case its modifications are carefully tracked by the reactive computation system, or it is deep-frozen, meaning that it cannot be modified and neither can its sub-objects, recursively.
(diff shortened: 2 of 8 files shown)


jberdine:
I have been thinking about this since opening the issue. One thing I haven't had a chance to check but wonder is whether there would be any bad effect of (in terms of the old names) replacing all occurrences of `Constant` in the exposed API with `Param` except where it is actually describing deep-frozen values that are indeed constant. I think the main change would be to make the collection types and context subtypes of `Param` instead of `Constant`. I don't know of anything that would be unhappy with that. Then we could leave `Constant` named as it is, as at least from the exposed API, it would only apply to deep-frozen values. Then maybe `Managed` would be an ok name instead of `Param`. If we consider deep-freezing an object as a form of management by the skip runtime, then only for the primitive values is "managed" a bit of a lie, but that seems ok to me. The `DepSafe` terminology seems ok to me, but something of a mouthful and I think not usefully more informative than just "managed". WDYT?

bennostein:
> One thing I haven't had a chance to check but wonder is whether there would be any bad effect of (in terms of the old names) replacing all occurrences of `Constant` in the exposed API with `Param` except where it is actually describing deep-frozen values that are indeed constant. I think the main change would be to make the collection types and context subtypes of `Param` instead of `Constant`. I don't know of anything that would be unhappy with that.

I think that should work -- possibly some worse type inferencing results and error messages? But that's all I can think of.

> Then we could leave `Constant` named as it is, as at least from the exposed API, it would only apply to deep-frozen values. Then maybe `Managed` would be an ok name instead of `Param`. If we consider deep-freezing an object as a form of management by the skip runtime, then only for the primitive values is "managed" a bit of a lie, but that seems ok to me. The `DepSafe` terminology seems ok to me, but something of a mouthful and I think not usefully more informative than just "managed". WDYT?

I maybe see that "bit of a lie" as being more problematic, and find deep-freezing as a form of "management" unintuitive -- that implies to me some active management, ceding of control to the runtime, etc.  Whereas "dependency-safe" (though I agree it doesn't roll off the tongue) describes a constraint being satisfied: give us your object/value/etc, just make sure that it doesn't have out-of-band dependencies.

To be clear, I think either choice would be reasonable and an improvement over status quo -- if you feel strongly (or others weigh in) its easy enough to use that Constant/Managed naming scheme

jberdine:
Not pressing, just to explain my thinking.

One thing that I have in mind that I now realize isn't clear is that I'm thinking of the version of deepFreeze that returns a copy of, rather than mutating the argument. It would need to be renamed. (This is a change that @pikatchu @mbouaziz and I have discussed, and Mehdi had made the change at some point, but it hasn't landed.) At that point the frozen aspect is an implementation detail, and we could instead change to an implementation that e.g. makes a copy on the skip heap and returns some "pointer" to it. And even currently, there is no (legit) way for user code to get a deep-frozen value other than calling our deepFreeze function. This is why I think of deep-freezing as a form of management, it just happens to be done on the js heap currently rather than on the skip heap.

On the primitives, I guess my thinking is that while for non-primitive types, we require calling deepFreeze to inject them into the type that would be the SKStore monad if it was presented that way, that is deepFreeze is the `return`/`pure`, and we just happen to not require user code to call any such injection on primitives since we are not making a type-level distinction between values of type T and "reactive" values of type T, the input and output types of what would be the monadic return. From that perspective, using values of primitive type directly as dependencies is a "cheat", which if we didn't do, would make their image under return legitimately "managed".

jberdine:
Thinking more, I have another proposal that I wonder what you think of. What if we renamed `Constant` to `Managed` and `Param` to `DepSafe`? The values tagged with `sk_frozen` are managed, and using "managed" abstracts over the implementation technique to achieve it. Being able to drop `Obj` and `Value` suffixes from `DepSafe` makes it more agreeable I think.

Separately, I was thinking of words for "dependency safe" and "reliable" and "dependable" came to mind. Do either of those strike you as good names for the type?

jberdine:
And re `Values`, it does squat a very general bit of the namespace, but it does read well in the api and example code in particular.

bennostein:
> Thinking more, I have another proposal that I wonder what you think of. What if we renamed `Constant` to `Managed` and `Param` to `DepSafe`? The values tagged with `sk_frozen` are managed, and using "managed" abstracts over the implementation technique to achieve it. Being able to drop `Obj` and `Value` suffixes from `DepSafe` makes it more agreeable I think.

I like that! Making the change now.

> Separately, I was thinking of words for "dependency safe" and "reliable" and "dependable" came to mind. Do either of those strike you as good names for the type?

I think using those in prose/marketing copy etc. where we have room to give description and context sounds good, but I'm not sure about using them for the types themselves, since they are a bit more open to interpretation imo

jberdine:
> > Separately, I was thinking of words for "dependency safe" and "reliable" and "dependable" came to mind. Do either of those strike you as good names for the type?
> 
> I think using those in prose/marketing copy etc. where we have room to give description and context sounds good, but I'm not sure about using them for the types themselves, since they are a bit more open to interpretation imo

Yes, prose can add context to clarify. But I have to note that I think `value: Dependable` is very close to "some value that can be a dependency".

jberdine (approved):

PR #587, opened 2024-12-10
Use POST payload instead of GET params for synchronous reads

### LINKED ISSUE #517: Resource keys interpreted as strings in URLs
The default REST server implementation passes path-params straight through to Skip, which causes some confusion when resources have non-string keys.

https://github.com/SkipLabs/skip/blob/146c9ff13004f468f054fb60464fbd6d6c932fd9/skipruntime-ts/server/src/rest.ts#L19

One option is to have some logic there to parse out a JS value of the correct type from the path parameter.  But, that may end up messy or inaccurate, since these really are just HTTP-layer strings.

So, I'd propose we just enforce that resource keys are strings, changing the return type of `instantiate` from `EagerCollection<Json, Json>` to `EagerCollection<string, Json>`, unless anyone has a more elegant solution (or there's something else I'm missing)

### LINKED ISSUE #535: Consider using POST instead of GET for resource reads

bennostein:
Resolves #535, resolves #517 

This has a few advantages:
 * Allows resources to have arbitrary key types without any string encoding weirdness resulting from shoving them into path parameters (obviating the need for #539)
 * Unifies the handling/specifying of resources/params between different API routes
 * New API makes very clear that the result is a _snapshot_ in time not a reactive value, and that a resource is being instantiated to respond to the synchronous read query
 
 I've tested against a few of the examples under `./skipruntime-ts/examples` and all seems good.  But, I can't check on `./examples/hackernews` until we publish @skipruntime/runtime to NPM.

### DIFF
--- a/skipruntime-ts/server/src/rest.ts
+++ b/skipruntime-ts/server/src/rest.ts
@@ -38,37 +38,23 @@ export function controlService(service: ServiceInstance): express.Express {
   });
 
   // READS
-  app.get("/v1/resources/:resource/:key", (req, res) => {
+  app.post("/v1/snapshot", (req, res) => {
     try {
-      service.getArray(
-        req.params.resource,
-        req.params.key,
-        req.query as { [param: string]: string },
-        {
-          resolve: (data: Json[]) => {
-            res.status(200).json(data);
-          },
-          reject: (err: unknown) => {
-            res.status(500).json(err instanceof Error ? err.message : err);
-          },
-        },
-      );
-    } catch (e: unknown) {
-      console.log(e);
-      res.status(500).json(e instanceof Error ? e.message : e);
-    }
-  });
-
-  app.get("/v1/resources/:resource", (req, res) => {
-    try {
-      service.getAll(req.params.resource, req.query as Record<string, string>, {
-        resolve: (data: Entry<Json, Json>[]) => {
+      const resource = req.body.resource as string;
+      const params = req.body.params as { [param: string]: string };
+      const callbacks = {
+        resolve: (data: Json[]) => {
           res.status(200).json(data);
         },
         reject: (err: unknown) => {
           res.status(500).json(err instanceof Error ? err.message : err);
         },
-      });
+      };
+      if (req.body.key) {
+        service.getArray(resource, req.body.key, params, callbacks);
+      } else {
+        service.getAll(resource, params, callbacks);
+      }
     } catch (e: unknown) {
       console.log(e);
       res.status(500).json(e instanceof Error ? e.message : e);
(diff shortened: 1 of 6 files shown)


jberdine (approved):
Nice!

skipruntime-ts/server/src/server.ts
```diff
@@ -25,15 +25,12 @@ export type SkipServer = {
...
+ * - `POST /v1/snapshot`:
+ *   Synchronous read of a resource.
  *
- *   Instantiates the named `resource` with parameters `{param1=value1,...,paramN=valueN}` and responds with the entire contents of the resource. The returned data will be a JSON-encoded value of type `[Json, Json[]][]`: an array of entries each of which associates a key to an array of its values.
- *
- * - `GET /v1/resources/:resource/:key?param1=value1&...&paramN=valueN`:
- *   Synchronous read of a single key from a resource.
- *
- *   Instantiates the named `resource` with parameters `{param1=value1,...,paramN=valueN}` and responds with the values associated with the given `key`. The returned data will be a JSON-encoded value of type `Json[]`: an array of the `key`'s values.
+ *   Requires a JSON-encoded request body of the form `{ resource: string; params: { [param: string]: string }; key?: Json }`.
```
beauby, line 31: Should we replace `key?: Json` with `keys?: Json[]`? I don't know whether the "fetching multiple (but not all) keys" is important in practice.

skipruntime-ts/server/src/rest.ts
```diff
@@ -38,37 +38,23 @@ export function controlService(service: ServiceInstance): express.Express {
...
+        resolve: (data: Json[]) => {
           res.status(200).json(data);
         },
         reject: (err: unknown) => {
           res.status(500).json(err instanceof Error ? err.message : err);
         },
-      });
+      };
+      if (req.body.key) {
+        service.getArray(resource, req.body.key as string, params, callbacks);
```
beauby: Is `key` always a string? The API docs state that it can be arbitrary JSON.

bennostein: Nice catch! No, this was just a mistake/partial refactor that I left in, should be fine as arbitrary JSON

skipruntime-ts/server/src/server.ts
bennostein: I could see it being potentially useful and happy to implement that if there's demand.  I didn't do it in this initial PR as this API seemed simpler and my intuition was that these two cases (get everything, get this one particular key) would be the most common

PR #590, opened 2024-12-10
Allow arbitrary `Json` values as resource parameters

### LINKED ISSUE #565: Generalize parameter encoding in Polled ExternalServices
Currently Polled external services pass parameters to the external service using [`URLSearchParams.toString`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString). This is a common format, but may not be suitable for all external services. We should make the parameter encoding parametric.

bennostein:
Not directly enabled by but related to #587 , this PR loosens the constraints we place on resource parameters, from `{ [param: string]: string }` to `{ [param: string]: Json }`, allowing for more richly structured data there. The typescript-side changes are pretty minimal, but I also changed the Skip-lang representation of parameters and cleaned up some things there.

~~This heightens the salience of #565 -- I've kept the logic there as is, but this means that communicating with non-Skip external services may entail JSON->string->JSON encoding/decoding and all the potential fun that comes with that~~ EDIT: actually just decided to do this, resolves #565

### DIFF
--- a/skipruntime-ts/helpers/src/external.ts
+++ b/skipruntime-ts/helpers/src/external.ts
@@ -95,6 +95,15 @@ export class TimerResource implements ExternalResource {
   }
 }
 
+function defaultParamEncoder(params: { [param: string]: Json }): string {
+  const queryParams: { [param: string]: string } = {};
+  for (const [key, value] of Object.entries(params)) {
+    if (typeof value == "object") queryParams[key] = JSON.stringify(value);
+    else queryParams[key] = value.toString();
+  }
+  return new URLSearchParams(queryParams).toString();
+}
+
 export class Polled<S extends Json, K extends Json, V extends Json>
   implements ExternalResource
 {
@@ -104,6 +113,9 @@ export class Polled<S extends Json, K extends Json, V extends Json>
     private readonly url: string,
     private readonly duration: number,
     private readonly conv: (data: S) => Entry<K, V>[],
+    private readonly encodeParams: (params: {
+      [param: string]: Json;
+    }) => string = defaultParamEncoder,
   ) {}
 
   open(
@@ -115,13 +127,7 @@ export class Polled<S extends Json, K extends Json, V extends Json>
     },
   ): void {
     this.close(params);
-    const queryParams: { [param: string]: string } = {};
-    for (const [key, value] of Object.entries(params)) {
-      if (typeof value == "object") queryParams[key] = JSON.stringify(value);
-      else queryParams[key] = value.toString();
-    }
-    const strParams = new URLSearchParams(queryParams).toString();
-    const url = `${this.url}?${strParams}`;
+    const url = `${this.url}?${this.encodeParams(params)}`;
     const call = () => {
       callbacks.loading();
       fetchJSON(url, "GET", {})
--- a/skipruntime-ts/helpers/src/external.ts
+++ b/skipruntime-ts/helpers/src/external.ts
@@ -3,7 +3,7 @@ import { fetchJSON } from "./rest.js";
 
 export interface ExternalResource {
   open(
-    params: { [param: string]: string | number },
+    params: { [param: string]: Json },
     callbacks: {
       update: (updates: Entry<Json, Json>[], isInit: boolean) => void;
       error: (error: Json) => void;
@@ -11,7 +11,7 @@ export interface ExternalResource {
     },
   ): void;
 
-  close(params: { [param: string]: string | number }): void;
+  close(params: { [param: string]: Json }): void;
 }
 
 export class GenericExternalService implements ExternalService {
@@ -21,7 +21,7 @@ export class GenericExternalService implements ExternalService {
 
   subscribe(
     resourceName: string,
-    params: { [param: string]: string | number },
+    params: { [param: string]: Json },
     callbacks: {
       update: (updates: Entry<Json, Json>[], isInit: boolean) => void;
       error: (error: Json) => void;
@@ -37,7 +37,7 @@ export class GenericExternalService implements ExternalService {
     resource.open(params, callbacks);
   }
 
-  unsubscribe(resourceName: string, params: { [param: string]: string }) {
+  unsubscribe(resourceName: string, params: { [param: string]: Json }) {
     const resource = this.resources[resourceName] as
       | ExternalResource
       | undefined;
@@ -58,7 +58,7 @@ export class TimerResource implements ExternalResource {
   private readonly intervals = new Map<string, { [name: string]: Timeout }>();
 
   open(
-    params: { [param: string]: string | number },
+    params: { [param: string]: Json },
     callbacks: {
       update: (updates: Entry<Json, Json>[], isInit: boolean) => void;
       error: (error: Json) => void;
@@ -85,7 +85,7 @@ export class TimerResource implements ExternalResource {
     this.intervals.set(id, intervals);
   }
 
-  close(params: { [param: string]: string | number }): void {
+  close(params: { [param: string]: Json }): void {
     const intervals = this.intervals.get(toId(params));
     if (intervals != null) {
       for (const interval of Object.values(intervals)) {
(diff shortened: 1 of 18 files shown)


www/docs/resources.md
```diff
@@ -81,7 +81,7 @@ POST /v1/streams
 DELETE /v1/streams/:uuid
 ```
 
-The `POST` route instantiates a resource according to the JSON-encoded request body (consisting of the resource identifier and any parameters, structured as `{ resource: string; params: { [param: string]: string } }`) and returns a UUID identifying the resource, which can then be used in a query to the streaming API.
+The `POST` route instantiates a resource according to the JSON-encoded request body (consisting of the resource identifier and any parameters, structured as `{ resource: string; params: { [param: string]: any } }`) and returns a UUID identifying the resource, which can then be used in a query to the streaming API.
```
jberdine, line 84: Why `any` rather than `Json`?

www/docs/resources.md
```diff
@@ -90,7 +90,7 @@ Synchronous reads from reactive resources can either access the resource in its
 POST /v1/snapshot
 ```
 
-This `POST` route requires a JSON-encoded request body of the form `{ resource: string; params: { [param: string]: string }; key?: any }`.
+This `POST` route requires a JSON-encoded request body of the form `{ resource: string; params: { [param: string]: any }; key?: any }`.
```
jberdine, line 93: Why `any` rather than `Json`? And now that I see it, also for `key`.

skipruntime-ts/helpers/src/external.ts
```diff
@@ -117,7 +117,8 @@ export class Polled<S extends Json, K extends Json, V extends Json>
     this.close(params);
     const queryParams: { [param: string]: string } = {};
     for (const [key, value] of Object.entries(params)) {
-      queryParams[key] = value.toString();
+      if (typeof value == "object") queryParams[key] = JSON.stringify(value);
+      else queryParams[key] = value.toString();
```
jberdine: I worry that this sort of conditional application of JSON encoding will be hard to detect and invert on the other end. Am I confused?

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -22,14 +22,15 @@ class ExistingSubscriptionException() extends Exception {
   }
 }
 
-class Params(value: Map<String, PValue>) extends SKStore.File uses Orderable {
+class Params(
+  sortedKeys: Array<String>,
+  sortedVals: Array<SKJSON.CJSON>,
```
jberdine: I'm curious about the change from a map to a pair of arrays. Would `Map<String, SKJSON.CJSON>` be an option? In either case, it seems that it would be better to construct `Params` by taking the key-value pairs in some form where the keys and values are directly correlated with each other, and sort and split in the `Params` constructor, close in the code to where the sorted invariants are relied on. It seems a big ask of the users of this type to sort the keys and make sure the values are in the corresponding order.

skipruntime-ts/helpers/src/external.ts
```diff
@@ -104,6 +113,9 @@ export class Polled<S extends Json, K extends Json, V extends Json>
     private readonly url: string,
     private readonly duration: number,
     private readonly conv: (data: S) => Entry<K, V>[],
+    private readonly encodeParams: (params: {
+      [param: string]: Json;
+    }) => string = defaultParamEncoder,
```
jberdine, line 118: This generalization is specific to `Polled` rather than appearing in `ExternalResource` because with `Polled` the user doesn't get to define `open`, while in the generic case they can implement whatever param encoding they need, right?

www/docs/resources.md
bennostein: My thinking was that this is talking about string-encoded JSON payloads on HTTP requests, not JavaScript values/types -- so "any" felt more general for e.g. web service clients not using JavaScript.

skipruntime-ts/helpers/src/external.ts
bennostein: No, not confused, the only value of doing it this way is to keep string query parameters encoded idiomatically, i.e. without quotes as `example.com?foo=bar&baz=quux`; if we just `JSON.stringify`'ed everything, that'd end up `example.com?foo="bar"&baz="quux"`.  On the other hand, if we don't `JSON.stringify` objects then they'll all just end up rendered as `[object Object]`.

I think this should be fine, especially now that the param encoding is configurable -- if people want different behavior, they can pass in a different encoding.

skipruntime-ts/native/src/Runtime.sk
bennostein: I made the change to avoid creating those arrays and sorting them on every single comparison, just doing it once upfront.  But I can bring the constructor in here as a static "create" function like I did with the toString -- would that help?

skipruntime-ts/helpers/src/external.ts
bennostein: Yes, I tried to push this down in the class hierarchy so that it doesn't constrain anything for the generic case -- in which you just have to know how to "open" and "close" with the given Json params.

I suppose there's an argument to lift it up to the ExternalService though, to avoid repetition if you have multiple ExternalResources on the same service, all using the same param encoding...

www/docs/resources.md
jberdine: The implementation will treat the body assuming the type with `Json` instead of `any`, and in particular pass the params to `skjson.exportJSON`, so the body will have to be JSON independently of the language users write in. But maybe just saying that the body is JSON-encoded is enough to imply that the values of params must also be JSON. I don't know, but what is written here is a typescript type, so it seems that it would be clearest to give the correct one here (and let users of other languages infer the gap themselves) or else document the required form of the body using just JSON (schemas?) and no typescript types.

skipruntime-ts/native/src/Runtime.sk
jberdine: I think that would help, yes.

Also note that if you're concerned about creating arrays, it would be better to avoid the `zip` on every compare as it will also create the array of pairs before immediately iterating it, and instead use a loop.

skipruntime-ts/helpers/src/external.ts
jberdine: I don't have an intuition about repetition over multiple ExternalResources, but guess that surely people can factor out their implementations to dedup that if needed. So I'd leave it as you have it for now.

www/docs/resources.md
bennostein: We could (and probably should) just define some [proper schema](json-schema.org)

In this case, though, I don't think there's any meaningful difference between `Json` and `any` -- both allow for any text-encodable JSON value. So the question is which one serves better as documentation, which for me would be `any` (expecting more people to be familiar with it than our `Json` as a \top type)

jberdine: Ok. To me `any` includes things like closures and objects with methods, so I find it confusing here, but maybe just me.

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -22,14 +22,22 @@ class ExistingSubscriptionException() extends Exception {
   }
 }
 
-class Params(value: Map<String, PValue>) extends SKStore.File uses Orderable {
+class Params(
```
jberdine: To disable unsafe construction:
```suggestion
class Params private (
```

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -22,14 +22,22 @@ class ExistingSubscriptionException() extends Exception {
...
+  }
   fun compare(other: Params): Order {
-    keys = this.value.keys().collect(Array).sorted();
-    okeys = other.value.keys().collect(Array).sorted();
-    kcomp = keys.compare(okeys);
+    kcomp = this.sortedKeys.compare(other.sortedKeys);
     if (kcomp == EQ()) {
-      for (key in keys) {
-        vcomp = this.value.get(key).compare(other.value.get(key));
+      for (i in Range(0, this.sortedVals.size())) {
```
jberdine, line 39: Should the sizes of the two vals arrays first be checked to be equal?

jberdine (approved):

skipruntime-ts/native/src/Runtime.sk
jberdine: I pulled this branch and checked that adding `private` here does seem to work. Did you not like that idea, or this comment got missed, or?

bennostein: I made the change but ended up switching to a different representation in #603 which didn't need the private modifier.  But I'll add it here and then land so that reviewing that is easier, per your comment over there!

PR #592, opened 2024-12-11
Add note about synchronous writes and non-JS backends in client docs

### LINKED ISSUE #566: [docs] Extend client example with writes
I think it would make sense to expand this example and explanation to cover sync writes to a collection.

_Originally posted by @jberdine in https://github.com/SkipLabs/skip/pull/558#discussion_r1863340255_

bennostein:
Some minor changes in the docs text, noting that (a) synchronous writes are also possible more-or-less exactly as described for synchronous reads and (b) the JS SkipServiceBroker is a thin wrapper around an HTTP API that other non-JS backends can access directly


jberdine (approved):
LGTM. Only comments I have are on the commit message. :-)

- it says do *writes* just like *writes*, second should be reads
- calling this REST bothers me, since if REST means anything more precise than HTTP, it must indicate that an API is not just RPC, which is what ours is AFAIU

PR #597, opened 2024-12-13
Change header font to Chicago, self-host fonts

jberdine:


www/src/css/custom.css
```diff
@@ -4,7 +4,23 @@
  * work well for content-centric websites.
  */
 
-@import url("https://fonts.googleapis.com/css2?family=Quicksand:wght@300..700&family=Space+Mono&display=swap");
```
jberdine, line 7: This seems to work fine, but lets google track all traffic to our docs site.

jberdine:
image

www/src/css/custom.css
pikatchu: I think that's ok

pikatchu:
Ship it!

pikatchu (approved):
Awesome thanks!

PR #602, opened 2024-12-13
Dedup definition of Entrypoint

jberdine:


skipruntime-ts/helpers/package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "eventsource": "^2.0.2",
     "express": "^4.21.1",
+    "@skipruntime/core": "^0.0.1",
```
jberdine: Is it ok to have the helpers package depend on core?

skiplabsdaniel (approved):

skipruntime-ts/helpers/package.json
beauby: I think it is acceptable. We just want to avoid the dependency in the opposite sense.

beauby (approved):

PR #603, opened 2024-12-13
Resources as path parameters

bennostein:
This PR follows up on #590 based on our conversation at yesterday's meeting.
(It's also stacked on that PR for hopefully easier reviewing)

I implemented more or less exactly the API we discussed, with one change:  I added a single-key-lookup synchronous read method to the API, as it was used in some of the examples and feels much more natural this way than forcing creation of another resource, then calling `getAll` on that resource instance (which, using its params, filters out the desired key), then stripping away the redundant structure in the response (which would always have the form `[[key, [val1, val2,...]]]` instead of just `[val1, val2, ...]`, etc.

So, I propose the API in this PR, with two synchronous read routes as documented in `skipruntime-ts/server/src/server.ts` and `www/docs/resources.md`, copied here:

Synchronous reads from reactive resources can either access the resource in its entirety or read the data for a single key, using routes:

```
POST /v1/snapshot/:resource
POST /v1/snapshot/:resource/lookup
```

The first route returns all of the values in the named `resource`, using the parameters provided in the JSON-encoded request body.
It instantiates the resource if needed, then returns a JSON-encoded array of key/value entries, with each entry a tuple of the form `[key, [value1, value2, ...]]`.

The second route requires the request body to be a JSON-encoded value with a `key` field and a `params` field.
It instantiates the resource if needed, then returns a JSON-encoded array of all values associated to `key` in the resource.

### DIFF
--- a/skipruntime-ts/examples/groups.ts
+++ b/skipruntime-ts/examples/groups.ts
@@ -8,22 +8,22 @@ import {
 
 import { runService } from "@skipruntime/server";
 
-type UserID = string;
-type GroupID = string;
+type UserID = number;
+type GroupID = number;
 type User = { name: string; active?: boolean; friends: UserID[] };
 type Group = { name: string; members: UserID[] };
 
 // Load initial data from a source-of-truth database (mocked for simplicity)
 const initialData: InitialData<ServiceInputs> = {
   users: [
-    ["bob", [{ name: "Bob", active: true, friends: ["alice", "carol"] }]],
-    ["alice", [{ name: "Alice", active: true, friends: ["bob", "carol"] }]],
-    ["carol", [{ name: "Carol", active: false, friends: ["bob", "alice"] }]],
-    ["eve", [{ name: "Eve", active: true, friends: [] }]],
+    [0, [{ name: "Bob", active: true, friends: [1, 2] }]],
+    [1, [{ name: "Alice", active: true, friends: [0, 2] }]],
+    [2, [{ name: "Carol", active: false, friends: [0, 1] }]],
+    [3, [{ name: "Eve", active: true, friends: [] }]],
   ],
   groups: [
-    ["group1", [{ name: "Group 1", members: ["alice", "carol", "eve"] }]],
-    ["group2", [{ name: "Group 2", members: ["bob", "carol"] }]],
+    [1001, [{ name: "Group 1", members: [1, 2, 3] }]],
+    [1002, [{ name: "Group 2", members: [0, 2] }]],
   ],
 };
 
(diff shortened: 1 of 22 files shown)


jberdine:
It would be nice for reviewing to push #590 to origin and then set the base of this PR to that branch.

skipruntime-ts/examples/groups.ts
```diff
@@ -8,22 +8,22 @@ import {
 
 import { runService } from "@skipruntime/server";
 
-type UserID = string;
-type GroupID = string;
+type UserID = number;
+type GroupID = number;
```
jberdine, line 12: I'm curious if this change was forced by the rest of this PR.

skipruntime-ts/server/src/server.ts
```diff
@@ -39,12 +46,19 @@ export type SkipServer = {
  *   The body of the request must be a JSON-encoded value of type `CollectionUpdate.values`, that is `[Json, Json[]][]`: an array of entries each of which associates a key to an array of its new values.
  *   Updates the named `collection` with the key-values entries in the request body.
  *
- * - `POST /v1/streams`:
+ * - `PUT /v1/inputs/:collection/:key`:
+ *   Update of a single key of an input collection.
+ *
+ *   The `collection` must be the name of one of the service's input collections, that is, one of the keys of the `Inputs` type parameter.
+ *   The body of the request must be a JSON-encoded value of type `Json[]`: an array of the `key`'s new values.
```
jberdine: You changed the other cases to not use types to describe the JSON, did you mean to do this one too?

www/docs/resources.md
```diff
@@ -77,22 +77,25 @@ The control API (on port 8081 by default) surfaces resource instantiation/deleti
...
 ```
-POST /v1/snapshot
+POST /v1/snapshot/:resource
+POST /v1/snapshot/:resource/lookup
 ```
 
-This `POST` route requires a JSON-encoded request body of the form `{ resource: string; params: { [param: string]: any }; key?: any }`.
-It instantiates a resource if needed to according to the `resource` and `params` fields of the request body, then either returns the values indicated by the `key` or _all_ keys/values if no `key` is provided.
-For reads of a specific `key`, data is returned as an array of values associated to that key; for reads of an entire resource, data is returned as an array of key/value entries, with each entry a tuple of the form `[key, [value1, value2, ...]]`.
+The first route returns all of the values in the named `resource`, using the parameters provided in the JSON-encoded request body.
```
jberdine:
```suggestion
The first route returns all of the entries in the named `resource`, using the parameters provided in the JSON-encoded request body.
```

jberdine (approved):
LGTM

skipruntime-ts/examples/groups.ts
bennostein: Not forced, but enabled!  I originally had these as numbers when putting this example together, but they were being interpreted as strings in URLs, so I just used strings and opened some issues to support numbers/other types in resource keys later on.

PR #608, opened 2024-12-17
Return Iterable from LazyCompute

### LINKED ISSUE #607: Audit/confirm type of LazyCompute.compute
The type of `LazyCompute.compute` is declared as `compute(self: LazyCollection<K, V>, key: K): Nullable<V>` in api.ts. It is not clear why the return type is `Nullable<V>`, I expected `Array<V>`. Perhaps I have gotten lost in the cross-lang interop, but it seems that `createLazyCollection` calls `createLazyCompute` which returns an `ExternLazyCompute` whose `compute` is defined as
```
  fun compute(self: LazyCollection, key: SKJSON.CJSON): Array<SKJSON.CJSON> {
    computeOfLazyCompute(this.eptr.value, self.getId(), key) match {
    | SKJSON.CJArray(value) -> value
    | _ -> invariant_violation("Invalid computed value.")
    }
  }
```
This seems more in line with returning an array than a nullable.

bennostein:
I went to see what it would take to do this re: #607 and turns out... very little!  no Skiplang changes needed, since the nullable was already being converted to a JS array (at skipruntime-ts/core/src/index.ts#792)

closes #607

### DIFF
--- a/skipruntime-ts/api/src/api.ts
+++ b/skipruntime-ts/api/src/api.ts
@@ -263,7 +263,7 @@ export interface EagerCollection<K extends Json, V extends Json>
  * The type of a _lazy_ reactive function which produces a value for some `key`, possibly using a `self` reference to get/produce other lazily-computed results.
  */
 export interface LazyCompute<K extends Json, V extends Json> {
-  compute(self: LazyCollection<K, V>, key: K): Nullable<V>;
+  compute(self: LazyCollection<K, V>, key: K): Iterable<V>;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


jberdine:
It seems good from the API symmetry perspective to make this change, but I do have to wonder if returning a JS Iterable somehow lets in some laziness that might interact poorly. I don't know.

But mainly I am not confident enough to approve this. In particular, my reading of the code in native/src/Extern.sk is that the skiplang code expects the called JS function to return an array. And my confidence that the tests we have, that construct only arrays as iterables, are enough to flush out incompatibilities with other Iterables is not high.

So I'd say let's wait for @skiplabsdaniel's opinion.

skiplabsdaniel (approved):

bennostein:
> In particular, my reading of the code in native/src/Extern.sk is that the skiplang code expects the called JS function to return an array. And my confidence that the tests we have, that construct only arrays as iterables, are enough to flush out incompatibilities with other Iterables is not high.

This part I'm quite confident about -- the array passed from JS to Skip is the one constructed at `skipruntime-ts/core/src/index.ts#792`, and `Array.from` is a robust way to produce an array from any iterable.

PR #617, opened 2024-12-20
Native implementations of reducers and typing improvements throughout `skipruntime-ts/native`

### LINKED ISSUE #488: Predefined/optimized mappers and reducers written in Skiplang for common operations

bennostein:
This moves all of our utility reducers (sum, min, max, count) from TS to Skiplang and closes #488, but also makes some wider cleanups around skipruntime-ts/native.

Previously, that code made various assumptions that all of the collections it deals with are JSONID/JSONFile but typed all of its collections internally as SKStore.Key/SKStore.File (i.e. top).  Now, the proper key/value types are reflected in the type parameters throughout, and I've pulled out a lot of unused abstraction, where structures carried around a lot of "converters" which were always instantiated for JSONID/JSONFile.

This lets us use utility reducers defined in `skiplang/prelude` which should be much more efficient than TS implementations and incur much lower FFI overhead.

I've still got to dig into the native node addon to get this plumbed through there, but pushing this and calling it a day for now, will figure that out later.

### DIFF
--- a/skiplang/prelude/src/skstore/Handle.sk
+++ b/skiplang/prelude/src/skstore/Handle.sk
@@ -258,6 +258,60 @@ fun sumReducer(): EReducer<IntFile, IntFile> {
   };
 }
 
+fun minReducer(): EReducer<IntFile, IntFile> {
+  EReducer{
+    type => IntFile::type,
+    canReset => true,
+    init => iter ~> {
+      min: ?Int = None();
+      for (x in iter) {
+        if (min.isNone() || x.value < min.fromSome()) !min = Some(x.value);
+      };
+      min match {
+      | Some(x) -> Array[IntFile(x)]
+      | None() -> Array[]
+      }
+    },
+    update => (state, old, new) ~> {
+      min = state[0].value;
+      for (x in new) {
+        if (x.value < min) !min = x.value;
+      };
+      for (x in old) {
+        if (x.value <= min) return None()
+      };
+      Some(Array[IntFile(min)])
+    },
+  }
+}
+
+fun maxReducer(): EReducer<IntFile, IntFile> {
+  EReducer{
+    type => IntFile::type,
+    canReset => true,
+    init => iter ~> {
+      max: ?Int = None();
+      for (x in iter) {
+        if (max.isNone() || x.value > max.fromSome()) !max = Some(x.value);
+      };
+      max match {
+      | Some(x) -> Array[IntFile(x)]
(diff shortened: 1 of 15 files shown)


skipruntime-ts/core/src/utils.ts
```diff
@@ -1,88 +1,63 @@
...
-
-/**
- * `Mapper` to count input values.
- *
- * A `Mapper` that associates each key in the output with the number of values it is associated with in the input.
- *
- * @remarks
- * If there are many values associated to keys in a collection, and they are updated frequently, using the `Count` `Reducer` instead could be more efficient.
- */
-export class CountMapper<
```
jberdine, line 81: Note that this is used in www/docs/functions.md.

skiplang/prelude/src/skstore/Handle.sk
```diff
@@ -258,6 +258,60 @@ fun sumReducer(): EReducer<IntFile, IntFile> {
...
+      | None() -> Array[]
+      }
+    },
+    update => (state, old, new) ~> {
+      min = state[0].value;
+      for (x in new) {
+        if (x.value < min) !min = x.value;
+      };
+      for (x in old) {
+        if (x.value < min) return None()
```
jberdine: Shouldn't this be `<=`?

skiplang/prelude/src/skstore/Handle.sk
```diff
@@ -258,6 +258,60 @@ fun sumReducer(): EReducer<IntFile, IntFile> {
...
+      | None() -> Array[]
+      }
+    },
+    update => (state, old, new) ~> {
+      max = state[0].value;
+      for (x in new) {
+        if (x.value > max) !max = x.value;
+      };
+      for (x in old) {
+        if (x.value > max) return None()
```
jberdine: Shouldn't this be `>=`?

jberdine (commented):
Looking good to me, but a couple conditions are slightly off I think.

skiplang/prelude/src/skstore/Handle.sk
bennostein: Yeah, nice catch, thanks! Made the change. (and for max too)

skipruntime-ts/core/src/utils.ts
bennostein: I've got a note to use `Count` there instead (and add a unit test exercising that native reducer), top of my list for when I'm properly back to work next week.

PR #619, opened 2024-12-23
Bump NPM package versions

bennostein:
Here's what I've been working with, but still running into issues building and packaging the native addon.

Putting this up as is for now, to see how circleCI handles it.


bennostein:
Landed and closing, these versions are all now live now on NPM (arm64 only for the addon so far, working on x86)

PR #626, opened 2025-01-09
Update 2 dependencies

mbouaziz:
Annoyed by a few irksome deprecation messages when I `npm install`, I decided to update `mocha` and switch the examples from `sqlite3` to `better-sqlite3`.

Example still works (without addon), see #634 and related fix in #635


beauby (approved):
The examples are not in CI (which is a shame). LGTM as long as the examples do not break.

jberdine:
> The examples are not in CI (which is a shame). LGTM as long as the examples do not break.

IIUC they are at least type-checked by the `check-ts` job.

PR #627, opened 2025-01-09
[skiplang/skc] Fix stale package invalidation.

### LINKED ISSUE #595: Undeclared dependency in cc on prelude
I am seeing build failures that appear to be caused by the cc package implementation using the stdlib without declaring a dependency on it, and doing so creates a circular dependency.

- skiplang/cc/src/utils.sk uses System
- System is defined in skiplang/prelude/src/stdlib/other/System.sk
- System needs the skargo dep line: `std = { path = "../prelude" }`
- adding that to skiplang/cc/Skargo.toml creates a circular dependency

beauby:
Bug introduced in cb0f383.
Closes #595.


skiplang/compiler/src/skipParse.sk
```diff
@@ -111,19 +117,19 @@ fun updatePackageFiles(
...
+      InputSource(name, Path.join(pkg_base_dir, src_path)),
       Array[SKStore.StringFile(get_file_contents(src_path))],
     )
   };
-  for (src_path in removed_files) {
+  for ((pkg_base_dir, src_path) in removed_files) {
     fileDir.writeArray(
       context,
-      InputSource(name, Path.join(pkg.base_dir, src_path)),
+      InputSource(name, Path.join(pkg_base_dir, src_path)),
```
jberdine, line 132: So was the core of the issue that modified or removed files can be from a different package, in which case using `pkg.base_dir` yields the wrong file?

jberdine (approved):
Thanks!

skiplang/compiler/src/skipParse.sk
beauby: Yes, in case the whole package (in this case `std` was replaced by an other implementation (in this case the bootstrap one vs the local one)) is replaced, all the files from the old package are removed, and all those of the new one added.

PR #629, opened 2025-01-10
Fix/document dep safety of slice and initialData

jberdine:
- Require slice args to be DepSafe

- Document that initialData is not dependency-safe

### DIFF
--- a/skipruntime-ts/api/src/api.ts
+++ b/skipruntime-ts/api/src/api.ts
@@ -364,7 +364,7 @@ export interface EagerCollection<K extends Json, V extends Json>
    * @param end - The greatest key whose entry will be kept in the result.
    * @returns The restricted eager collection.
    */
-  slice(start: K, end: K): EagerCollection<K, V>;
+  slice(start: K & DepSafe, end: K & DepSafe): EagerCollection<K, V>;
 
   /**
    * Create a new eager collection by keeping only the elements whose keys are in at least one of the given ranges.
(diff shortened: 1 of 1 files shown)


skipruntime-ts/api/src/api.ts
```diff
@@ -364,15 +364,15 @@ export interface EagerCollection<K extends Json, V extends Json>
...
+  slice(start: K & DepSafe, end: K & DepSafe): EagerCollection<K, V>;
 
   /**
    * Create a new eager collection by keeping only the elements whose keys are in at least one of the given ranges.
    *
    * @param ranges - The pairs of lower and upper bounds on keys to keep in the result.
    * @returns The restricted eager collection.
    */
-  slices(...ranges: [K, K][]): EagerCollection<K, V>;
+  slices(...ranges: [K & DepSafe, K & DepSafe][]): EagerCollection<K, V>;
```
jberdine, line 375: AFAIK it isn't possible to pass an array directly to slices, and instead the elements need to be listed individually. Given that, it is enough to ask only for the keys to be DepSafe. If that is wrong, then the whole array should be DepSafe, which will be less convenient for users.

skipruntime-ts/api/src/api.ts
```diff
@@ -646,7 +646,11 @@ export interface SkipService<
   Inputs extends NamedCollections = NamedCollections,
   ResourceInputs extends NamedCollections = NamedCollections,
 > {
-  /** Initial data for this service's input collections. */
+  /**
+   * Initial data for this service's input collections.
+   *
+   * @remarks While the initial data is not required to have a `DepSafe` type (only a subtype of `Json` is required); note that any modifications made to any objects passed as `initialData` will *not* be seen by a service once started.
```
jberdine, line 652: We could instead constraint the type to make this comment unnecessary, but that would force users to call deepFreeze on all their initialData values. I'm unsure, but it seems that maybe here it is clear enough that this is *initial* data and the reactive computation will not react to later changes to it. But technically this is an opportunity to let staleness in.

mbouaziz (approved):

PR #636, opened 2025-01-13
Revise fetchJSON to add timeout parameter

### LINKED ISSUE #631: fetchJSON has a one second timeout
This is too aggressive in many cases.

JV asked me to file this.

jberdine:
Changes the function signature to take an object of optional parameters so
that they can be passed independently.

closes #631

### DIFF
--- a/skipruntime-ts/examples/database-client.ts
+++ b/skipruntime-ts/examples/database-client.ts
@@ -32,12 +32,9 @@ await sleep(1000);
 
 console.log('Set /user/123 to { name: "daniel", country: "UK" }');
 
-await fetchJSON(
-  `${url}/user/123`,
-  "PUT",
-  {},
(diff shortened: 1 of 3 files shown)


jberdine:
Unfortunately the api docs generator can't handle this, getting the default values wrong:

image

bennostein (approved):
Code change LGTM.

I wonder if we can explicitly specify defaults for the parameter object's fields a la https://jsdoc.app/tags-param#optional-parameters-and-default-values ?  I haven't spun up a local docs site to test that out, though

jberdine:
I tried
```
 * @param [options={timeout: 1000}] - Optional parameters.
 * @param [options.timeout=1000] - Timeout for request, in milliseconds.
```
and neither of those have any effect on the generated docs. Grr.

PR #642, opened 2025-01-14
git ignore www/package-lock.json

jberdine:


skiplabsdaniel (approved):

mbouaziz:
Or shall we do the opposite and keep all of them?
https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json states that
> This file is intended to be committed into source repositories

PR #647, opened 2025-01-15
Fix examples url

jberdine:


jberdine:
I also noticed that parts of deploying.md are not in sync with the hackernews example, but the example code hasn't changed since the docs landed, so I guess that discrepancy is intended. E.g. `backend skip` vs `backend stream` in haproxy.cfg.

bennostein (approved):

bennostein:
> I also noticed that parts of deploying.md are not in sync with the hackernews example, but the example code hasn't changed since the docs landed, so I guess that discrepancy is intended. E.g. `backend skip` vs `backend stream` in haproxy.cfg.

Yes, I think that was intended at time of writing, under the logic that it would be clearer/easier to understand.  But I'm not married to anything, feel free to change around as you see fit.

PR #649, opened 2025-01-16
fix userid type capitalization in getting started docs

spro:
`userID` -> `UserID`


mbouaziz (approved):
Thanks for taking the time to help!

PR #658, opened 2025-01-17
Make console.log on ObjectHandle transparent

### LINKED ISSUE #648: Proxied objects appear as opaque `ObjectHandle` in `console.log`
Easy to work around by calling `toString` or `JSON.stringify` explicitly, but I could see this being confusing if people are printf-debugging their skip code and seeing `ObjectHandle`s rather than their application data type. (this was me a couple of hours ago)

It would be nice, I think, if we could set up our proxies so that they're transparent to `console.log`; I tried for a bit but set it aside when I couldn't find an obvious way to do so.

mbouaziz:
Fix #648 
+ moving code from proxy to ObjectHandle


bennostein (commented):
Nice, looks good to me!

PR #660, opened 2025-01-17
[skipruntime] Manage ExternalService instance id.

skiplabsdaniel:
Prevent race condition when using same external services in multiple resources.

### DIFF
--- a/skipruntime-ts/core/src/index.ts
+++ b/skipruntime-ts/core/src/index.ts
@@ -963,6 +963,7 @@ export class ToBinding {
   SkipRuntime_ExternalService__subscribe(
     sksupplier: Handle<ExternalService>,
     writerId: string,
+    instance: string,
     resource: string,
     skparams: Pointer<Internal.CJObject>,
   ): void {
@@ -970,7 +971,7 @@ export class ToBinding {
     const supplier = this.handles.get(sksupplier);
     const writer = new CollectionWriter(writerId, this.refs());
     const params = skjson.importJSON(skparams, true) as Json;
-    supplier.subscribe(resource, params, {
+    supplier.subscribe(instance, resource, params, {
       update: writer.update.bind(writer),
       error: writer.error.bind(writer),
       loading: writer.loading.bind(writer),
@@ -979,13 +980,10 @@ export class ToBinding {
 
   SkipRuntime_ExternalService__unsubscribe(
     sksupplier: Handle<ExternalService>,
-    resource: string,
-    skparams: Pointer<Internal.CJObject>,
+    instance: string,
   ): void {
-    const skjson = this.getJsonConverter();
     const supplier = this.handles.get(sksupplier);
-    const params = skjson.importJSON(skparams, true) as Json;
-    supplier.unsubscribe(resource, params);
+    supplier.unsubscribe(instance);
   }
 
   SkipRuntime_ExternalService__shutdown(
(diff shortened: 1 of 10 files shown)


skipruntime-ts/core/src/index.ts
```diff
@@ -72,6 +72,12 @@ class Handles {
     return this.objects[id] as T;
   }
 
+  replace<T>(id: Handle<T>, v: T): T {
+    const current = this.objects[id] as T;
+    this.objects[id] = v;
+    return current;
+  }
```
jberdine: Is this used somewhere I have overlooked?

skipruntime-ts/core/src/index.ts
```diff
@@ -83,6 +89,16 @@ class Handles {
...
+
+  deleteAll<T>(): T[] {
+    const removed: T[] = [];
+    for (const value of this.objects) {
+      if (value) {
+        removed.push(value as T);
+      }
+    }
+    return removed;
+  }
```
jberdine: This too, I don't see it used.

skipruntime-ts/core/src/index.ts
```diff
@@ -56,7 +56,7 @@ export { Sum, Min, Max, Count } from "./utils.js";
 export type JSONMapper = Mapper<Json, Json, Json, Json>;
 export type JSONLazyCompute = LazyCompute<Json, Json>;
 
-class Handles {
+export class Handles {
```
jberdine: I don't see where this is used as export.

skipruntime-ts/helpers/src/external.ts
```diff
@@ -64,11 +68,13 @@ export class GenericExternalService implements ExternalService {
 }
 
 type Timeout = ReturnType<typeof setInterval>;
+type Timeouts = { [name: string]: Timeout };
 
 export class TimerResource implements ExternalResource {
-  private readonly intervals = new Map<string, { [name: string]: Timeout }>();
+  private readonly intervals: { [name: string]: Timeouts } = {};
```
jberdine: Not objecting, but FMI is this change functionally significant, or mostly just style?

skipruntime-ts/helpers/src/external.ts
```diff
@@ -31,31 +34,32 @@ export class GenericExternalService implements ExternalService {
...
-      | ExternalResource
-      | undefined;
-    if (!resource) {
-      throw new Error(`Unkonwn resource named '${resourceName}'`);
+  unsubscribe(instance: string) {
+    const resource = this.instances[instance];
+    if (resource) {
+      resource.close(instance);
+      // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+      delete this.instances[instance];
```
jberdine:
```suggestion
      delete resource;
```

jberdine (commented):
I'm not sure I fully understand this PR. There seem to be 2 changes, and I'm not sure how interconnected they are:
1. Add a call to `delete` the `ExternalResource` object in `unsubscribe`.
2. Change the api to use an instance id string instead of internally generating one from the resource and params.

Is that correct? If so, can you expand a bit on whether/how these 2 aspects are related.

jberdine:
Rephrasing: Is the issue just that the internally generated ids are not unique? Would it be reasonable to generate them internally but in a way that makes them unique?

skiplabsdaniel:
It's exactly that, but I need to have access to the generated unique Id from skiplang code to use it for unsubscribe.
It's probably possible to return generated one from subscribe method, but as I'm not sure we can have a synchronous generated unique Id from all javascript platform, I prefer to generate one from skiplang code.

jberdine (approved):
I see now, thanks. LGTM

jberdine:
FTR I had a quick look through www/docs and did not see anything that should be updated to stay in sync with these changes.

PR #661, opened 2025-01-17
Type ObjectProxy more precisely; drop unsafe cast

bennostein:
Noticed this the other day while investigating #648 ; giving a better type to `ObjectProxy`s lets us avoid the spooky `expr as unknown as type` cast


skiplang/skjson/ts/binding/src/index.ts
```diff
@@ -158,14 +158,16 @@ export type Exportable =
...
-export type ObjectProxy<Base extends { [k: string]: Exportable }> = {
-  [sk_isObjectProxy]: true;
-  [sk_managed]: true;
-  __pointer: Pointer<Internal.CJSON>;
-  clone: () => ObjectProxy<Base>;
-  toJSON: () => Base;
-  keys: IterableIterator<keyof Base>;
-} & Base;
+export type ObjectProxy<Base extends { [k: string]: Exportable }> =
+  ObjectHandle<Internal.CJObject> &
```
mbouaziz, line 162: Not sure about this, e.g. I don't think you can access `.pointer` directly

mbouaziz:
Heads-up: I force-pushed this branch to move it off the pre-rewrite history. The repository history was rewritten (~2025-03) to purge the precompiled bootstrap binaries into submodules, which re-hashed every commit after `5cc20f746`. This PR was still based on that old history, so GitHub could not compute a real diff or merge. I replayed the branch's own commits onto the new-history twin of their original base — commit contents are byte-identical, only the base SHAs changed. It is mergeable against `main` again.

mbouaziz:
Coming back to this — my 2025 review note ("I don't think you can access `.pointer` directly") turns out to be the crux, so I dug in and checked it against the current strict TS config.

The tension is fundamental: you can have **either** an `ObjectProxy` type that faithfully describes the proxy's observable surface **or** a checked (non-`unknown`) construction cast — not both. `new Proxy(target, handler)` is typed as `typeof target` (`ObjectHandle`), and TS can't model a `Proxy` retyping its target; `ObjectHandle` and the honest `ObjectProxy` don't overlap (private `fields`), so TS forces the `as unknown as` step (`TS2352`).

This PR buys the checked `as` by intersecting `ObjectProxy` with the whole `ObjectHandle` class — which is exactly what my `.pointer` note was worried about. Verified against the strict config:

- With this PR's type, `p.pointer`, `p.get("x")`, `p.has("x")`, and `p.getOwnPropertyDescriptor("x")` all typecheck as real callable/accessible members.
- At runtime the get-trap does **not** surface those — `proxy.get` resolves to `hdl.get("get")`, a *data-field* lookup. So the type asserts methods the proxy doesn't have, and `noPropertyAccessFromIndexSignature` is silently defeated for any JSON key literally named `get`/`has`/`pointer`/`getOwnPropertyDescriptor`.

The `unknown` cast here isn't hiding a *checkable* fact — no cast can verify what the handler produces — so I'd rather keep the type honest and confine the one intrinsic cast than widen the exported type to make the cast "check." I pushed that as an alternative in #1428 (extract the cast into a documented `newObjectProxy` helper; `ObjectProxy`'s `.d.ts` stays byte-identical to `main`).

Thanks for chasing the `as unknown as` down, @bennostein — the instinct was right, it's just that the safest fix is to *localize* the cast rather than *eliminate* it. Proposing we close this in favor of #1428, but happy to hear if you see it differently.

— Mehdi (Claude Opus 4.8 (1M context), high reasoning effort)

PR #662, opened 2025-01-17
[skipruntime-ts] Postgres external service

bennostein:
## Overview

Here's the Postgres-wrapper external service I mentioned yesterday!

See `skipruntime-ts/tests/src/tests.ts` for an example of how it looks/feels in user code; in summary, you specify a `PostgresExternalService` with connection info/credentials as part of your service definition, then `useExternalResource` that `service` (naming a postgres table as the resource `identifier` and specifying a key column in the `params`) and you'll have a reactively-updating `EagerCollection` of that table's data.

This is powered by `pg_notify`-calling trigger functions on the postgres database, with a relatively direct/straightforward implementation there: 
 - Each Skip service has a unique client identifier; when the Skip service subscribes to a Postgres table we create a trigger function on that table which will `pg_notify` the Skip client of any changed keys.
 - When the Skip client is notified of a changed key, it `select`s the relevant rows and updates the reactive collection.

### Random thoughts

There's a bunch of boilerplate here to handle `TEXT`/`BIGINT`/`INTEGER` keys properly, and row values are otherwise type-converted from Postgres into JS objects using the `pg` node package's standard treatment which should be fine for most common data types and is extensible/pluggable in a way that we should be able to expose to Skip users if needed.

Setting up potentially-many trigger functions on a high-load prod Postgres might not be desirable -- in that case, Josh suggested that one could "have one skip instance that is just dedicated to dealing with the notifcations, and others can mirror from it, or one per table, or whatever".  But we can cross that bridge when we come to it.

It would also be nice to support more expressive `params` than just specifying a `key` column -- in particular, maybe a `where` clause describing the data you want a reactive view on.  Doing that in an injection-safe way is beyond the capabilities of postgres' `format` utility for us; maybe the move would be to allow setting up a temporary Postgres view and then triggering pg_notify's off of that...

### Testing notes
 - Postgres tests will _pass_ by default if there's no suitable postgres running on your system, but print some warning text with instructions on how to spin that up.  If that's annoying happy to tweak behavior there.
 - I'm working on a bigger integration test that can `docker compose build` and `docker compose up` a la our hackernews example to demonstrate how this works, but put this up in the meantime.
 - ~~I'm not positive that CircleCI will be happy with me invoking docker to spin up a db; may be red for a bit until I get that all sorted out~~ all good now


mbouaziz:
That looks great! I'm wondering whether we should put this in a separate package to avoid the dependency on `pg` for our more general packages

bennostein:
Yeah, I think that's a good idea (especially as we accumulate more such integrations)

skipruntime-ts/adapters/postgres/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "license": "MIT",
+  "description": "Adapter providing reactive access to PostgreSQL from the Skip runtime",
+  "engines": {
+    "node": ">=22.6.0 <23.0.0"
+  },
+  "dependencies": {
+    "pg": "^8.13.1",
+    "pg-format": "^1.0.4",
+    "@skipruntime/core": "^0.0.5",
+    "@skipruntime/api": "^0.0.6"
```
jberdine: These need to be updated after https://github.com/SkipLabs/skip/pull/668 right, and in any case want fixed versions:
```suggestion
    "@skipruntime/core": "0.0.5",
    "@skipruntime/api": "0.0.6"
```

skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -0,0 +1,211 @@
...
+          "Error connecting to Postgres at " + JSON.stringify(db_config),
+        );
+      },
+    );
+  }
+
+  subscribe(
+    instance: string,
+    table: string,
+    params: Json,
```
jberdine: Would it make sense / be possible to refine this to something like `Json & {key: "INTEGER" | "BIGINT" | "TEXT"}`?

jberdine (approved):
LGTM just a couple comments

mbouaziz:
Do you know what's going on with the added step in the skipruntime CI job?
It takes 4 minutes to initialize the pg container but finishes with `Build was canceled`

bennostein:
> Do you know what's going on with the added step in the skipruntime CI job? It takes 4 minutes to initialize the pg container but finishes with `Build was canceled`

That's the message that gets printed was CircleCI tears everything down once all jobs are done -- and it runs in parallel with the main pipeline in the other container, not adding 4 minutes in init time.  If that doesn't line up with what you're seeing, would you link me a job and I'll take a deeper look?

PR #666, opened 2025-01-20
Fix passing limit in hn example

### LINKED ISSUE #664: hackernews example PostsResource constructor bug
https://github.com/SkipLabs/skip/blob/fd42b0158a50f9437521fbfc98adc12b4cbd0ea5/examples/hackernews/reactive_service/src/hackernews.service.ts#L93
`param` is an `object` based on

https://github.com/SkipLabs/skip/blob/fd42b0158a50f9437521fbfc98adc12b4cbd0ea5/examples/hackernews/web_service/app.py#L50

jberdine:
close #664

### DIFF
--- a/examples/hackernews/web_service/app.py
+++ b/examples/hackernews/web_service/app.py
@@ -47,7 +47,7 @@ def posts_index():
     if 'text/event-stream' in request.accept_mimetypes:
         resp = requests.post(
             f"{REACTIVE_SERVICE_URL}/streams/posts",
-            json={ 'limit': 10 },
+            json=10,
         )
         uuid = resp.text
 
@@ -56,7 +56,7 @@ def posts_index():
     else:
         resp = requests.post(
             f"{REACTIVE_SERVICE_URL}/snapshot/posts",
-            json={ 'limit': 10 },
+            json=10,
         )
 
         # The reactive service returns an array of (id, values) where


examples/hackernews/web_service/app.py
```diff
@@ -47,7 +47,7 @@ def posts_index():
     if 'text/event-stream' in request.accept_mimetypes:
         resp = requests.post(
             f"{REACTIVE_SERVICE_URL}/streams/posts",
-            json={ 'limit': 10 },
+            json=10,
```
beauby, line 50: Maybe it would be clearer to update the [`PostsResource` constructor](https://github.com/SkipLabs/skip/blob/fd42b0158a50f9437521fbfc98adc12b4cbd0ea5/examples/hackernews/reactive_service/src/hackernews.service.ts#L93) to properly deconstruct the `params` instead? Up to you, this is fine as well.

beauby (approved):

examples/hackernews/web_service/app.py
jberdine: I don't have a preference, just chose this way since the [previous change](https://github.com/SkipLabs/skip/pull/603/files#diff-62a4133aed6a78a7f4d69073606968eaa4916e6bdde24aaa02e4505cec96a4af) was to adjust the constructor to take only a number rather than a dictionary. Any opinion @bennostein ?

bennostein: No strong opinion in this case -- a bit cleaner this way, a bit more explicit as an object.  That PR just made it so you could pass arbitrary JSON as resource params, so I went with a number as it only had the one param (making clear that it didn't _have_ to be an object).

PR #667, opened 2025-01-20
Switch to exact versions in package.json

jberdine:
Fix specific dependency versions prior to hitting 1.0, as 0. versions are
unstable, see https://semver.org/#spec-item-4. The plan is to switch back to
`^` versions after 1.0 is released.


beauby (approved):
CI is red, but LGTM on the principle. Maybe explain in the commit message that we will change back to `^` style constraints once we reach 1.0 (with a link to https://semver.org/#spec-item-4)?

PR #670, opened 2025-01-21
Do not swallow error responses in hn example

jberdine:


bennostein (approved):
LGTM (though FYI I am in the process of deleting these patch requests, just letting the writes bubble up from Postgres via #662)

jberdine:
Ok cool! From a diversity of example code perspective, would it be good to keep this sort of example somewhere?

bennostein:
Makes sense to me -- I can equally set up a second example instead of modifying this one.

PR #671, opened 2025-01-21
Exclude SKDB from skipruntime-ts `build` make target

bennostein:
This target previously did an overly-broad build of all of our TS sources, slowing down local development workflow and incurring multiple unnecesasary builds in CI.

NPM's `--workspace/-w` CLI flags only offer a whitelist, not a blacklist, so this specifies all of the workspaces _except_ for those under `sql/ts/**`: the configs at the repo root, and all packages under `skiplang/**` and `skipruntime-ts/**`.

### DIFF
--- a/skipruntime-ts/Makefile
+++ b/skipruntime-ts/Makefile
@@ -1,10 +1,27 @@
 SHELL := /bin/bash
 
 SKARGO_PROFILE?=release
+NPM_WORKSPACES=\
+    -w eslint-config\
+    -w tsconfig\
+    -w skiplang/prelude/ts/binding\
+    -w skiplang/prelude/ts/wasm\
+    -w skiplang/skdate/ts\
+    -w skiplang/skjson/ts/binding\
+    -w skiplang/skjson/ts/wasm\
+    -w skipruntime-ts/api\
+    -w skipruntime-ts/core\
+    -w skipruntime-ts/helpers\
+    -w skipruntime-ts/tests\
+    -w skipruntime-ts/wasm\
+    -w skipruntime-ts/addon\
+    -w skipruntime-ts/runtime\
+    -w skipruntime-ts/server\
+    -w skipruntime-ts/examples
 
 .PHONY: build
 build:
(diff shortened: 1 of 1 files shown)


bennostein:
Looks like a ~ 40% speedup for skipruntime-native and skipruntime-wasm workflows on CircleCI

skipruntime-ts/Makefile
```diff
@@ -1,10 +1,27 @@
...
+    -w skiplang/skjson/ts/wasm\
+    -w skipruntime-ts/api\
+    -w skipruntime-ts/core\
+    -w skipruntime-ts/helpers\
+    -w skipruntime-ts/tests\
+    -w skipruntime-ts/wasm\
+    -w skipruntime-ts/addon\
+    -w skipruntime-ts/runtime\
+    -w skipruntime-ts/server\
+    -w skipruntime-ts/examples
```
jberdine: I'm not sure about quoting the shell command in the makefile, but this jq command looks like what you want:
```suggestion
NPM_WORKSPACES=$(jq --raw-output "[.workspaces[] | select(startswith(\"sql\") | not)] | map(\"-w \" + .) | .[]" package.json)
```

jberdine: Actually, `$(shell jq` not `$(jq`, and `../package.json`:
```
NPM_WORKSPACES=$(shell jq --raw-output "[.workspaces[] | select(startswith(\"sql\") | not)] | map(\"-w \" + .) | .[]" ../package.json)
```

jberdine (approved):
I'm unsure if we want to exclude examples, but LGTM

bennostein:
I'll re-include examples -- maybe I got a little aggressive stripping those out, it would be good to fail loudly if changes are breaking the examples.  Though, maybe we can exclude that once we get more rigorous testing of the examples into CI.

jberdine:
Though maybe it is enough that check-ts typechecks them.

bennostein:
> Though maybe it is enough that check-ts typechecks them.

ahh, good shout, I will re-exclude them in that case

PR #673, opened 2025-01-21
Complete CRUD for posts, fix create_post

AndreyGritsa:
The implementation might not be the best but the idea is to show all of the CRUD operations for the input collections in the example. 

Initially that wasn't quite clear for me and perhaps not only for me :)

### DIFF
--- a/examples/hackernews/web_service/app.py
+++ b/examples/hackernews/web_service/app.py
@@ -161,17 +161,7 @@ def upvote_post(post_id):
     # Write into the reactive input collection.
     resp = requests.patch(
         f"{REACTIVE_SERVICE_URL}/inputs/upvotes",
-        json=[
-            [
-                upvote_id,
-                [
-                    {
-                        "post_id": post_id,
-                        "user_id": user_id,
-                    }
-                ],
-            ]
-        ],
+        json=[[upvote_id, [{"post_id": post_id, "user_id": user_id}]]],
     )
 
     return resp.reason, resp.status_code
--- a/examples/hackernews/web_service/app.py
+++ b/examples/hackernews/web_service/app.py
@@ -2,9 +2,10 @@ from flask import Flask, json, request, redirect
 import psycopg2
 import requests
 
-REACTIVE_SERVICE_URL = 'http://reactive_cache:8081/v1'
+REACTIVE_SERVICE_URL = "http://reactive_cache:8081/v1"
 
-def get_db(): 
+
+def get_db():
     conn = psycopg2.connect(
         host="db",
         user="postgres",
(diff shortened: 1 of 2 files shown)


examples/hackernews/web_service/app.py
```diff
@@ -65,19 +68,73 @@ def posts_index():
...
+            cur.execute(f"DELETE FROM upvotes WHERE post_id={post_id}")
+            # Delete the post.
+            cur.execute(f"DELETE FROM posts WHERE id={post_id}")
+            db.commit()
+
+    # Write into the reactive input collection.
+    requests.patch(
+        f"{REACTIVE_SERVICE_URL}/inputs/upvotes",
+        json=[[upvote_id, []] for upvote_id in upvote_ids],
+    )
```
jberdine, line 115: Shouldn't the status of the first patch be checked before making the second one?

examples/hackernews/web_service/app.py
```diff
@@ -94,24 +152,31 @@ def upvote_post(post_id):
...
+            [
+                upvote_id,
+                [
+                    {
+                        "post_id": post_id,
+                        "user_id": user_id,
+                    }
+                ],
+            ]
+        ],
```
jberdine: ?

jberdine (commented):
I haven't read the whole diff carefully, just a couple minor comments so far.

You can run `make fmt` to resolve the ci failure.

examples/hackernews/web_service/app.py
AndreyGritsa: My bad, haven't noticed that auto formatting did this

examples/hackernews/web_service/app.py
AndreyGritsa: I'm wandering what's the proper way of doing it. If the first patch is failed and we do not request the second one, it means the data on the client is stale since the records form the db are deleted. Should it be re-requested? Doesn't it make an example to be more complex?

jberdine: Good point. I think it would be better to make the delete request report failure if there is any failure while servicing it. So to also keep the example simple, perhaps just save the first response and then after the second one, report status OK only if both succeeded, and otherwise report the first, say, failing reason and code.

examples/hackernews/web_service/app.py
jberdine: No worries. I guess we should add running black (or something) on the python code as part of `make fmt`.

jberdine (approved):

jberdine:
Could you rebase this PR and check that it is in a good state afterward?

AndreyGritsa:
Okay, this is probably not what I was supposed to do😅

PR #675, opened 2025-01-22
skiplang and skip docker images

mbouaziz:
Removed things that were useful for skdb but not skipruntime. I.e.
`skdb-base` = `skip` + `sqlite3`, `playwright`, `sdk`, `gradle`, `java` (3.35GB)
`skip` = `skiplang` + nodejs (2.04GB)
`skiplang` (1.27GB)

Once `skiplang` and `skip` are pushed, we can use it in the CI in place of skdb-base for most things.

Also merged the shell scripts to reduce copy-pastes.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,9 +1,9 @@
 Dockerfile
 **/target/
 node_modules/
-/compiler/bootstrap/*.ll
-/compiler/runtime/*.bc
-/compiler/runtime/*.a
-/compiler/runtime/*.o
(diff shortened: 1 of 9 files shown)


beauby:
Nice! What do you think about splitting it further:
- `skiplang-base`: everything required to build the skiplang tool chain,
- `skiplang`: `skiplang-base` + the skiplang tool chain installed system-wide (i.e. the current `skip-base`),
- `skip`: `skiplang` + nodejs.

PR #680, opened 2025-01-22
Autoformat python code

jberdine:


jberdine:
Draft, CI will fail until the dockerfile changes are included in a published image. Maybe @mbouaziz can sneak them into https://github.com/SkipLabs/skip/pull/675? :-)

mbouaziz:
LGTM
Well, I don't think I have any rights to publish docker images...

jberdine:
> Well, I don't think I have any rights to publish docker images...

Invite sent

Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apt-get update && \
     apt-get install -q -y automake clang-15 clang-format-15 curl file gawk gcc git jq lld-15 llvm-15 make nodejs parallel sqlite3 unzip zip && \
```
beauby: This is probably where you want to add the `apt-get install pip`, rather than in the first `apt-get` invocation, since the latter has an `apt-get update` happening before (the reason it is split is that the first invocation just grabs what's necessary to add (signed) apt repositories before the `apt-get update`).

beauby (approved):

mbouaziz:
Works with my latest push of docker images, but if you don't mind I'll merge #675 first

mbouaziz (approved):

PR #682, opened 2025-01-22
[skipruntime/core] Fix dependency

skiplabsdaniel:


beauby (approved):

mbouaziz:
Why wasn't it broken in the CI? Or maybe it was?

skiplabsdaniel:
> Why wasn't it broken in the CI? Or maybe it was?

This is due to the way `npm` manages packages globally at the root of the project. `@skipruntime/core` sees `@skip-wasm/std` when it shouldn't.

PR #684, opened 2025-01-23
Add support for SERIAL/BIGSERIAL keys in PG adapter

bennostein:


bennostein:
Publishing this version as`@skip-adapter/postgres` 0.0.4 and landing without review -- feel free to comment/review after the fact and I'll follow up.

PR #686, opened 2025-01-23
Reduce the number of package to publish

skiplabsdaniel:
- Internal packages are managed with symbolic links.

### DIFF
--- a/prebuild.mjs
+++ b/prebuild.mjs
@@ -1,0 +1,33 @@
+import * as fs from "fs";
+import * as path from "path";
+import { fileURLToPath } from "url";
+
+const symlinks = {
+  "skiplang-std": "./skiplang/prelude/ts/binding/src",
+  "skipwasm-std": "./skiplang/prelude/ts/wasm/src",
+  "skipwasm-date": "./skiplang/skdate/ts/src",
+  "skiplang-json": "./skiplang/skjson/ts/binding/src",
+  "skipwasm-json": "./skiplang/skjson/ts/wasm/src",
+};
+
+const filename = fileURLToPath(import.meta.url);
+const dirname = path.dirname(filename);
+
+export function link(packages) {
+  for (const [simlink, target] of Object.entries(symlinks)) {
+    if (!packages.has(simlink)) continue;
+    if (!fs.existsSync(simlink)) {
+      fs.symlinkSync(path.resolve(dirname, target), simlink, "dir");
+    }
+  }
+}
+
+if (process.argv.length <= 2) {
+  const thisBin = process.argv[1];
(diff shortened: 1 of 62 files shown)


prebuild.mjs
```diff
@@ -0,0 +1,34 @@
+import * as fs from "fs";
+import * as path from "path";
+import { fileURLToPath } from "url";
+
+const simlinks = {
```
bennostein: typo: symlinks (and same below for simlink)

bennostein (commented):
LGTM -- it'll be nice to clean up the namespaces / "skip" search results on the NPM registry.

As far as I can tell this doesn't affect any public-facing interfaces of `@skipruntime/*` packages -- is that right?  There are a couple imports from `@skipruntime/core/internal.js` and `@skipruntime/core/json.js` in the addon code which raised my eyebrows, but that's no different from before, I suppose.

Another thought is that the `prebuild.mjs` step is going to create a lot of symlinks; separate ones for each subpackage.  Is there a reason we have to do that?  or might it be possible to put them in one shared location, e.g. `./skipruntime-ts` or `./skipruntime-ts/internals` or something like that?

skiplabsdaniel:
> LGTM -- it'll be nice to clean up the namespaces / "skip" search results on the NPM registry.
> 
> As far as I can tell this doesn't affect any public-facing interfaces of `@skipruntime/*` packages -- is that right? There are a couple imports from `@skipruntime/core/internal.js` and `@skipruntime/core/json.js` in the addon code which raised my eyebrows, but that's no different from before, I suppose.
> 
> Another thought is that the `prebuild.mjs` step is going to create a lot of symlinks; separate ones for each subpackage. Is there a reason we have to do that? or might it be possible to put them in one shared location, e.g. `./skipruntime-ts` or `./skipruntime-ts/internals` or something like that?

Creating links in this way means that only the files required for each package are packed. It could be considered that delivering the same file several times in different packages is not a problem, but in this case, with the use of `unique symbol`, we would end up with type conflicts with types which are theoretically the same.
We could reduce the number of links by not checking the internal packages individually, with the risk of writing a type in the wrong place.
And just for information, as things stand, a tool like `pnpm` would create roughly the same number of links.

skiplabsdaniel:
> @skiplabsdaniel , do you see any problems with this? It undoes your change as mentioned, but my impression from reading that is that the change was incidental.

As the process of using the native version is more complicated, the purpose of using initService was to make the use of the `Wasm` or the `native` version explicit, so that it is a conscious choice for the user. It's true that it can be done like that, but perhaps a message is missing when the `Wasm` version is used.

bennostein (approved):

PR #688, opened 2025-01-24
Fix broken doc links to put and patch

jberdine:


package-lock.json
```diff
@@ -387,6 +396,25 @@
...
+      "version": "0.0.4",
+      "resolved": "https://registry.npmjs.org/@skipruntime/addon/-/addon-0.0.4.tgz",
+      "integrity": "sha512-YyXRJ9+ZDu0uzYnVfRc/BczlOrFh8akSuwvOKo6yvWYh3rYv04CJKfMjLdZGyaK4D5wh+hJKrNbU2MZDOkGESQ==",
+      "hasInstallScript": true,
+      "optional": true,
+      "dependencies": {
+        "@skipruntime/core": "^0.0.5"
+      }
+    },
+    "node_modules/@skipruntime/addon/node_modules/@skipruntime/core": {
```
mbouaziz: Are sub-`node_modules` expected?

beauby: This indeed seems fishy and is caused by the explicit call to `npm install` at the end of [the npm build script](https://github.com/SkipLabs/skip/blob/main/skipruntime-ts/addon/package.json#L11).

bennostein: Much (all?) of the package-lock.json stuff here is the result of stale versions in `examples/hackernews/reactive_service/package.json` which are fixed by #691. Should be fine after that lands and this rebases.

The actual docs changes LGTM though!

bennostein (approved):

PR #690, opened 2025-01-24
Fix postgres setup/teardown sequencing

bennostein:
2 unrelated fixes:
 
 * the changes I made to the postgres test setup to accomodate running wasm/native clients separately against in f9fc0f9661bf8f4098676c2a9ea4fc66eafb148a meant that the "nice" warning message printed on systems without a running postgres instance was instead just an exception.
 * Mehdi noticed some error messages in CI on teardown, with "drop function" pointing at functions that don't exist.  This should fix that -- the issue was a race with the runtime calling both unsubscribe and shutdown on the postgres service.

### DIFF
--- a/skipruntime-ts/adapters/postgres/package.json
+++ b/skipruntime-ts/adapters/postgres/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@skip-adapter/postgres",
-  "version": "0.0.4",
+  "version": "0.0.5",
   "type": "module",
   "exports": {
     ".": "./dist/index.js"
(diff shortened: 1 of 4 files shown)


bennostein:
Looks all good in CI now. The postgres container still shows up as grey rather than green in CircleCI detailed view, but I think that makes sense semantically -- it's not a step that succeeds/fails but rather a bit of scaffolding that runs alongside the actual tests.

mbouaziz (approved):

mbouaziz:
The CI looks better indeed. We should find a way to differentiate the good grey from the bad grey...

PR #691, opened 2025-01-24
Update HN example with new packages

### LINKED ISSUE #677: Update hn example after publishing next npm packages
From https://github.com/SkipLabs/skip/pull/668/commits/445a3a2cf38c503210b9a2480a52bf229f54add6:

Need to be updated after next publication:
- add  "@skipruntime/wasm" dependency
- hackernews.service.ts
  ```"@skipruntime/api" => "@skipruntime/core"```
- server.js
  ```
  import { initService } from "@skipruntime/wasm";
  const instance = await     initService(serviceWithInitialData(posts, users, upvotes));
  runService(instance);
  ```

bennostein:
Use the new postgres adapter code in the hackernews example, dropping the stopgap setup of writing to the reactive service's input collections from the web service and instead just writing to the DB and letting data bubble up to the reactive layer.

Also close #677 by updating all of our package versions and service initializaiton code.

### DIFF
--- a/examples/hackernews/reactive_service/server.js
+++ b/examples/hackernews/reactive_service/server.js
@@ -1,15 +1,5 @@
-import postgres from "postgres";
 import { runService } from "@skipruntime/server";
-import { serviceWithInitialData } from "./dist/hackernews.service.js";
+import { initService } from "@skipruntime/wasm";
+import { service } from "./dist/hackernews.service.js";
 
-const sql = postgres("postgres://postgres:change_me@db:5432");
-async function selectAll(table) {
-  return (await sql`SELECT * FROM ${sql(table)}`).map((r) => [r.id, [r]]);
-}
-
-// Initial values.
-const posts = await selectAll("posts");
-const users = await selectAll("users");
-const upvotes = await selectAll("upvotes");
-
-runService(serviceWithInitialData(posts, users, upvotes));
+runService(await initService(service));
--- a/examples/hackernews/reactive_service/package.json
+++ b/examples/hackernews/reactive_service/package.json
@@ -9,9 +9,10 @@
   "license": "MIT",
   "description": "",
   "dependencies": {
-    "postgres": "^3.4.5",
-    "@skipruntime/api": "0.0.6",
-    "@skipruntime/server": "0.0.8"
+    "@skipruntime/core": "0.0.6",
+    "@skipruntime/wasm": "0.0.7",
+    "@skipruntime/server": "0.0.9",
+    "@skip-adapter/postgres": "0.0.5"
   },
   "devDependencies": {
     "@skiplabs/eslint-config": "^0.0.1",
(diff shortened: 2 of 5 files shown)


examples/hackernews/reactive_service/server.js
```diff
@@ -1,15 +1,5 @@
-import postgres from "postgres";
 import { runService } from "@skipruntime/server";
-import { serviceWithInitialData } from "./dist/hackernews.service.js";
+import { initService } from "@skipruntime/wasm";
```
jberdine, line 2: We shouldn't need to import from internal packages in examples. Looking around, I think that the package reorg left runService uncallable using only public apis, so probably we should export initService from core/src/api.ts before merging this PR. I don't know whether to do that before or after https://github.com/SkipLabs/skip/pull/686.

jberdine (changes requested):
LGTM but for one issue inline

examples/hackernews/reactive_service/server.js
bennostein: It is publicly callable via @skipruntime/server though -- that's what's happening here, and the example works resolving everything through the NPM registry

jberdine: Maybe I'm confused but I don't see ServiceInstance or initService exported from @skipruntime/server?

jberdine: (I'm not saying that this code doesn't actually work, just that it is pointing out that people need to use things from internal packages, that are also not documented.)

bennostein: Ahh, I see what you mean -- I was referring to `runService` not `initService` as being callable via `@skipruntime/server`

As it stands, user code needs to call the`initService` from either `@skipruntime/wasm` or `@skipruntime/native`.  I'm not sure this is a bad thing though (outside of the question of documentation) since they initialize the service differently but are both exported publicly.  Under the current dependency structure, it would be difficult to expose `initService` at the level of `@skipruntime/core`

(`ServiceInstance` is already exported in `@skipruntime/core` so that's fine)

jberdine: So as it stands, user code needs to change depending on whether the wasm or native version is desired? I didn't appreciate that. Ok, this PR is not making anything worse. I'm not sure yet what to do about this situation though.

jberdine (approved):

examples/hackernews/reactive_service/server.js
bennostein: I've got a draft PR baking adding the two `initService`s to the docusaurus site, but there's a bit of a question of what to do with `ServiceInstance` -- we don't want its actual structure to be part of the public API, but it is needed as an opaque/existential type, i.e. the thing produced by `initService` and passed to `runService`.

I've set that aside to work on other things for now, but we should chat about how to make things better :)

PR #668, opened 2025-01-20
[skipruntime] Reduce the number of packages.

skiplabsdaniel:
-- The `addon` install/build process need to be updated and documented.
-- A Other PR is coming with a tool that regroup all internal packages `@skiplang/*` and `@skipwasm/*`  in respectively  `@skipruntime/core` and  `@skipruntime/wasm`

PR #694, opened 2025-01-28
Avoid explicit `initService` in user code by dynamic dispatch in `runService`

bennostein:
Previously (before #668) we passed a `SkipService` _definition_ to `runService`, which figured out which runtime (wasm or native) was available, initialized the service, and spun up an HTTP server. Since that PR, `runService` expects an initialized service, leaving it up to client code which runtime to use.

This PR returns us to the pre-#668 status quo, dispatching to the proper `initService` from within `runService` depending on what's available at runtime.  The main benefits are to simplify the public interface, allow code to be more portable across platforms, and reduce the surface of APIs to document/expose.

@skiplabsdaniel , do you see any problems with this? It undoes your change as mentioned, but my impression from reading that is that the change was incidental.

cc @jberdine re: comments on #691

### DIFF
--- a/skipruntime-ts/examples/database.ts
+++ b/skipruntime-ts/examples/database.ts
@@ -6,7 +6,6 @@ import type {
 } from "@skipruntime/core";
 
 import { runService } from "@skipruntime/server";
-import { initService } from "@skipruntime/wasm";
 
 import Database from "better-sqlite3";
 
@@ -79,9 +78,7 @@ const data: Entry<string, User>[] = db
 
 db.close();
 
-const instance = await initService(serviceWithInitialData(data));
-
-const closable = runService(instance, {
+const closable = await runService(serviceWithInitialData(data), {
   streaming_port: 8080,
   control_port: 8081,
 });
--- a/skipruntime-ts/examples/database.ts
+++ b/skipruntime-ts/examples/database.ts
@@ -78,13 +78,13 @@ const data: Entry<string, User>[] = db
 
 db.close();
 
-const closable = await runService(serviceWithInitialData(data), {
+const server = await runService(serviceWithInitialData(data), {
   streaming_port: 8080,
   control_port: 8081,
 });
 
 function shutdown() {
-  closable.close();
+  server.close();
 }
 
 process.on("SIGTERM", shutdown);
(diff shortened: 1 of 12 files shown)


skiplabsdaniel:
> @skiplabsdaniel , do you see any problems with this? It undoes your change as mentioned, but my impression from reading that is that the change was incidental.

As the process of using the native version is more complicated, the purpose of using `initService` was to make the use of the Wasm or the native version explicit, so that it is a conscious choice for the user. It's true that it can be done like that, but perhaps a message is missing when the Wasm version is used.
In addition, having a server that is not aware of the different versions would allow it to be used with other possible versions of `skipruntime`.

jberdine:
> As the process of using the native version is more complicated, the purpose of using `initService` was to make the use of the Wasm or the native version explicit, so that it is a conscious choice for the user. It's true that it can be done like that, but perhaps a message is missing when the Wasm version is used.

Do you think it would suffice if `initService` was as in this PR but took an additional argument, say of type `"wasm" | "native" | "auto"`, that users could use to indicate which version was desired, where the `"wasm"` and `"native"` alternatives would fail if the specified version was not available?

jberdine:
> NB: stacked on top of #693 as there is some overlap and it was useful for testing, but the core of this PR is just its changes under `skipruntime-ts/server`.

Note that if you push these PR branches to origin, then you can specify a base branch different from main in order to stack them, and have github show the diff only for the commits that are specific to the stacked PR. The downside is that the stacked PRs need to be merged in order, or else github will merge a stacked PR into the PR on which it is stacked.

skipruntime-ts/examples/database.ts
```diff
@@ -79,9 +78,7 @@ const data: Entry<string, User>[] = db
 
 db.close();
 
-const instance = await initService(serviceWithInitialData(data));
-
-const closable = runService(instance, {
+const closable = await runService(serviceWithInitialData(data), {
```
jberdine: While you are here, note that at some point `runService` got a return type named `SkipServer`, so renaming `closable` to `server` might make sense. And similarly for the other examples.

skipruntime-ts/server/src/server.ts
```diff
@@ -81,22 +81,68 @@ export type SkipServer = {
  *
  * @typeParam Inputs - Named collections from which the service computes.
  * @typeParam ResourceInputs - Named collections provided to resource computations.
- * @param instance - The Service instance.
+ * @param service - The SkipService definition to run
```
jberdine:
```suggestion
 * @param service - The SkipService definition to run.
```

jberdine (approved):
Nice!

bennostein:
The red CI on #697 is the result of these changes (it's stacked on top) so not merging until I get that sorted out.

Automating those tests already paying off!

PR #696, opened 2025-01-29
Generalize param encoding of Polled

### LINKED ISSUE #687: useExternalResource Polled design consideration
Current usage of Polled is
`externalServices: {
      service: new GenericExternalService({
        api: new Polled(
         url,
          ...
        ),
      }),
    },` and then `context.useExternalResource({
      service: "service",
      identifier: "api",
      params: params,
    })`. 

The problem is that if for some reason I need to change not only query params but the url itself, say an external service expects an API endpoint to be built like `/articles/:articleId/comments`. As for now in this case I need to implement custom Polled where url is passed as part of the `params`. 

Preferred feels to be this way 
 `context.useExternalResource({
      service: "service",
      identifier: "api",
      url: url,
      params: params,
    })`.

jberdine:
Move the responsibility for generating the `?` query string separator from
the `Polled` constructor to the `encodeParams` function. This makes it
possible to use some params to construct the part of the URL preceding the
`?`.

close #687

### DIFF
--- a/skipruntime-ts/helpers/src/external.ts
+++ b/skipruntime-ts/helpers/src/external.ts
@@ -118,8 +118,8 @@ function defaultParamEncoder(params: Json): string {
       if (typeof value == "object") queryParams[key] = JSON.stringify(value);
       else queryParams[key] = value.toString();
     }
-    return new URLSearchParams(queryParams).toString();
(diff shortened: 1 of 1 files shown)


AndreyGritsa:
Looks good for me. Thanks!

bennostein (approved):

PR #697, opened 2025-01-29
Test examples in CI; refactor/cleanup examples to get everything up to date.

bennostein:
Add expectation tests on each of the example scenarios in `skipruntime-ts/examples`.

Various fixes and cleanups were required in the examples themselves (I ripped out all of the old "scenarios" code; it's much easier to use our APIs directly now that they've been built), but all basically semantics-preserving.


mbouaziz:
That looks great!
- I'll let @skiplabsdaniel have a look at the removal of scenarios and the REPL to run examples.
- Is there a way to merge the bash scripts?
- Should we do something about the `[Object]` in exp files?
- Could you add a Makefile target to regenerate all exp files?

bennostein:
>     * I'll let @skiplabsdaniel have a look at the removal of scenarios and the REPL to run examples.
:+1:

>     * Is there a way to merge the bash scripts?

Probably but it's a bit of a pain; I did it this way because they have varying numbers of background/server tasks to run.

>     * Should we do something about the `[Object]` in exp files?

Good call, can do

>     * Could you add a Makefile target to regenerate all exp files?

Sure

jberdine:
I can repro the CI failure locally. Is it already understood?

bennostein:
> I can repro the CI failure locally. Is it already understood?

~~I'm working on repro'ing and understanding, not there yet.  `make test-examples` worked for me locally before pushing, but maybe something peculiar with my environment.~~

I can repro too, working on it

jberdine:
I'd say that it would be best to get this in sooner than later, and we can react to any comments Daniel has, such as on the scenario-running stuff, later.

PR #698, opened 2025-01-29
Use apt-install.sh in Dockerfile

mbouaziz:
- Avoid duplicate code
- Preparation for updating LLVM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,8 @@
 FROM ubuntu:22.04 AS skiplang-base
 
 ENV DEBIAN_FRONTEND=noninteractive
-RUN apt-get update && \
-    apt-get install -q -y wget gnupg && \
-    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
-    echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main" >> /etc/apt/sources.list.d/llvm.list && \
-    echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main" >> /etc/apt/sources.list.d/llvm.list && \
-    apt-get update && \
-    apt-get install -q -y automake clang-15 clang-format-15 curl file gawk gcc git jq lld-15 llvm-15 make parallel unzip zip
-
-RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-15 100 && \
-    update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-15 100 && \
-    update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-15 100 && \
-    update-alternatives --install /usr/bin/llc llc /usr/bin/llc-15 100 && \
-    update-alternatives --install /usr/bin/llvm-ar llvm-ar /usr/bin/llvm-ar-15 100 && \
-    update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-15 100 && \
-    update-alternatives --install /usr/bin/llvm-link llvm-link /usr/bin/llvm-link-15 100 && \
-    update-alternatives --install /usr/bin/wasm-ld wasm-ld /usr/bin/wasm-ld-15 100
+RUN --mount=type=bind,source=./bin/apt-install.sh,target=/tmp/apt-install.sh \
+    /tmp/apt-install.sh skiplang-build-deps
 
 ENV CC=clang
 ENV CXX=clang++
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,22 +1,8 @@
...
-RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-15 100 && \
-    update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-15 100 && \
-    update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-15 100 && \
-    update-alternatives --install /usr/bin/llc llc /usr/bin/llc-15 100 && \
-    update-alternatives --install /usr/bin/llvm-ar llvm-ar /usr/bin/llvm-ar-15 100 && \
-    update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-15 100 && \
-    update-alternatives --install /usr/bin/llvm-link llvm-link /usr/bin/llvm-link-15 100 && \
-    update-alternatives --install /usr/bin/wasm-ld wasm-ld /usr/bin/wasm-ld-15 100
+COPY ./bin/apt-install.sh /tmp/
+RUN /tmp/apt-install.sh llvm
```
jberdine: Just curious, why copy instead of running it from the original location?

bin/apt-install.sh
```diff
@@ -1,22 +1,58 @@
...
+for step in "${steps[@]}"; do
+    case "$step" in
+        llvm)
+            apt-get update
+            apt-get install -q -y wget gnupg
+            wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+            echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-$LLVM_VERSION main" >> /etc/apt/sources.list.d/llvm.list
+            echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-$LLVM_VERSION main" >> /etc/apt/sources.list.d/llvm.list
+            apt-get update
+            apt-get install -q -y automake clang-$LLVM_VERSION clang-format-$LLVM_VERSION curl file gawk gcc git jq lld-$LLVM_VERSION llvm-$LLVM_VERSION make parallel unzip zip
```
jberdine: I find naming this step "llvm" confusing since many of the packages installed here are nothing to do with llvm, just other deps we have. The real distinction in the steps is between things that are needed to build the skiplang toolchain and the rest. I'm not 100% sure that the existing code follows that, or if some thing could be delayed.

jberdine (commented):
Seems mostly ok, a couple comments inline, but I am not understanding the motivation for splitting out apt-install, it seems like an added complication. Is the plan that it will be used in multiple places?

beauby:
Where is the `apt-install.sh` script used other than this `Dockerfile`? If nowhere, what is the intent of having it as a shell script rather than in the `Dockerfile`?

Dockerfile
mbouaziz: Like that?
```
RUN --mount=type=bind,source=./bin/apt-install.sh,target=/tmp/apt-install.sh /tmp/apt-install.sh llvm
```

bin/apt-install.sh
mbouaziz: Not sure either, I'll rename it to `skiplang-dev-deps`

mbouaziz:
`apt-install.sh` was added recently (#668), it was probably meant to be referenced in the docs

Dockerfile
jberdine: Sorry, I didn't expect it to be so involved. I was thinking `RUN ./bin/apt-install.sh`... :-)

bin/apt-install.sh
```diff
@@ -1,22 +1,58 @@
...
+            update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-$LLVM_VERSION $PRIORITY
+            update-alternatives --install /usr/bin/llvm-link llvm-link /usr/bin/llvm-link-$LLVM_VERSION $PRIORITY
+            update-alternatives --install /usr/bin/wasm-ld wasm-ld /usr/bin/wasm-ld-$LLVM_VERSION $PRIORITY
+            ;;
+        node)
+            wget -O - https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | apt-key add -
+            echo "deb https://deb.nodesource.com/node_22.x nodistro main" >> /etc/apt/sources.list.d/nodejs.list
+            apt-get update
+            apt-get install -q -y nodejs
+            npm install -g bun
```
jberdine: With the addon only working with npm, bun is now unused except for some dead test skdb make targets and dead examples test targets. Maybe it could be left out.

bin/apt-install.sh
```diff
@@ -1,22 +1,58 @@
...
+            npm install -g prettier
```
jberdine: This ought to go below with black I guess, they have the same status.

PR #702, opened 2025-01-30
[skiplang/skstore] Fix locking in `runWithGcIntern`.

beauby:
When calling `runWithGc()` without a `synchronizer`, the global lock was released after each iteration, and never re-acquired.

The code path did not seem to be actively used, as it would systematically crash.

### DIFF
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -1566,6 +1566,9 @@ fun runWithGcIntern(
     }) match {
     | Some(x) -> x
     | None() ->
+      if (synchronizer is None _) {
+        gGlobalUnlock();
+      };
       destroyObstack(pos);
       return context
     };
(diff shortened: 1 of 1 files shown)


skiplang/prelude/src/skstore/Context.sk
```diff
@@ -1569,7 +1572,6 @@ fun runWithGcIntern(
           sync,
           underLockF,
         );
-        gGlobalUnlock();
```
mbouaziz, line 1584: Don't we have the same issue in the `CStop` branch?

mbouaziz (commented):
There's a lot that could be factored in these 4 branches

skiplang/prelude/src/skstore/Context.sk
beauby: I don't think so, as (in the synchronizer-free case) the lock is now acquired at the beginning, held through all iterations, and released after processing the `CStop()` which triggers the end of the loop.

beauby:
> There's a lot that could be factored in these 4 branches

Agreed. IIUC @skiplabsdaniel is working on cleaning up those APIs. The scope of this PR is only to fix the glaring bug.

mbouaziz (approved):

skiplabsdaniel (approved):

PR #709, opened 2025-01-31
Add @skipruntime/skip meta-package

bennostein:
I think this should do the trick and let people `npm i @skipruntime/skip` instead of the current `npm i @skipruntime/core @skipruntime/server @skipruntime/helpers`.

Will also be helpful, I think, to keep versions in lockstep for alpha adopters.

### DIFF
--- a/skipruntime-ts/metapackage/package.json
+++ b/skipruntime-ts/metapackage/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "@skiplabs/skip",
+  "version": "0.0.1",
+  "dependencies": {
+    "@skipruntime/core": "0.0.6",
+    "@skipruntime/server": "0.0.9",
+    "@skipruntime/helpers": "0.0.9"
+  },
+  "optionalDependencies": {
+    "@skipruntime/native": "0.0.5",
+    "@skipruntime/wasm": "0.0.7",
+    "@skip-adapter/postgres": "0.0.5"
+  }
+}
--- a/skipruntime-ts/metapackage/package.json
+++ b/skipruntime-ts/metapackage/package.json
@@ -4,11 +4,11 @@
   "dependencies": {
     "@skipruntime/core": "0.0.6",
     "@skipruntime/server": "0.0.9",
-    "@skipruntime/helpers": "0.0.9"
+    "@skipruntime/helpers": "0.0.9",
+    "@skipruntime/wasm": "0.0.7"
   },
   "optionalDependencies": {
     "@skipruntime/native": "0.0.5",
-    "@skipruntime/wasm": "0.0.7",
     "@skip-adapter/postgres": "0.0.5"
   }
 }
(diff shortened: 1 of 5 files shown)


mbouaziz:
Should we also install `@skipruntime/wasm` by default so people have something that works?

bennostein:
Yeah, makes sense to me -- the argument against is that it pulls in an unnecessary dependency for a prod deployment using the native runtime, but if that's a problem you can just manually specify each subpackage.

skipruntime-ts/metapackage/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "0.0.1",
+  "dependencies": {
+    "@skipruntime/core": "0.0.6",
+    "@skipruntime/server": "0.0.9",
+    "@skipruntime/helpers": "0.0.9",
+    "@skipruntime/wasm": "0.0.7"
+  },
+  "optionalDependencies": {
+    "@skipruntime/native": "0.0.5",
+    "@skip-adapter/postgres": "0.0.5"
```
jberdine, line 12: Maybe just me, but wouldn't it be easier to keep track of if we made all these version numbers the same, e.g. 0.0.10? There is no issue with skipping some version numbers, right?

bennostein: Sure, but the reason we ended up in this state (and may want to continue this way) is that we sometimes update a "leaf" package without updating all of its parents.  We could enforce that everything's at the same version number, but it would mean that every release has to update everything. Not sure which way is preferable.

jberdine: I'm not sure, but I would have defaulted to something that amounts to a function from the git commit hash of the whole repo to a single version number to be used for each of the packages in the repo, perhaps just as a matter of being conservative wrt what the dependencies actually are and which combinations of package versions we test versus what users can possibly have.

I have also been trying to keep the docs site in sync with the released packages, and that would certainly be easier if each git commit corresponded to a single version number.

bennostein: Yeah, bumping everything all together whenever we make a change seems reasonable to me -- esp. with docs, if/when we start having to host multiple versioned instances of the docs site.  And it's not too much trickier to update all of the packages than to update just one package, especially if we maintain the invariant that they all share a package number.

Let's chat with rest of the team and decouple from this PR though, I think

jberdine (approved):

skipruntime-ts/metapackage/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@skiplabs/skip",
+  "version": "0.0.1",
+  "dependencies": {
+    "@skipruntime/core": "0.0.6",
+    "@skipruntime/server": "0.0.9",
+    "@skipruntime/helpers": "0.0.9",
```
jberdine, line 7: Sorry to be late, but is helpers possibly better as an optional dep?

PR #708, opened 2025-01-30
Add detail to postgres adapter error messages

bennostein:
closes #707

PR #710, opened 2025-01-31
Tweak postgres unsubscribe flow to avoid printing spurious error

bennostein:
Related to previous issues with sequencing of unsubscribe/shutdown, this stops printing spurious error messages surfaced by #708 .

### DIFF
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -230,17 +230,18 @@ FOR EACH ROW EXECUTE FUNCTION %I();`,
   }
 
   unsubscribe(instance: string): void {
-    this.client
-      .query(format("DROP FUNCTION IF EXISTS %I CASCADE;", instance))
-      .then(
-        () => this.open_instances.delete(instance),
-        (e: unknown) => {
-          console.error(
-            "Error unsubscribing from resource instance: " + instance,
-          );
-          throw e;
-        },
-      );
+    if (this.open_instances.has(instance))
+      this.client
+        .query(format("DROP FUNCTION IF EXISTS %I CASCADE;", instance))
+        .then(
+          () => this.open_instances.delete(instance),
+          (e: unknown) => {
+            console.error(
+              "Error unsubscribing from resource instance: " + instance,
+            );
+            throw e;
+          },
+        );
   }
 
   shutdown(): void {


skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -230,17 +230,19 @@ FOR EACH ROW EXECUTE FUNCTION %I();`,
...
+    if (this.open_instances.has(instance))
+      this.client
+        .query(format("DROP FUNCTION IF EXISTS %I CASCADE;", instance))
+        .then(
+          () => this.open_instances.delete(instance),
+          (e: unknown) => {
+            console.error(
+              "Error unsubscribing from resource instance: " + instance,
+            );
+            console.log(e);
```
beauby: Do we want to log before throwing here? If so, do we want to do it consistently throughout the postgres adapter?

skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -230,17 +230,19 @@ FOR EACH ROW EXECUTE FUNCTION %I();`,
...
-      .then(
-        () => this.open_instances.delete(instance),
-        (e: unknown) => {
-          console.error(
-            "Error unsubscribing from resource instance: " + instance,
-          );
-          throw e;
-        },
-      );
+    if (this.open_instances.has(instance))
```
beauby, line 233: Do we want to throw a specific error in case the user attempts to `unsubscribe()` from an instance not in `open_instance`, as it would likely indicate a bug in user code (vs this package)?

beauby (approved):

skipruntime-ts/adapters/postgres/src/index.ts
bennostein: Nope, that's debugging code that I should strip back out...

skipruntime-ts/adapters/postgres/src/index.ts
bennostein: At least for the time being, it doesn't necessarily indicate a bug in user code -- on the native side, `closeService_` calls `unsubscribe` on each resource instance as well as `shutdown` on the whole external service; I don't want to mess around with that or its ordering without consulting Daniel.

So this works around the semantics there, making the operations idempotent and also doing the "right" thing for isolated calls to unsubscribe and shutdown.

There could be cases where a user double-unsubscribes or unsubscribes an instance that never actually existed, but I'm not sure whether it's worth trying to catch that.

PR #711, opened 2025-01-31
More precise/descriptive SkipError exceptions

bennostein:
I've been meaning to do this for a while... move us away from `throw new Error(...)` with the generic JS "Error" type and instead set up a simple hierarchy of `SkipError`s which give a bit more information/context about the exception and allow client code to selectively catch/handle as needed.

To my eyes, there's not much downside here but I may not be considering something -- what do you guys think?

I also put in a very simple best-effort syntactic check that new code doesn't `throw new Error` but that's coarse and will only catch the most egregious cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,7 @@ test-skjson:
 
 .PHONY: test-skipruntime-ts
 test-skipruntime-ts:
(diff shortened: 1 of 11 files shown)


mbouaziz (approved):
I like this

PR #714, opened 2025-01-31
Include error types in public api

jberdine:
Users might need to know about them, and they are referenced from the docs.


bennostein (approved):
Good call, lgtm

PR #718, opened 2025-02-03
Remove 'auto' option for skip runtime platform

bennostein:


jberdine (approved):
I guess there is no loss in removing it as "auto" is currently broken anyhow.

PR #719, opened 2025-02-03
Add error handling around CollectionWriter FFI

bennostein:
~~I'm chasing other FFI calls to convince myself there are no other gaps in the error handling logic, but submitting this change for now.~~
EDIT: After covering the three CollectionWriter methods, I audited all of the other `getErrorHdl` on the native side and confirmed that they're covered by corresponding logic on the JS side.  So we should be watertight now.


jberdine (approved):
I haven't redone the audit for other call sites, but LGTM.

PR #723, opened 2025-02-04
Update NPM package versions

bennostein:
This takes all of our Skip runtime packages to 0.0.10, increments all of our core `@skip-wasm/*` and `@skiplang/*` packages (some of which are already at 1.0.x, and none of which users will install, so I left out of the everything-to-0.0.10 edict), and stops versioning the test package.json.


jberdine (approved):
LGTM as long as CI agrees. Too bad some packages are already at 1.0 (they really should not be right?) and some still have their own version numbers.

PR #727, opened 2025-02-05
Parallelize test_diff

mbouaziz:
Fix and parallelize this test to speed up CI

### DIFF
--- a/sql/test_diff.sh
+++ b/sql/test_diff.sh
@@ -13,12 +13,18 @@ pass() { printf "%-50s OK\n" "$1:"; }
 fail() { printf "%-50s FAILED\n" "$1:"; }
 
 run_diff () {
+    creation_script=$1
+    shift
+    views_script=$1
+    shift
+    more_scripts=("$@")
+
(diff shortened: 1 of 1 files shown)


mbouaziz:
CircleCI's not happy

bennostein (approved):
Changes look good to me and this branch passes `make test-native` for me locally.

Not sure why CircleCI is mad, though..

mbouaziz:
I suspect `parallel` gets the wrong value about free memory, we may have to hard-code a `--jobs` in this case

jberdine:
Since skdb starts by allocating 16GB which isn't resident until used later, I expect that parallel lets many start as the machine RSS looks ok initially, and then the workers get into trouble when they use the memory they have allocated. Using `--memfree` might work better since when memory pressure builds the younger workers can be killed to make room. Still, running multiple 16GB jobs on an 8GB machine is tricky.

mbouaziz:
Explicitly specifying `--jobs 8` works for now. It reduces `skdb` run time from 44min to 22min

PR #730, opened 2025-02-05
Handle postgres time-stamp types

bennostein:
The default type parser from the `pg` NPM library converts these to `Date` objects which then isn't handled nicely in the Skip heap.  This instead just keeps the raw string, which clients can handle as they choose.

Also adds test coverage of a non-lowercase column, which came up in some user code.

### DIFF
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -14,6 +14,10 @@ import {
 import pg from "pg";
 import format from "pg-format";
 
+// Pass timestamp strings straight through instead of attempting to convert to JS Date object, which would be clobbered in the Skip heap
+pg.types.setTypeParser(pg.types.builtins.TIMESTAMP, (x: string) => x);
+pg.types.setTypeParser(pg.types.builtins.TIMESTAMPTZ, (x: string) => x);
+
 /**
  * Exception indicating an error while establishing the connection between PostgreSQL and Skip.
  * @hideconstructor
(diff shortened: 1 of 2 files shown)


jberdine (approved):
LGTM but makes me wonder what really goes wrong and how we can prevent people writing their own adapters from hitting similar issues.

mbouaziz:
Could we convert them to integers instead?
Ordering on string will be wrong

bennostein:
> LGTM but makes me wonder what really goes wrong and how we can prevent people writing their own adapters from hitting similar issues.

The default typeParser from the underlying package produces a `Date` object (see [here](https://node-postgres.com/features/types#date--timestamp--timestamptz)) which we were then trying to store in the Skip heap, clobbering its non-serializable members.

The issue lies in the `as Json` casts in the calls to `callbacks.update` -- those casts are unsafe.  I had assumed that, because it was data coming from the DB it would be serializable, but the `pg` library's conversions from raw postgres data to JS values was introducing some non-serializable stuff like this.

As for how to prevent it... not sure what we could change that wouldn't be similarly vulnerable to incorrect typecasts like this.

bennostein:
> Could we convert them to integers instead? Ordering on string will be wrong

We could convert to unix epoch integers -- that would be better for ordering indeed, but potentially a little more opaque/non-intuitive.  My thinking was that it would be obvious to users what these raw date value strings are (it's the same thing you'd see in e.g. a `psql` repl, and then they can do whatever they want with them in mappers.  And ordering is generally unspecified for collections, right? (outside of the take/slice stuff we were talking through yesterday)

mbouaziz (approved):
Fine, let's worry about ordering later.

I guess we can't really add dynamic checks to detect those cases? Anyway they're limited to what the pg library produces

PR #733, opened 2025-02-06
Simplify docs workflow

jberdine:
This PR adds a git submodule for the docs_site repo and adjusts the docs
workflow to update its contents, removing the need to manually sync the
generated files between repos. The workflow for publishing the docs to the
docs site is now (from www/README.md):
```
- $ make docs-publish
```
which will suggest to run:
```
Test locally: make docs-serve
Push to live site: cd www/docs_site/; git add -A; git commit -m 'update to <commit>'; git push; cd -
```
Once the suggested command ending in `git push` is executed, the live docs
site will update in a few seconds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,4 +22,4 @@ COPY --from=bootstrap /work/compiler/stage0/lib/ /usr/lib/
 FROM skiplang AS skip
 
 RUN --mount=type=bind,source=./bin/apt-install.sh,target=/tmp/apt-install.sh \
-    /tmp/apt-install.sh skipruntime-deps other-CI-tools
+    /tmp/apt-install.sh skipruntime-deps other-CI-tools other-dev-tools
--- a/bin/apt-install.sh
+++ b/bin/apt-install.sh
@@ -54,6 +54,9 @@ for step in "${steps[@]}"; do
             npm install -g prettier
             update-alternatives --auto clang
             ;;
+        other-dev-tools)
+            apt-get install -q -y rsync
+            ;;
         *)
             echo "Unknown step $step"
             usage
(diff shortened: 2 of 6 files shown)


bin/apt-install.sh
```diff
@@ -49,7 +49,7 @@ for step in "${steps[@]}"; do
             ;;
         other-CI-tools)
             # Assumes other steps have been run before
-            apt-get install -q -y clang-format-$LLVM_VERSION parallel pip shellcheck
+            apt-get install -q -y clang-format-$LLVM_VERSION parallel pip rsync shellcheck
```
mbouaziz: This step is used to build docker images for our CI, if `rsync` is not used in the CI, I suggest you put this another step (`other-dev-tools`?)

mbouaziz (approved):
LGTM, I tested all targets but the push itself

Dockerfile
```diff
@@ -22,4 +22,4 @@ COPY --from=bootstrap /work/compiler/stage0/lib/ /usr/lib/
 FROM skiplang AS skip
 
 RUN --mount=type=bind,source=./bin/apt-install.sh,target=/tmp/apt-install.sh \
-    /tmp/apt-install.sh skipruntime-deps other-CI-tools
+    /tmp/apt-install.sh skipruntime-deps other-CI-tools other-dev-tools
```
mbouaziz, line 25: This docker image is used by the CI.

Ah, maybe you use it for development too?

jberdine: Oh, you don't? :-) My understanding was that all the build and whatnot workflow was expecting us to be using these docker images, possibly extended with things for our personal use in `skip/$USER/Dockerfile` where the image build script looks for user-specific stuff.

mbouaziz: I work on Ubuntu, things just work :-D

We can split up the image further if needed later, let's keep moving

PR #735, opened 2025-02-06
Some cleaning in packages

skiplabsdaniel:


skipruntime-ts/wasm/package.json
```diff
@@ -3,7 +3,10 @@
   "version": "0.0.10",
   "type": "module",
   "exports": {
-    ".": "./dist/src/skip-runtime.js"
+    ".": {
+      "node": "./dist/src/node.js",
+      "default": "./dist/src/browser.js"
```
jberdine, line 8: Interesting. FMU, where do these "node" and "default" keys come from? Is it something built in to npm?

skiplang/prelude/ts/wasm/src/sk_types.ts
```diff
@@ -852,15 +844,14 @@ export function isNode() {
...
 
-export async function loadEnv(extensions: EnvInit[], envVals?: string[]) {
-  // hack: this way of importing is deliberate so that web bundlers
-  // don't follow the node dynamic import
-  const nodeImport = "./sk_node.js";
-  const environment = await (isNode()
-    ? import(/* @vite-ignore */ nodeImport)
-    : import("./sk_browser.js"));
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-  const env = environment.environment(envVals) as Environment;
```
jberdine, line 863: So good to clean this up!

jberdine (approved):
LGTM

skipruntime-ts/wasm/package.json
skiplabsdaniel: Exactly, it's something built in `npm`, https://nodejs.org/api/packages.html#conditional-exports

Issue #734, opened 2025-02-06
Update LLVM version


mbouaziz:
Not currently an issue per se but may become blocking at some point.

Currently SkipLang builds with LLVM 15.
LLVM 15 is already not available on Ubuntu 24.04 repository, though using deb packages from older distributions still works for now.

I have reviewed LLVM 16 and 17 toolchain changelogs and detected nothing that should affect us, apart from a change in a few LLVM IR function annotations.
I managed to build with LLVM 16 (not thoroughly tested through CI yet) but with LLVM 17, the build script of the prelude fails with no more details.

jberdine:
I expect we will need some changes due to opaque pointers becoming mandatory.

IIRC there were some issues with pathological compilation times due to inlining going crazy with later versions.

beauby:
@jberdine IIRC the large compile time are already an issue with LLVM 15 (as it is due to the new pass manager in LLVM), but is currently mitigated by passing `-inline-threshold 0` (which may not be ideal, though some inlining is already done by `skc`).

mbouaziz:
Some history and a status update, since this issue predates the bump that actually happened.

## What landed: 15 → 20, in one hop (July 2025)

@jberdine's call in the second comment here was exactly right — opaque pointers were the blocker. That's precisely why the LLVM 17 build failed in the description: pointee types were officially dropped in LLVM 17, and `skc` was still emitting typed pointers.

Three PRs did the work:

- **#946** — *Transition to opaque pointers in LLVM IR* (the structural prerequisite). Not just a compatibility change: it cut `skc`'s IR by 17% (85.8 MB → 71.3 MB, 1.3M → 1.1M lines) and sped up building `skc` by ~10% (167s → 151s).
- **#970** — the smaller textual-IR syntax deltas: `llvm.lifetime.start/end`'s first argument became `i64 immarg`, and the intrinsic attributes were dropped since `argmemonly` → `memory(argmem: readwrite)` and `readnone` → `memory(none)`.
- **#942** — flipped the version everywhere: `LLVM_VERSION` 15 → 20 in `bin/apt-install.sh`, `INSTALL.md`, and the `ensureCompatibleLLVMVersion()` check in `compile.sk`, plus the bootstrap bump.

So the original motivation — LLVM 15 having fallen off the distro — is resolved. The question is whether to do it again.

## Where things stand now

[...]

PR #736, opened 2025-02-06
Allow llvm16

mbouaziz:
Part of my work towards #734

I probably won't go further for the time being.

Before merging, I want to:
- push docker images with LLVM 16 and skiplang built with it
- check all CI jobs are still green


beauby:
Nice. If we start supporting several versions of `LLVM`, we should have CI jobs for the compiler tests using each supported (major) version.

mbouaziz:
We probably don't want to have the burden of maintaining compatibility with several versions, either they are 99% compatible and we only have to worry when someone enters the 1% edge case, or we have to stick to one.

The CI's not happy so I'll probably give up until it becomes a higher priority issue

jberdine:
The skipruntime test is flaky, rerunning it passed.

bennostein:
> The skipruntime test is flaky, rerunning it passed.

I suspect a race like the one Lucas hit -- working on some fixes to make them more robust to timing variability.

mbouaziz:
Shall we merge this? (without the last commit) considering LLVM 15 and 16 seem compatible enough for our use cases.

Or do you think we should allow only one version?

.circleci/base.yml
```diff
@@ -12,7 +12,7 @@ commands:
 jobs:
   fast-checks:
     docker:
-      - image: skiplabs/skip:latest
+      - image: skiplabs/skip:llvm16
```
beauby, line 15: Do we want to tag docker images by llvm version?

skiplang/compiler/src/compile.sk
```diff
@@ -177,15 +177,17 @@ fun runShell(args: Array<String>, verbose: Bool = false): void {
 }
 
 fun ensureCompatibleLLVMVersion(): void {
-  kLLVMVersion = "15.";
+  kAllowedLLVMVersions = Array["15.", "16."];
```
beauby, line 180: Do we want to allow multiple versions? If so, we should clarify support expectations.

.circleci/base.yml
mbouaziz: If we don't plan on officially supporting and/or using more than one version, probably not.
Here it was just for testing

skiplang/compiler/src/compile.sk
mbouaziz: It's clearly the goal of this PR and I updated the doc accordingly.
Though, I agree I forgot to mention that we are (still) testing on LLVM 15 only.

jberdine: We want to make sure that if external users try to use a version different from what we are testing, it fails, right?

mbouaziz:
Closing in favor of what @beauby is currently doing

PR #737, opened 2025-02-06
De-flake skipruntime example expect-tests

bennostein:
Increase buffer times to de-flake the expectation tests on our examples, leaving enough time for servers to spin up fully before clients start hitting them.

This passed 20x in a row locally just now, but circleCI seems more inconsistent in general so testing up there as well.  Marking as draft for now as I expect this may require a couple iterations.

### DIFF
--- a/skipruntime-ts/examples/departures-client.ts
+++ b/skipruntime-ts/examples/departures-client.ts
@@ -14,6 +14,7 @@ const service = new SkipServiceBroker({
 const local_delay = 10;
 const remote_delay = 2000;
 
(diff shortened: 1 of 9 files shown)


bennostein:
Okay, got it to 20x consecutive passes in CircleCI now by increasing some tolerances/sleeps around the start of the examples.  Marking as ready for review now. (will obviously not land the second commit)

jberdine (approved):

mbouaziz (approved):

PR #740, opened 2025-02-10
Remove specialized Mappers and update docs

jberdine:


www/docs/resources.md
```diff
@@ -29,15 +34,16 @@ const service = {
 await runService(service);
 ```
 
-The service has two input collections `"users"` and `"groups"` (populated here with some initial data), no external dependencies (i.e. the service definition does not define the optional `externalServices` field), and one resource: `ActiveFriends`, defined as follows:
```
jberdine, line 32: Not sure I have understood the existing convention for the `"` marks, but they aren't in the source, so I removed them here.

jberdine:
@bennostein I iterated on this a number of times, so the chance that something incoherent slipped through as I am subconsciously used to reading the same thing over and over is high.

bennostein (approved):

bennostein:
> @bennostein I iterated on this a number of times, so the chance that something incoherent slipped through as I am subconsciously used to reading the same thing over and over is high.

Nothing jumped out to me this time!

jberdine:
Thanks! Any comment on the higher-level question of removing the specialized Mappers after seeing the changes in examples? It would be possible to merge the examples and docs changes for today and leave them in the API if you are not convinced.

bennostein:
> Thanks! Any comment on the higher-level question of removing the specialized Mappers after seeing the changes in examples? It would be possible to merge the examples and docs changes for today and leave them in the API if you are not convinced.

I think there are some pretty clear pros/cons: reducing the API surface is great (especially since those specialized mappers each had slightly different and tricky to remember syntax) but there is definitely more boilerplate in some client code now.

Looking at that client code, though, I think that writing out the fully general types/values for things may also help build people's mental models and suggest more creative/complicated use cases.

On balance I'd lean towards landing but open to argument either way

PR #742, opened 2025-02-10
Fix clz undefined behaviour

mbouaziz:
As reported by @beauby and confirmed by @jberdine, compilation with LLVM > 16 fails because of `clz` being called with 0 which is an undefined behaviour.

Here is a fix + a simplification of the computation of the free table slots + a light specification of free table slot functions.

This fixes compilation of the compiler itself with LLVM 17, I leave the rest to the CI.

...why is free table management done in `palloc.c` in 64-bits (not even behind an `#ifdef`?) and in `memory.c` in 32-bits?

### DIFF
--- a/skiplang/prelude/runtime/palloc.c
+++ b/skiplang/prelude/runtime/palloc.c
@@ -587,23 +587,27 @@ int sk_is_static(void* ptr) {
 /* Free table. */
 /*****************************************************************************/
 
+typedef size_t slot_t;
+
 size_t sk_bit_size(size_t size) {
   return (size_t)(sizeof(size_t) * 8 - __builtin_clzl(size - 1));
 }
 
-size_t sk_pow2_size(size_t size) {
+slot_t sk_slot_of_size(size_t size) {
   size = (size + (sizeof(void*) - 1)) & ~(sizeof(void*) - 1);
-  return (1 << sk_bit_size(size));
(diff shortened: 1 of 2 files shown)


skiplang/prelude/runtime/palloc.c
```diff
@@ -587,23 +587,33 @@ int sk_is_static(void* ptr) {
...
+    size = sizeof(void*);
+  }
+  return __builtin_stdc_bit_width(size - 1);
 }
 
-size_t sk_pow2_size(size_t size) {
-  size = (size + (sizeof(void*) - 1)) & ~(sizeof(void*) - 1);
-  return (1 << sk_bit_size(size));
+size_t sk_size_of_slot(slot_t slot) {
+  // Must return a multiple of sizeof(void*) and at least sizeof(void*)
```
beauby, line 608: This is a bit confusing. Maybe explicitly state that `slot` is expected to be greater or equal to `log2(sizeof(void *))`?

beauby (approved):
LGTM, tested it on LLVM 19 and it worked.

beauby:
> ...why is free table management done in palloc.c in 64-bits (not even behind an #ifdef?) and in memory.c in 32-bits?

This is most probably an oversight. Would you mind opening an issue?

PR #743, opened 2025-02-11
Run the example test with native platform

skiplabsdaniel:


skipruntime-ts/examples/database.ts
```diff
@@ -9,6 +9,9 @@ import { runService } from "@skipruntime/server";
 
 import Database from "better-sqlite3";
 
+import { checkOptions } from "./utils.js";
+
+const options = checkOptions(8080, 8081);
```
beauby: This is much less explicit than the previous API. What is the context here?

skipruntime-ts/tests/examples/departures.sh
```diff
@@ -2,12 +2,24 @@
...
+
+trap cleanup EXIT
+
+if [ "$#" -lt 2 ]; then
+    echo "Usage: $0 <stdout target> <stderr target> <platform>"
+fi
+
+if [ "$3" = "native" ]; then
+    echo "Run departures with native platform"
+    LD_LIBRARY_PATH=$(realpath ../../build/skipruntime) node dist/departures.js >/dev/null &
```
bennostein: This line (and similar throughout, excluding `database.sh`) need `-p native` flags, right?  If I'm reading correctly, then they're all still running under Wasm.

skipruntime-ts/examples/database.ts
bennostein: I agree that it's less explicit. The context IIUC is that there's now added command-line arg parsing to support switching platforms.

But it seems very heavyweight to me:  all we really need (and all that it's being used for here) is a single boolean flag `-p native`, signalling to run the native runtime.

If that's the case, we can avoid all of the added utilities and indirection by just looking at `process.argv[1]` in each of the examples, passing `native` and/or `wasm` there from the make targets.

jberdine: Alternately, it might be simplest to check if an environment variable is set.

skiplabsdaniel: OK for environment variable

skipruntime-ts/examples/sum.ts
```diff
@@ -4,6 +4,19 @@ import { ManyToOneMapper } from "@skipruntime/core";
...
+switch (process.env["SKIP_PLATFORM"]) {
+  case "wasm":
+    platform = "wasm";
+    break;
+  case "native":
+    platform = "native";
+    break;
+  default:
+    break;
+}
```
bennostein: Why let the platform be undefined?  This is semantically equivalent to the simpler:

```suggestion
const platform : "wasm" | "native" = process.env["SKIP_PLATFORM"] == "native" ? "native" : "wasm";
```

(since runService defaults to wasm when platform is undefined)

skiplabsdaniel: Just to make typescript happy, `Type 'string' is not assignable to type '"wasm" | "native" | undefined'`

bennostein (approved):
Looks good!  I've got some minor followups but will keep that separate

PR #747, opened 2025-02-11
Reduce testPostgres false failure rate

skiplabsdaniel:
- The last number of postgres timestamp string has not always 6 digits
- Manage correctly session close in case of failure


bennostein (approved):
Nice, finally-blocks across the board (and in some examples too probably) seem like a good move.

PR #748, opened 2025-02-12
Move util Reducers from core to helpers

jberdine:
Their only use is to be reexported by helpers, so move them there.

### DIFF
--- a/skipruntime-ts/tests/src/tests.ts
+++ b/skipruntime-ts/tests/src/tests.ts
@@ -15,7 +15,8 @@ import type {
   ExternalService,
   ServiceInstance,
 } from "@skipruntime/core";
-import { SkipNonUniqueValueError, Count, Sum } from "@skipruntime/core";
+import { SkipNonUniqueValueError } from "@skipruntime/core";
+import { Count, Sum } from "@skipruntime/helpers";
 
 import {
   TimerResource,
--- a/skipruntime-ts/tests/src/tests.ts
+++ b/skipruntime-ts/tests/src/tests.ts
@@ -16,12 +16,12 @@ import type {
   ServiceInstance,
 } from "@skipruntime/core";
 import { SkipNonUniqueValueError } from "@skipruntime/core";
-import { Count, Sum } from "@skipruntime/helpers";
-
 import {
-  TimerResource,
+  Count,
   GenericExternalService,
-} from "@skipruntime/helpers/external.js";
+  Sum,
+  TimerResource,
+} from "@skipruntime/helpers";
 
 import { it as mit, type AsyncFunc } from "mocha";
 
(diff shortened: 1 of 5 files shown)


bennostein (approved):

jberdine:
I'm unsure this is ok wrt js packaging deps, etc., but if this PR only teaches me that is isn't, that's worth it.

skipruntime-ts/tests/src/tests.ts
```diff
@@ -15,7 +15,8 @@ import type {
   ExternalService,
   ServiceInstance,
 } from "@skipruntime/core";
-import { SkipNonUniqueValueError, Count, Sum } from "@skipruntime/core";
+import { SkipNonUniqueValueError } from "@skipruntime/core";
+import { Count, Sum } from "@skipruntime/helpers";
```
jberdine: On this line there is a compilation failure:
```
src/tests.ts:19:28 - error TS2307: Cannot find module '@skipruntime/helpers' or its corresponding type declarations.
```
even though `tests/package.json` has `@skipruntime/helpers` among the `dependencies`. I don't know what is going on.

skiplabsdaniel (approved):

bennostein:
@jberdine The issue was file exports in the helpers package.json -- just pushed a fix, should be good to land if green

jberdine:
> @jberdine The issue was file exports in the helpers package.json -- just pushed a fix, should be good to land if green

Thanks! I don't understand, but will take it. Was it a case of it was wrong and just happened to work before, but my changes were enough to reveal it, or was the change I made itself wrong?

bennostein:
> > @jberdine The issue was file exports in the helpers package.json -- just pushed a fix, should be good to land if green
> 
> Thanks! I don't understand, but will take it. Was it a case of it was wrong and just happened to work before, but my changes were enough to reveal it, or was the change I made itself wrong?

The change to the `./skipruntime-ts/helpers/package.json` build script resulted in a change to the structure of the produced `./dist`, so corresponding change was also needed in the exports.

`tsc` by default transpiles `src/foo.ts` directly to `dist/foo.js` when you have exactly one `src` directory.  But the new build script gives `tsc` two source directories so your resulting `dist` has `dist/src/foo.js`, `dist/skipwasm-std/bar.js`, etc.  Then, when `import` tries to load the module, it doesn't find the entrypoint where it expects and gives the TS2307 you got.  Seems to me there should be some distinction between "you don't have this module" and "this module's entrypoint can't be found" but alas.

for reference, I debugged by: (1) `npm ls` in `skipruntime-ts/tests` to confirm it does have the expected package (2) `require("@skipruntime/helpers")` from a `node` repl in that dir, gave a "file not found" error, chased that and poke around

PR #749, opened 2025-02-12
`@skip-adapter/kafka`: external service wrapper around Kafka data streams

bennostein:
Putting this up as a draft for now in case anyone has comments or interest.

The core functionality is in place to subscribe to a Kafka "topic" as a Skip external resource, but some more work is required, namely:

- [x] Set up CircleCI with a Kafka cluster to test against
- [x] Build out a more realistic example for `./examples`, for use in docs/tutorials/publicity/etc.
- [x] Documentation of public API
- [x] Add (and integrate) support for asynchronous `shutdown` methods on external services

... and possibly other things as they arise.

One notable point in the design space is how we should map Kafka messages to data in skip collections.

Each Kafka message comes to us essentially as a `{key: string, value: string}` pair, so I'm mapping that stream directly to the natural corresponding `EagerCollection<string, string>`.  But, I've seen some examples out there which use "keys" less as unique identifiers and more as "categories" for partitioning messages under the same topic, or even just leave the keys null and just use the value field. (in this corner case we treat that as the key being the string `"null"`)
I could see this leading to issues where basically _all_ messages end up in the same directory/key in SKStore, and end up triggering excessive recomputations downstream of that.  A bridge to cross when we come to it (probably by adding an optional `({key:string, value:string}) => Json` parameter on KafkaExternalService that lets you define your own key structure) but just flagging it here.


jberdine:
Is it expected that people will use this to create a Skip collection of messages that monotonically grows, or to create an "inbox" collection of messages that are not yet processed, or both? I'm wondering about supporting removing messages in the inbox case. That will generally want to operate at the key granularity, which makes me think that leaving keys out of the messages, or having them be something very coarse like "categories", is indeed going to be a problem. It seems plausible that people will end up wanting to generate uids for keys to identify messages within the adapter code.

I'm not sure I understand your proposal to add a `({key:string, value:string}) => Json` function though. In the most general case, wouldn't we end up needing something like a function that takes a message, so `{key: string, value: string}` I guess, and returns some key-value associations for the collection, so `Iterable<[K, V]>` where `K extends Json` and `V extends Json`? Putting that in, with a default that just produces `[key,value]` for `{key, value}`, earlier than later seems like a good idea to me.

bennostein:
> Is it expected that people will use this to create a Skip collection of messages that monotonically grows, or to create an "inbox" collection of messages that are not yet processed, or both? I'm wondering about supporting removing messages in the inbox case. That will generally want to operate at the key granularity, which makes me think that leaving keys out of the messages, or having them be something very coarse like "categories", is indeed going to be a problem. It seems plausible that people will end up wanting to generate uids for keys to identify messages within the adapter code.
> 
> I'm not sure I understand your proposal to add a `({key:string, value:string}) => Json` function though. In the most general case, wouldn't we end up needing something like a function that takes a message, so `{key: string, value: string}` I guess, and returns some key-value associations for the collection, so `Iterable<[K, V]>` where `K extends Json` and `V extends Json`? Putting that in, with a default that just produces `[key,value]` for `{key, value}`, earlier than later seems like a good idea to me.

Yeah, I like that design and agree it's more general! My thinking was to just take a lambda which constructs a `Json` key for a given message, but I didn't think all too deeply about it.  It would be good to also support dropping messages, manipulating values, etc.

bennostein:
Marking this as ready-for-review: example is now complete and working well, and can run with `cd examples/chatroom && docker compose build && docker compose up`.

I'm still writing some docs, and need to drop the in-example copy of the kafka adapter in favor of an NPM-published version once this is ready to go and published.

mbouaziz:
Wow, that looks great! Is there any particular part of the code you want a review of?

bennostein:
Right, sorry, a bit big of a PR to just toss out like that.  I think the most crucial bit is what's under `./skipruntime-ts` -- the rest is an example using the adapter which users may want to copy/paste from, but won't actually be in any of our published packages.  And I'm sure my frontend code is no good in any case :)

skipruntime-ts/adapters/kafka/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "scripts": {
+    "build": "tsc",
+    "clean": "rm -rf dist",
+    "lint": "eslint"
+  },
+  "author": "SkipLabs",
+  "license": "MIT",
+  "description": "Adapter providing reactive access to Apache Kafka from the Skip runtime",
+  "engines": {
+    "node": ">=22.6.0 <23.0.0"
```
jberdine: Is `<23` needed? I have seemed to get away with using 23.9.

examples/chatroom/README.md
```diff
@@ -0,0 +1,15 @@
+# Reactive chat room example
+
+This example is made of three parts: the skip reactive service (in
+`reactive_service/`), the web service (in `web_service/`), and the front-end (in
+`www/`).
+
+It is meant to serve as an example configuration of a web service with a Kafka
+event-store serving as the primary data source-of-truth and a Skip reactive
+service providing reactive computationsand real-time streaming to clients.
```
jberdine:
```suggestion
service providing reactive computations and real-time streaming to clients.
```

examples/chatroom/compose.yml
```diff
@@ -0,0 +1,56 @@
+services:
+  # Reverse proxy
+  haproxy:
+    image: reactive-chatroom/load-balancer
+    build: ./load_balancer
```
jberdine: Would choosing one term between "reverse proxy" and "load balancer" be easier to understand?

examples/chatroom/compose.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  # Web frontend
+  www:
+    image: reactive-chatroom/www
+    build: ./www
+
+  # Kafka event store
+  kafka:
+    image: bitnami/kafka:latest
+    command: sh -c "((sleep 5 && /opt/bitnami/kafka/bin/kafka-topics.sh --bootstrap-server kafka:19092 --create --topic skip-chatroom-messages && /opt/bitnami/kafka/bin/kafka-topics.sh --bootstrap-server kafka:19092 --create --topic skip-chatroom-likes)&) &&sed -i 's/INFO/WARN/g' /opt/bitnami/kafka/config/log4j.properties &&/opt/bitnami/scripts/kafka/run.sh"
```
jberdine:
```suggestion
    command: sh -c "((sleep 5 && /opt/bitnami/kafka/bin/kafka-topics.sh --bootstrap-server kafka:19092 --create --topic skip-chatroom-messages && /opt/bitnami/kafka/bin/kafka-topics.sh --bootstrap-server kafka:19092 --create --topic skip-chatroom-likes)&) && sed -i 's/INFO/WARN/g' /opt/bitnami/kafka/config/log4j.properties && /opt/bitnami/scripts/kafka/run.sh"
```

examples/chatroom/load_balancer/haproxy.cfg
```diff
@@ -0,0 +1,39 @@
+frontend www
+    mode http
+    bind :80
+    bind :443 ssl crt /etc/ssl/selfsigned.crt
+    http-request redirect scheme https unless { ssl_fc }
+
+
+#    # Redirect /streams/... to the Skip streaming backend.
```
jberdine:
```suggestion
    # Redirect /streams/... to the Skip streaming backend.
```

examples/chatroom/load_balancer/haproxy.cfg
```diff
@@ -0,0 +1,39 @@
...
+
+backend stream
+    mode http
+#
+#    # Rewrite public-facing url /streams/... to /v1/streams/... for
+#    # the Skip server.
+    http-request set-path /v1%[path]
+#
+#    # Forward request to the Skip server's streaming endpoint.
+    server stream1 reactive_cache:8080
```
jberdine:
```suggestion

    # Rewrite public-facing url /streams/... to /v1/streams/... for
    # the Skip server.
    http-request set-path /v1%[path]

    # Forward request to the Skip server's streaming endpoint.
    server stream1 reactive_cache:8080
```

examples/chatroom/reactive_service/src/chatroom.service.ts
```diff
@@ -0,0 +1,101 @@
...
+
+class JoinUniqueLikers
+  implements Mapper<number, Message, number, LikedMessage>
+{
+  constructor(private likersByMessage: EagerCollection<number, string>) {}
+  mapEntry(
+    id: number,
+    messages: Values<Message>,
+  ): Iterable<[number, LikedMessage]> {
+    const msg = messages.getUnique() as Message;
```
jberdine: Is `as Message` needed?

examples/chatroom/reactive_service/src/chatroom.service.ts
```diff
@@ -0,0 +1,101 @@
...
+  },
+  (msg) => {
+    const value = JSON.parse(msg.value);
+    return [[-Number(value.id), value]];
+  },
+);
+
+class GroupByMessage implements Mapper<number, Like, number, string> {
+  mapEntry(_id: number, likes: Values<Like>): Iterable<[number, string]> {
+    return likes.toArray().map((like) => [-like.message_id, like.author]);
```
jberdine, line 46: The negation here and a few lines above is done in order to make the default ordering used by `take` give the most recent messages, right? This seems unfortunately subtle.

examples/chatroom/web_service/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "version": "1.0.0",
+  "scripts": {
+    "build": "tsc",
+    "start": "node dist/app.js"
+  },
+  "author": "SkipLabs",
+  "license": "MIT",
+  "description": "",
+  "dependencies": {
+    "@skipruntime/helpers": "0.0.10",
```
jberdine:
```suggestion
    "@skipruntime/helpers": "0.0.12",
```

examples/chatroom/web_service/src/app.ts
```diff
@@ -0,0 +1,92 @@
...
+  brokers: ["kafka:19092"],
+  clientId: "web-backend",
+});
+
+function encode(
+  msg: { author: string } & ({ body: string } | { message_id: number }),
+): { value: string } {
+  // generate numeric IDs by concatenating the current time with 4 digits of noise
+  // so that IDs are temporally ordered and vanishingly unlikely to collide
+  const timestamp = Date.now();
```
jberdine, line 28: `Date.now` is in ms, so with > 1000 messages per second there will necessarily be collisions. Is the temporal order significant? Or just somewhat close is ok? I guess for a chatroom this rate is ok, but wonder if doing something more robust would be better as an example that people might base things off of in other situations where it is less ok.

Perhaps maintain the previous time in a variable to check against and avoid clashes. But at that point, why not just have a counter and assign the messages consecutive ids.

examples/chatroom/web_service/src/app.ts
```diff
@@ -0,0 +1,92 @@
...
+  const id = Math.floor(10_000 * (timestamp + Math.random()));
```
jberdine, line 29: Not sure this is the right parenthesization to achieve 4 digits of noise.

jberdine (commented):
Looking good! First round of comments, I have not gone through the non-example changes yet.

skipruntime-ts/tests/src/tests.ts
```diff
@@ -1215,6 +1275,70 @@ export function initTests(
...
+      );
+      console.warn("To test properly, run the following then retry:");
+      console.warn("\tdocker pull apache/kafka:latest");
+      console.warn(
+        "\tdocker run -d --name=skip-kafka-container -p 9092:9092 apache/kafka",
+      );
+      console.warn(
+        "\tdocker exec -ti skip-kafka-container /opt/kafka/bin/kafka-topics.sh --bootstrap-server localhost:9092 --create --if-not-exists --topic skip-test-topic",
+      );
+      return;
```
jberdine: I guess this should be modified similar to done in #799

skipruntime-ts/adapters/kafka/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+    const consumer = this.client.consumer({
+      ...this.consumerOptions,
+      groupId: instance,
+    });
+    const fromBeginning = params.fromBeginning ?? true;
+    const setup = async () => {
+      await consumer.connect();
+      await consumer.subscribe({ topic, fromBeginning });
+      await consumer.run({
+        // eslint-disable-next-line @typescript-eslint/require-await
```
jberdine: This makes me slightly nervous, but I guess no way around? And I guess that the situation where this would lead to dropped messages is if the system crashes, so nothing better to do anyhow.

jberdine (approved):
LGTM. Nothing looks off, but I didn't dig into the kafka docs to figure out if this is poking it in the right way, so...

examples/chatroom/reactive_service/src/chatroom.service.ts
bennostein: Yes, exactly.  I agree that it's subtle but not sure what a better alternative would be.

examples/chatroom/web_service/src/app.ts
bennostein: it is -- `timestamp` is a positive int, `Math.random` is a float in `[0,1.0)`

examples/chatroom/web_service/src/app.ts
bennostein: This setup means that messages sent at the _exact_ same millisecond will be ordered arbitrarily, and will have non-colliding IDs 99.99% of the time.  Seems robust enough to me for this example -- if a system is dealing with high enough load for that to be an issue, then it's quite likely to have some existing logic to generate temporally-ordered unique identifiers.  And I didn't want to wade into that here.

skipruntime-ts/adapters/kafka/src/index.ts
bennostein: I applied the same transformation (drop `async` keyword, add `return Promise.resolved(undefined)`) that you did on 795 and it lets us drop this lint suppression!

PR #751, opened 2025-02-12
Manage timeout on SkipServiceBroker

skiplabsdaniel:


jberdine (approved):
LGTM. One question: there is also a call to `fetchJSON` in `deleteUUID` without a timeout, should that be changed too? And similarly, `getStreamUUID` uses `fetch` without a timeout.

skiplabsdaniel:
It's true, I missed them.

PR #753, opened 2025-02-12
Changing the order in which we update sub-directories.

pikatchu:
We use to have a scheme that was pretty simple. Update the directories in their order of creation. The scheme works, but the problem is that this order is sub-optimal when the topology of subdir changes.

Imagine the following scenario:

```
creating dir A/ which is empty
creating dir B/ which reads A

A gets updated, and now creates subdirs.
```

From this point on, the subdirs in A will always be updated **after** B. Which means that, more often than not, B will be computed twice.

To fix this, this diff introduces the notion of a TimeStack. The order in which the directories are computed is no longer based on just creation time, but the stack of all the creation time of all the creators.

So if directory A, creates a subdir S, the timeStack will be
```Array[A.creationTime, S.creationTime]```.

This way, we guarantee that the subdirs are always updated before the
next directories at the same level as the parent.

### DIFF
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -159,12 +159,12 @@ fun updateDirtyReaders(
       context.!dirtyReaders = List.Cons(reader, context.dirtyReaders);
     };
     time = if (reader.parentName == reader.childName) {
-      (-child.getTime(), -child.getTime())
+      -child.getTimeStack()
     } else {
-      (child.getTime(), parent.getTime())
+      child.getTimeStack()
     };
     arrow = Arrow(reader.parentName, reader.childName);
-    context.addToUpdate(time, arrow);
+    context.addToUpdate(time, parent.getTime(), arrow);
   };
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: skiplang/prelude/src/skstore/Context.sk [hunk 1]
```diff
@@ -297,6 +297,38 @@ value class Time(value: Int) uses Orderable, Show {
...
+
+  fun negate(): this {
+    !this.values = this.values.map(x -> -x);
+    this
+  }
+
+  fun compare(other: TimeStack): Order {
+    for (i in Range(0, this.values.size())) {
+      if (i >= other.values.size()) return GT();
+      cmp = this.values.compare(other.values);
```
mbouaziz: I don't have enough knowledge of SKStore yet to understand the rest but this is fishy: `cmp` is invariant.

pikatchu: Thanks! good catch!

skiplabsdaniel (approved):

PR #754, opened 2025-02-12
Improve addon build step

### LINKED ISSUE #700: Node addon "symbol lookup error"
Specifying "auto" as the platform and letting `runService` try to load the native addon when it's not present leads to a symbol lookup error like [this](https://app.circleci.com/pipelines/github/SkipLabs/skip/2376/workflows/68fcbfbb-8edd-4380-9167-45806bb569ee/jobs/8415?invite=true#step-103-7983_129)

Discussed with @jberdine and agreed that we should just default to Wasm for now and get the examples/test changes from #697 landed.  (see: [4f80aee2720d57c745813b248a8ea3488bd385bb](https://github.com/SkipLabs/skip/pull/697/commits/4f80aee2720d57c745813b248a8ea3488bd385bb) )

But, when Daniel is back it would be good to discuss and make a proper fix.  I'm not sure exactly what's going on.

skiplabsdaniel:
Use directly node-gyp in build step instead of using full install
With SKIPRUNTIME empty -lskip-runtime-ts ignored during link so SKIPRUNTIME is forced to be a valid directory SKIP_new_Obstack is forced to be a defined symbol just in case.

close #700

### DIFF
--- a/skipruntime-ts/Makefile
+++ b/skipruntime-ts/Makefile
@@ -2,7 +2,9 @@ SHELL := /bin/bash
 
 SKARGO_PROFILE?=release
 NPM_WORKSPACES=$(shell jq --raw-output "[.workspaces[] | select((startswith(\"sql\") or contains(\"examples\")) | not)] | map(\"-w \" + .) | .[]" ../package.json)
+SKIPRUNTIME?=$(realpath $(CURDIR)/..)/build/skipruntime
 
+export SKIPRUNTIME
 
 .PHONY: install
 install:
--- a/.circleci/base.yml
+++ b/.circleci/base.yml
@@ -32,7 +32,7 @@ jobs:
       - setup
       - run:
           name: Typecheck and lint typescript sources
-          command: npm install && npm run build && npm run lint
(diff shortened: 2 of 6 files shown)


skipruntime-ts/Makefile
```diff
@@ -2,7 +2,9 @@ SHELL := /bin/bash
 
 SKARGO_PROFILE?=release
 NPM_WORKSPACES=$(shell jq --raw-output "[.workspaces[] | select((startswith(\"sql\") or contains(\"examples\")) | not)] | map(\"-w \" + .) | .[]" ../package.json)
+SKIPRUNTIME?=$(realpath $(shell dirname $(CURDIR)))/build/skipruntime
```
jberdine: Just curious, but why not:
```suggestion
SKIPRUNTIME?=$(realpath $(CURDIR)/..)/build/skipruntime
```

www/docs_site
jberdine, line 1: Unintended change right?

jberdine (approved):

PR #757, opened 2025-02-13
[skdb] Fix circleci test_diff

skiplabsdaniel:
The problem was due to the fact that `skdb` was called through `skargo`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ test-%:
 
 .PHONY: test-native
 test-native: build/skdb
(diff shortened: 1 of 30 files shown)


mbouaziz:
Weird, and sad, try `--jobs 1`

skiplabsdaniel:
Can fail also with one, probably not only the number of jobs is involved in that.

mbouaziz (approved):
:rocket:

Issue #765, opened 2025-02-21
Transform NonEmptyIterator into a NonEmptySequence


mbouaziz:
Given that the iterator is now (since #276) fully traversed and saved into an `Array` when the second element is accessed, I suggest that it is transformed into a lazy `Sequence`-like structure, i.e.:
- unlike an `Iterator`, we have the guarantee it can be traversed many times
- we still want to be able to access the first element and check whether it is unique without traversing the full iterator
- as a lazy value it still needs to be `mutable` though it should observably look like an immutable value

jberdine:
Are we sure that the change to eagerly drain iterators and perform a couple copies to store the values in arrays is the desired thing to do long-term on the skip side? I see how for the js api wrappers the copies are needed, but that isn't the only use of the skip NonEmptyIterator (they are used in EHandles), and it is not clear to me that it is best to make them so much heavier long-term.

mbouaziz:
From what I read, `NonEmptyIterator`s usage are:
1. use `first` and ignore the rest (usually because the author knows there are no other values), though I consider this a bad pattern
2. use `first` and fail if there are other values, this is a fix of the previous usage
3. use all values
4. map to another `NonEmptyIterator`
5. skipruntime

I want to keep 2 and 4 lazy.

The main problem I currently see with `NonEmptyIterator` is that it sometimes acts like a sequence, e.g. `map` and `clone`, and sometimes like an iterator which traverses the sequence. I want to fix it.

jberdine:
I'm not sure I understand which behaviors you associate with "like a sequence" vs "like an iterator", but agree that no matter what NonEmptyIterator is fishy. My understanding is that there are 2 uses: one as an internal implementation detail for handles in order to support an emptiness test so that in the empty case an optimization to avoid the allocation of a TWriter can be done, and another use in the skipruntime. This second one comes from, AFAIU, the type of e.g. EHandle.map being
```
  fun map<K2: Key, V2: File>(
    typeKey: Key ~> K2,
    type: File ~> V2,
    context: mutable Context,
    dirName: DirName,
    f: (
      mutable Context,
      mutable TWriter<K2, V2>,
      K,
      mutable NonEmptyIterator<V>,
    ) ~> void,
    rangeOpt: ?Array<KeyRange> = None(),
  ): EHandle<K2, V2>
```
[...]

PR #766, opened 2025-02-24
Various Iterator cleanups

mbouaziz:
Preparatory work before #765 and more

### DIFF
--- a/skiplang/compiler/src/skipTypedAstExpand.sk
+++ b/skiplang/compiler/src/skipTypedAstExpand.sk
@@ -171,8 +171,7 @@ fun type__inter(
         Some(elt)
       }
     })
-    .filter(opt -> opt is Some _)
-    .map(opt -> (opt as Some _).value)
+    .filterNone()
     .collect(Array)
 }
 
--- a/skiplang/prelude/src/core/primitives/String.sk
+++ b/skiplang/prelude/src/core/primitives/String.sk
@@ -552,23 +552,23 @@ private fun isWhitespace(c: Char): Bool {
   }
 }
 
-private mutable class StringBytes(
+private class StringBytes(
   private str: String,
   private range: Range,
 ) extends Bytes {
-  readonly fun ptr(): Unsafe.Ptr<UInt8> {
+  fun ptr(): Unsafe.Ptr<UInt8> {
     Unsafe.Ptr<UInt8>(Unsafe.string_ptr(this.str, this.range.start))
   }
 
-  readonly fun get(index: Int): UInt8 {
+  fun get(index: Int): UInt8 {
     String.getByte(this.str, this.range.start + index)
   }
 
-  readonly fun slice(start: Int, end: Int = Int::max): readonly Bytes {
-    mutable StringBytes(this.str, this.range.subrange(start, end))
+  fun slice(start: Int, end: Int = Int::max): Bytes {
+    StringBytes(this.str, this.range.subrange(start, end))
   }
 
-  readonly fun size(): Int {
+  fun size(): Int {
     this.range.size()
   }
 }
--- a/skiplang/prelude/src/skstore/Handle.sk
+++ b/skiplang/prelude/src/skstore/Handle.sk
@@ -45,17 +45,11 @@ mutable base class FileIterator<+T> extends Iterator<T> {
     | Some(v) -> mutable NonEmptyIterator(v, iter)
     }
   }
-
-  readonly fun isEmpty(): Bool;
 }
 
 mutable class EmptyIterator<T>() extends FileIterator<T> {
   mutable fun next(): ?T {
     None()
-  }
-
-  readonly fun isEmpty(): Bool {
-    true
   }
 }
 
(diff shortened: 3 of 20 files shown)


skiplang/prelude/src/core/primitives/String.sk
```diff
@@ -552,23 +552,23 @@ private fun isWhitespace(c: Char): Bool {
   }
 }
 
-private mutable class StringBytes(
```
beauby, line 555: I think we can also make `ArrayBytes` immutable with the current design (separate `Bytes` vs `MutableBytes`).

sql/src/SqlCsv.sk
```diff
@@ -624,12 +624,7 @@ fun applyDiffStrategy(
     isEligibleForTomb = (tick, source) ~>
       tick.value < snapshot.default(0) || source == writer;
 
-    !dir = dir.reset(
-      newRoot,
-      writer,
-      Array[].collect(SortedSet),
-      isEligibleForTomb,
-    );
+    !dir = dir.reset(newRoot, writer, SortedSet[], isEligibleForTomb);
```
beauby, line 627: This works because `SortedSet[]` desugars to `SortedSet::createFromItems(Array[])`, right?

skiplang/prelude/src/skstore/Handle.sk
```diff
@@ -45,18 +45,12 @@ mutable base class FileIterator<+T> extends Iterator<T> {
     | Some(v) -> mutable NonEmptyIterator(v, iter)
     }
   }
-
-  readonly fun isEmpty(): Bool;
 }
 
 mutable class EmptyIterator<T>() extends FileIterator<T> {
```
beauby: Can this class be removed altogether now?

jberdine: It seems to be needed by FileIterator.create.

jberdine:
Is the test failure known to be unrelated?

sql/src/SqlSchemaMigration.sk
```diff
@@ -16,16 +16,13 @@ private fun resetReactiveViews(
...
-          alteredDirs.contains(from.i0.name)
-        )
+  viewsToReset = viewsDir.unsafeGetFileIter().flatMap((x) ->
+    x.i1.filter((file) ->
+      SelectFile::type(file).cselect.from.any((from) ->
+        alteredDirs.contains(from.i0.name)
       )
     )
-    .collect(Array);
+  );
```
mbouaziz: The failure is due to this change, I'm investigating

skiplang/prelude/src/skstore/DeletedDir.sk
```diff
@@ -35,7 +35,7 @@ class DeletedDir{time: Time, dirName: DirName} extends Dir {
   }
 
   fun getIter(mutable Context, Key): mutable Iterator<File> {
-    mutable EmptyIterator()
+    yield break
```
jberdine: To me this seems less clear than before. But anyhow, there are unresolved issues with `yield`, so I would default to avoiding it.

skiplang/compiler/src/skipTypedAstExpand.sk
```diff
@@ -171,8 +171,7 @@ fun type__inter(
         Some(elt)
       }
     })
-    .filter(opt -> opt is Some _)
-    .map(opt -> (opt as Some _).value)
+    .filterNone()
```
jberdine, line 174: :-( This would be good if `filterNone` wasn't misnamed: it is confusing that `filter` keeps the elements that for which the predicate is true while `filterNone` is the opposite keeping the elements that are not `None`.

jberdine (approved):
This LGTM modulo a few comments. My main worry is that is one of the seemingly-redundant instances of `.collect(Array)` is needed, then the others might be too and we just don't manage to trigger it in the test suite. So maybe just leave them all in.

sql/src/SqlSchemaMigration.sk
beauby: Do you have a stack trace?

skiplang/prelude/src/core/primitives/String.sk
mbouaziz: Ah ah, not really `ArrayBytes` work on `readonly Array`, so it has to be `mutable`, though only `readonly` instances are used

skiplang/compiler/src/skipTypedAstExpand.sk
mbouaziz: I agree, be me renaming guest :D

skiplang/prelude/src/skstore/DeletedDir.sk
mbouaziz: I agree. I needed this change to remove `extends Iterator<File>` from `FileIterator`...
I'll create a real `EmptyIterator`, not restricted to `File`s.

sql/src/SqlCsv.sk
mbouaziz: Indeed! (source: `call_from_items` in `skipNaming.sk`)

sql/src/SqlSchemaMigration.sk
mbouaziz: No, from step by step in gdb, at some point we just jump to an unreachable instruction (#769), the stack trace show this function just below `SKIP_unreachableWithExplanation`, so I suspect it's a compilation bug, worth debugging, but that'll be another PR

mbouaziz:
I reverted the remove of `collect(Array)` for (`NonEmpty`)`Iterator`s

PR #767, opened 2025-02-24
EagerDir cleanups

mbouaziz:
A few cleanups while trying to fix the CI on #766...


jberdine (commented):
This LGTM except for the `Get rid of IterHolder` commit, which I'm unsure of. I have not really understood why removing IterHolder helps, and removing it does make the surface language a little less nice.

mbouaziz:
I've removed that commit and I'll resubmit it in another PR to not block this one

jberdine (approved):

PR #771, opened 2025-02-25
Rework ValuesImpl and NonEmptyIterator

mbouaziz:
Iterators are dangerous, they are used to traverse a (materialized or computed-on-the-fly) sequence. But it's Iterator-dependent whether they represent a point in the traversal or the traversal itself.
In some cases, they can be cloned, reused, restarted. In others they can't.
Most of the times, they can't and must be used linearly.
E.g. `it.drop(1).first()` and `_ = it.drop(1); it.first()` may and may not give the same result, depending on the implementation.
`next` and `sizeHint` are an exception though.

This PR moves the materialization of values from `NonEmptyIterator` (Skip) to `ValuesImpl` (JS), then restricts how `NonEmptyIterator` is used. I probably went a bit too far there and I'm OK going back a few steps now that we know there are currently no misuse.


mbouaziz:
Now ready for review.

It looks like some ts/wasm targets are built using `skc`'s stdlib rather than the stdlib on the repo, which led to compilation errors because I changed the interface of `NonEmptyIterator` (it was complaining about missing `toArray` and `next__ForJS`).
Fortunately there is a common denominator... so I don't have to update the docker image now.
If anyone knows how to fix this in general, that'd be great!

PR #773, opened 2025-02-25
Fix multiple iterations over NonEmptyIterator...

mbouaziz:
...in `populateDefinitions`, as revealed by #771

### DIFF
--- a/skiplang/compiler/src/skipExpand.sk
+++ b/skiplang/compiler/src/skipExpand.sk
@@ -332,10 +332,11 @@ fun populateDefinitions(
         name = key match {
         | SKStore.SID(x) -> x
         };
+        defFilesArray = defFiles.collect(Array);
         if (name.startsWith("class:")) {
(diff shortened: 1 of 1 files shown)


skiplang/compiler/src/skipExpand.sk
```diff
@@ -332,10 +332,11 @@ fun populateDefinitions(
         name = key match {
         | SKStore.SID(x) -> x
         };
+        defFilesArray = defFiles.collect(Array);
```
jberdine, line 335: Is `collect(Array)` the best choice here? What about `collect(Vector)`?

jberdine (approved):
LGTM, one question/suggestion.

skiplang/compiler/src/skipExpand.sk
mbouaziz: It doesn't matter for the iteration, but `setArray` expects an array

PR #774, opened 2025-02-26
Skdate cleanup

mbouaziz:
- C++ -> C
- simplify build
- use `cc` instead of `skbuild`

As a result, `skbuild` is not used anymore, should it be killed?


beauby (approved):
LGTM

PR #779, opened 2025-02-27
[skip/server] Add option to disable CORS.

### LINKED ISSUE #778: Consider adding a `--no-cors` mode to the server.
This should help with local dev workflows without reverse proxies.

beauby:
Closes #778.

### DIFF
--- a/skipruntime-ts/server/src/server.ts
+++ b/skipruntime-ts/server/src/server.ts
@@ -4,8 +4,10 @@
  * @packageDocumentation
  */
 
-import { type SkipService, ServiceInstance } from "@skipruntime/core";
+import { type SkipService } from "@skipruntime/core";
 import { controlService, streamingService } from "./rest.js";
+import type { Express, Request, Response, NextFunction } from "express";
+import express from "express";
 
 /**
  * A running Skip server.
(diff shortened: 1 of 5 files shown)


skiplabsdaniel (approved):

skipruntime-ts/server/src/server.ts
```diff
@@ -142,6 +144,9 @@ export async function runService(
       );
     },
   );
+  if (no_cors) {
+    app.use(cors());
```
mbouaziz: Isn't it?
```suggestion
    app.use(cors);
```

skipruntime-ts/server/src/server.ts
```diff
@@ -151,3 +156,15 @@ export async function runService(
     },
   };
 }
+
+function cors(req, res, next) {
```
mbouaziz: Should it be called no_cors instead?
Do you mind adding types?

skipruntime-ts/server/src/server.ts
beauby: Correct, this is currently untested (adding tests as we speak).

bennostein:
merging to include in 0.0.12...

PR #782, opened 2025-02-28
Allow passing extra args to skc from skargo

mbouaziz:
For example `skargo build --skcopt --no-inline --skcopt --no-peephole` will make calls to `skc` start with `--no-inline --no-peephole`.

Inspired by `ocamlc`'s `-ccopt`


beauby:
LGTM on the principle, but you might want to add the extra options in the fingerprinting mix (either by adding `skc_extra_options` to `BuildConfig` or directly in `BuildRunner.fingerprint_for()`) so that successive invocations with changing `--skcopt` args would indeed trigger a rebuild.

mbouaziz:
Thanks! Is it better now?

beauby (approved):

PR #783, opened 2025-02-28
Check TG_OP to determine trigger type in postgres

bennostein:
See [discord thread](https://discord.com/channels/1093901946441703434/1345016769634500698) for context -- triggers were failing to fire on certain updates with null fields, since a row "IS NOT NULL" in postgres only when ALL of its fields are non-null.

### DIFF
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -206,9 +206,9 @@ export class PostgresExternalService implements ExternalService {
             `
 CREATE OR REPLACE FUNCTION %I() RETURNS TRIGGER AS $f$
 BEGIN
-  IF NEW IS NOT NULL THEN
+  IF TG_OP = 'INSERT' OR TG_OP = 'UPDATE' THEN
     PERFORM pg_notify(%L, NEW.%I::text);
-  ELSIF OLD IS NOT NULL THEN
+  ELSE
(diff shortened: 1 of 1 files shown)


skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -206,9 +206,9 @@ export class PostgresExternalService implements ExternalService {
             `
 CREATE OR REPLACE FUNCTION %I() RETURNS TRIGGER AS $f$
 BEGIN
-  IF NEW IS NOT NULL THEN
+  IF TG_OP = 'INSERT' OR TG_OP = 'UPDATE' THEN
```
beauby, line 209: What happens when `TG_OP` is `DELETE` or `TRUNCATE`?

beauby (approved):

skipruntime-ts/adapters/postgres/src/index.ts
bennostein: 'OLD' is set for DELETE triggers and handled in the ELSE branch here; see https://www.postgresql.org/docs/current/plpgsql-trigger.html 

as for TRUNCATE, it won't invoke row-level triggers like this one -- I'll file an issue for now and look into the best way to handle that going forward.  would need another statement-level trigger on the PG side, and some additional logic on the skip side to clear out the collection when that trigger fires.

PR #787, opened 2025-03-02
[skip/helpers] Add `join_one()`/`join_many()` helpers.

beauby:
Example usage for `join_one()`:
```typescript
type User = { name: string; email: string; };
type Post = { title: string; body: string; author_id: number; };
type PostWithAuthor = { title: string; body: string; author: User; };
...
// The following turns `Post`s and `User`s into `PostWithAuthor`s.
join_one(posts, users, {
  on: "author_id",
  name: "author",
})
```

Example usage for `join_many()`:
```typescript
type Upvote = { post_id: number; user_id: number; };
type Post = { title: string; body: string; };
type PostWithUpvotes = {
  title: string;
  body: string;
  upvotes: { user_id: number; }[];
};
...
// The following turns `Post`s and `Upvote`s into `PostWithUpvote`s.
join_many(posts, upvotes, {
  on: "post_id",
  name: "upvotes",
})
```


skipruntime-ts/helpers/src/join.ts
```diff
@@ -0,0 +1,112 @@
...
+    values: Values<VLeft>,
+  ): Iterable<
+    [KLeft, VLeft & Record<JoinedProperty, Omit<VRight, IdProperty>[]>]
+  > {
+    return values.toArray().map((value_left: VLeft) => {
+      const value_right = {
+        [this.joined_property]: this.right.getArray(key) as Omit<
+          VRight,
+          IdProperty
+        >[],
```
mbouaziz:
```suggestion
        [this.joined_property]: this.right.getArray(key),
```

skipruntime-ts/helpers/src/join.ts
```diff
@@ -0,0 +1,112 @@
...
+    key: KLeft,
+    values: Values<VLeft>,
+  ): Iterable<
+    [KLeft, Omit<VLeft, IdProperty> & Record<JoinedProperty, VRight>]
+  > {
+    return values.toArray().map((v: VLeft) => {
+      const { [this.id_property]: key_right, ...value_left } = v;
+      const value_right = {
+        [this.joined_property]: this.right.getUnique(key_right),
+      } as Record<JoinedProperty, VRight>;
```
mbouaziz: I wanted to avoid the `as`, and got this Typescript-happy:
```ts
      const value_right: { [K in JoinedProperty]?: VRight } = {};
      value_right[this.joined_property] = this.right.getUnique(key_right);
```
though the type of `value_out` is `Omit<VLeft, IdProperty> & { [K in JoinedProperty]?: VRight | undefined; }`, TS is not complaining about assigning it to `Omit<VLeft, IdProperty> & Record<JoinedProperty, VRight>` :disappointed: 

I give up

mbouaziz (commented):
LGTM
Like I said, I'm wondering whether we could make the syntax more user-friendly.

I see you didn't need the `const` attribute for type parameters after all?

mbouaziz (approved):

skipruntime-ts/helpers/src/join.ts
beauby: Yes, I'll make the mappers take a single object parameter instead of two strings. Do you see an other possible improvement?

mbouaziz: We'll see with usage?

beauby:
> Like I said, I'm wondering whether we could make the syntax more user-friendly.

Turns out, the type constraint on `EagerCollection.map`, forcing the mapper's constructor parameters to be `...params: Params` with `Params extends DepSafe[]` prevents us from using an object literal to mimic named parameters (since the object literal won't be `DepSafe`). I'll merge it with the initial API (`constructor(other_collection, join_key, resulting_property)`), as it is in `helpers` and not part of the core API, and we can revisit later.

skipruntime-ts/helpers/src/index.ts
```diff
@@ -14,3 +14,4 @@ export {
 export { SkipExternalService } from "./remote.js";
 export { SkipServiceBroker, fetchJSON, type Entrypoint } from "./rest.js";
 export { Count, Max, Min, Sum } from "./utils.js";
+export { join_one, join_many } from "./join.js";
```
jberdine: Are the snake_case haters going to object?

skipruntime-ts/helpers/src/join.ts
```diff
@@ -0,0 +1,156 @@
...
+  ) {}
+
+  mapEntry(
+    key: K,
+    values: Values<VLeft>,
+  ): Iterable<[K, Omit<VLeft, IdProperty> & Record<JoinedProperty, VRight>]> {
+    return values.toArray().map((v: VLeft) => {
+      const { [this.on]: key_right, ...value_left } = v;
+      const value_right = {
+        [this.name]: this.right.getUnique(key_right),
```
jberdine: Is it obvious that this should use getUnique vs mapEntry producing an association for each value of key_right?

skipruntime-ts/helpers/src/join.ts
```diff
@@ -0,0 +1,156 @@
...
+export function join_one<
+  IdProperty extends keyof V1,
+  JoinedProperty extends string,
+  K extends Json,
+  V1 extends JsonObject & { [P in IdProperty]: Json },
+  V2 extends Json,
+>(
+  left: EagerCollection<K, V1>,
+  right: EagerCollection<V1[IdProperty], V2>,
+  options: {
```
jberdine: I don't understand why these 2 args are packed into a record. Just to try to have named arguments, or something else?

skipruntime-ts/helpers/src/join.ts
```diff
@@ -0,0 +1,156 @@
...
+    on: IdProperty;
+    name: JoinedProperty;
```
jberdine: I'm unsure of this `name` name. Is there an established terminology? E.g. it was not clear to me that it had anything to do with the produced result, as opposed to selecting things from the inputs. I don't know, maybe "to"?

bennostein (approved):

skipruntime-ts/helpers/src/join.ts
beauby: Yes, it's just to mimic named arguments. I'm not sure which is best, but the rationale was that two consecutive positional string arguments might be more error prone.

skipruntime-ts/helpers/src/join.ts
beauby: Agreed, `name` is not very helpful.

skipruntime-ts/helpers/src/join.ts
beauby: Not sure I understand. The `getUnique()` comes from the fact that this is the implementation of `JoinOne`. Do you mean that it is not clear whether this should be a left/right/inner join? As it stands, what we have is semantically the equivalent of having a non-nullable foreign key in `left`, rather than a general left join.
It could be helpful to provide a helper for the case of a nullable foreign key in `left` (which would make the output type `Iterable<[K, Omit<VLeft, IdProperty> & Record<JoinedProperty, VRight | null>]>` and replace the `getUnique()` with a `getArray()[0]`.

skipruntime-ts/helpers/src/index.ts
beauby: I'm happy to rename for consistency.

skipruntime-ts/helpers/src/join.ts
jberdine: > The `getUnique()` comes from the fact that this is the implementation of `JoinOne`.

:-) I'm going in the other direction: trying to understand what `JoinOne` is meant to do from this code.

> Do you mean that it is not clear whether this should be a left/right/inner join? As it stands, what we have is semantically the equivalent of having a non-nullable foreign key in `left`, rather than a general left join.

I suppose I'm thinking of this as just a multimap operation, and not in terms of sql, so my first expectation is that e.g.
```
    h ↦ [{α, l:k}]
ₗ⊔ᵣ k ↦ [v₁, v₂,..., vₙ]
 =  h ↦ [{α, r:v₁}, {α, r:v₂},..., {α, r:vₙ}]
```
should hold. SQL joins have not yet become intuitive to me though, so maybe I'm just wrong.

I just don't like `getUnique`.

> It could be helpful to provide a helper for the case of a nullable foreign key in `left` (which would make the output type `Iterable<[K, Omit<VLeft, IdProperty> & Record<JoinedProperty, VRight | null>]>` and replace the `getUnique()` with a `getArray()[0]`.

Although `getArray()[0]` will not be equivalent if there are multiple rather than zero values.

skipruntime-ts/helpers/src/join.ts
bennostein: I think `on`/`name` as the Id/Joined property for _both_ `join_one` and `join_many` is also potentially confusing...

In `join_one`, `on` is a foreign-key pointer from `left` into `right`, with the assumption that it's unique in `right`.

In `join_many`, `on` is a foreign-key pointer from `right` into `left`, with no uniqueness assumption.

The actual semantics seem good/useful for dealing with SQL one-to-many (join_one, where left is the "many" and right is the "one" table) and many-to-many (join_many, where left is one of the two "many" tables and right is the linking table) relationships but the naming/terminology is super tricky/under-specified here.  I wonder if we could borrow that SQL relationship language here instead of talking about left/right/join (with all the connotations those carry)

skipruntime-ts/helpers/src/join.ts
beauby: > > The `getUnique()` comes from the fact that this is the implementation of `JoinOne`.
> 
> :-) I'm going in the other direction: trying to understand what `JoinOne` is meant to do from this code.
>
> > Do you mean that it is not clear whether this should be a left/right/inner join? As it stands, what we have is semantically the equivalent of having a non-nullable foreign key in `left`, rather than a general left join.
> 
> I suppose I'm thinking of this as just a multimap operation, and not in terms of sql, so my first expectation is that e.g.
> 
> ```
>     h ↦ [{α, l:k}]
> ₗ⊔ᵣ k ↦ [v₁, v₂,..., vₙ]
>  =  h ↦ [{α, r:v₁}, {α, r:v₂},..., {α, r:vₙ}]
> ```
> 
> should hold. SQL joins have not yet become intuitive to me though, so maybe I'm just wrong.

Ok I understand now. I think there are two different notions:
- what I had in mind, which basically corresponds to the mundane use case of "I have two collections of objects, the first of which (collection `A`) contains objects with an `b_id` field that corresponds to (exactly) one of the objects from the second one (collection `B`), and I want to build a collection of objects where the `b_id` field (from `A`) is replaced with the actual corresponding object (from `B`).". It implicitly assumes `B` is a simple map (i.e. only one value per key), and that each `b_id` field from `A` actually corresponds to an entry in `B`. It is basically the semantics of an SQL join with a non-null _foreign key_.
- what you mentioned, which makes more sense from an "operations on multimaps" perspective, is a generalization of what I had in mind, where the foreign key constraint is lifted (i.e. the `b_id` field of `A` can correspond to multiple objects of `B`). While I think it makes sense to offer that option to users, I expect most of the time people will want the former (i.e. having more than one value per key in `B` would be the result of a bug, and thus the `getUnique()` in `join_one()` would rightfully warn them that something went wrong). Maybe the solution is to offer a generic way to enforce that a collection is indeed a simple map? In that case, we can use your semantic for `join_one()` and keep offering the right guarantees.
> I just don't like `getUnique`.
> 
> > It could be helpful to provide a helper for the case of a nullable foreign key in `left` (which would make the output type `Iterable<[K, Omit<VLeft, IdProperty> & Record<JoinedProperty, VRight | null>]>` and replace the `getUnique()` with a `getArray()[0]`.
> 
> Although `getArray()[0]` will not be equivalent if there are multiple rather than zero values.

skipruntime-ts/helpers/src/join.ts
bennostein: i.e. maybe `join_one(left,right,{on,name})` could instead be `joinOneToMany({one, many, oneToManyFKey, asCol})` and `join_many(left,right,{on,name})` could instead be `joinManyToMany({many, linking, linkingFKey, asCol})` or something along those lines...

This line of thinking also suggests an optional third collection argument for `joinMany`, corresponding to the other `many` table -- e.g. if you wanted to have an array of user objects instead of just user IDs in the `upvotes` field of the test case here.

beauby: I think `as` is already much better than `name`, yes.


> This line of thinking also suggests an optional third collection argument for `joinMany`, corresponding to the other `many` table -- e.g. if you wanted to have an array of user objects instead of just user IDs in the `upvotes` field of the test case here.

Yes, this could be an optional `through` option, eg.
```
joinManyToMany(posts, users, {
  through: upvotes,
  left_id: "post_id",  // Not sure about the `left_id`/`right_id` names.
  right_id: "user_id",
  as: "upvotes",
})
```
and keep the simple case without a "join table" as:
```
joinManyToMany(posts, upvotes, {
  // `through` and `right_id` are implicitly null/undefined here.
  left_id: "post_id",
  as: "upvotes",
})
```

mbouaziz:

PR #790, opened 2025-03-06
Use resource instance IDs instead of generating postgres client IDs

bennostein:
@jberdine , this should fix the excess querying you were seeing.

Now that we've got framework-managed unique identifiers per-resource-instance, we can just use those as the pg_notify "channel" identifier.


jberdine (approved):
Nice

PR #791, opened 2025-03-06
Validate more Postgres numeric type aliases; extract to a utils file

bennostein:
Picking some low-hanging fruit here with bit-width variants and aliases of things we already handle.

(second commit is NFC, just moving things around)

### DIFF
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -4,150 +4,17 @@
  * @packageDocumentation
  */
 
-import {
-  type Entry,
-  type ExternalService,
-  type Json,
-  SkipError,
-} from "@skipruntime/core";
+import { type Entry, type ExternalService, type Json } from "@skipruntime/core";
 
 import pg from "pg";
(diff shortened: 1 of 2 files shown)


skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -77,15 +101,39 @@ function validateKeyParam(params: Json): {
...
+        )
+          throw new SkipPostgresError(`Invalid ${type} key: ${key}`);
+        return format(
+          "SELECT * FROM %I WHERE %I = " + keyNum.toString() + ";",
+          table,
+          col,
+        );
+      };
+      break;
+    case "SMALLSERIAL":
```
mbouaziz: Tempted to suggest a factorization of all numeric cases...

mbouaziz (approved):

skipruntime-ts/adapters/postgres/src/index.ts
bennostein: twist my arm! 28cbb7da366da27be519b06516d32d6472faf414

PR #792, opened 2025-03-06
[skip] Add reactive sessions.

beauby:
This commit makes a `context.session` object available to resources. It works by implicitly adding a `__sessions` input collection to services, which maps `string`s to arbitrary `Json`, and using the `Session-Id` HTTP header passed to the server's read end points to inject the required session into the context.

Sessions can be manipulated via two new control end points:
- `PUT /v1/sessions/:session_id`,
- `DELETE /v1/sessions/:session_id`.


skipruntime-ts/server/src/rest.ts
```diff
@@ -13,9 +13,15 @@ export function controlService(service: ServiceInstance): express.Express {
 
   // Streaming control API.
   app.post("/v1/streams/:resource", (req, res) => {
+    const session_id = req.header("Session-Id") ?? null;
```
bennostein: Could we document this (and other endpoints which read the `Session-Id` header) in `www/docs/resources.md`?

skipruntime-ts/core/src/api.ts
```diff
@@ -318,6 +318,8 @@ export interface LazyCompute<K extends Json, V extends Json> {
  * Skip Runtime internal state.
  */
 export interface Context extends Managed {
+  readonly session: (Json & DepSafe) | null;
```
bennostein: Exposing the session on the Context seems great, big improvement on exposing the collection directly!

This could use a docstring explaining what it is/how to use it, though.

skipruntime-ts/tests/src/tests.ts
```diff
@@ -759,6 +759,55 @@ const mapWithExceptionOnExternalService: SkipService<Input_SN, Input_SN> = {
...
+
+  constructor(session: Json | null) {
+    this.session = session ? (session as { foo: string }) : null;
+  }
+
+  mapEntry(key: number, values: Values<Post>): Iterable<[number, PostWithFoo]> {
+    return values
+      .toArray()
+      .map((v: Post) => [key, { ...v, foo: this.session?.foo ?? null }]);
+  }
```
bennostein: Would something like this work?  i.e. checking instead of casting to the type of the session, and extracting `foo` in the constructor instead of every map.

```suggestion
  private readonly sessionData: string | null;

  constructor(session: Json | null) {
    this.sessionData = session?.foo ?? null;
  }

  mapEntry(key: number, values: Values<Post>): Iterable<[number, PostWithFoo]> {
    return values
      .toArray()
      .map((v: Post) => [key, { ...v, foo: this.sessionData }]);
  }
```

beauby: Yes, good call.

skipruntime-ts/server/src/rest.ts
```diff
@@ -100,6 +114,31 @@ export function controlService(service: ServiceInstance): express.Express {
...
+    try {
+      service.update("__sessions", [
+        [req.params.session_id, [req.body as Json]],
+      ] as Entry<string, Json>[]);
+      res.sendStatus(200);
+    } catch (e: unknown) {
+      console.log(e);
+      res.status(500).json(e instanceof Error ? e.message : e);
+    }
+  });
```
jberdine: The data associated with each session-id is an opaque json blob as far as skip is concerned, right? But if someone wants to use it as a key-value store, they have to set the whole thing at once. Since there is no way to get the current value to use as the basis of an extension, e.g. to add a new key-value association, then anyone wanting to call PUT needs to themselves keep track of the whole thing. I don't know if that fits the expected uses. Would it be good to add a GET method to enable extending rather than only overwriting the session data?

skipruntime-ts/core/src/api.ts
jberdine: One important aspect is just when it is null vs non-null.

PR #794, opened 2025-03-07
Add hacking section on building against local packages

jberdine:


HACKING.md
```diff
@@ -7,3 +7,33 @@ This document is a perpetually under construction collection of tips and potenti
...
+In some cases, `npm link` can be used to create a symbolic link in the client code's `node_modules` that points to the locally modified skip package.
+If the client runs in docker, however, this approach does not work since the target of the link will not be available in the container.
+
+An alternative is to use `npm pack`. For example:
+```
+> cd path/to/skip
+> mkdir path/to/client/pack
+> npm pack --pack-destination path/to/client/pack -w skipruntime-ts/core
+```
+This creates a tarball of the `skipruntime-ts/core` package in `path/to/client/pack` directory.
```
bennostein:
```suggestion
This creates a tarball of the `@skipruntime/core` package in `path/to/client/pack` directory.
```

bennostein (approved):
Just tried this workflow and it works perfectly -- nice, thanks Josh!

One note is that the current example Dockerfile also needs tweaking to accomodate the `npm pack`'ed package(s), adding a `COPY pack pack` before the `RUN npm install`.  We could also just run the copy-everything `COPY . .` before the npm install to skip that step, but it would mean running the install layer on every rebuild.

PR #795, opened 2025-03-11
Restructure pg adapter `subscribe`; improve error messages.

bennostein:
This PR breaks down the async `setup()` of the postgres adapter's `subscribe` method into three separate phases, provides different and more descriptive error messages for errors thrown in each one, and puts the resource into an `error` state when an exception. is caught and logged.

### DIFF
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -63,9 +63,9 @@ export class PostgresExternalService implements ExternalService {
       },
       (e: unknown) => {
         console.error(
-          "Error connecting to Postgres at " + JSON.stringify(db_config),
+          `Error connecting to Postgres at ${JSON.stringify(db_config)}:`,
         );
-        throw e;
+        console.error(e);
       },
     );
   }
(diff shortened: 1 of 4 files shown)


skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -144,28 +153,38 @@ FOR EACH ROW EXECUTE FUNCTION %I();`,
...
             (changes) => {
               const k = key.type == "TEXT" ? msg.payload! : Number(msg.payload);
-              callbacks.update([[k, changes.rows as Json[]]], false);
-            },
-            (e: unknown) => {
-              console.error(`Error executing Postgres query: ${query}`);
-              throw e;
+              try {
+                callbacks.update([[k, changes.rows as Json[]]], false);
+              } catch (e: unknown) {
```
jberdine: Do you see this actually catching exceptions thrown from mapEntry? I have similar try-catch in uncommitted debugging exploration and do not see it actually catching them.

bennostein: I did in an earlier iteration, but it's been through a couple of those and may no longer be the case... testing/tweaking now to be sure.

jberdine: I will also check with the repro I have been using

jberdine (approved):
LGTM. I tested with the repro I have been using and this does log (but not rethrow) as advertised.

I added a couple commits I needed to get the build working locally.

PR #801, opened 2025-03-13
Skc: option to canonize paths in generated ll

mbouaziz:
The checked in bootstrap `.ll.gz` files contain absolute paths that only make sense for the person building them.
Paths are found in `!DIFile` directives (both absolute CWD `directory` and absolute file path `filename`) and other positions (e.g. those found in messages for unreachableMethodCall calls).
This PR:
- makes `!DIFile`'s `filename` relative to the `directory`, _all the time_, from what I could find, this seems legit
- makes all file names relative to `CWD`, **if** `--canonize-paths` is given

Hence, to compile `skc` with canonized paths, use `skargo build --skcopt --canonize-paths`

### DIFF
--- a/skiplang/compiler/Skargo.toml
+++ b/skiplang/compiler/Skargo.toml
@@ -8,6 +8,8 @@ test-harness = "SkcTests.main"
 std = { path = "../prelude" }
 cli = { path = "../cli" }
 arparser = { path = "../arparser" }
+
+[dev-dependencies]
 sktest = { path = "../sktest" }
 
 [[bin]]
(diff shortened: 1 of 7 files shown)


jberdine (approved):
LGTM though I wonder about the term "anonymize". The main point is to make the reproducible across machines, right? Maybe "canonize".

mbouaziz:
The main point is to not show my username in a publicly available binary.
A side effect is indeed a _canonicalization_.
I can rename the option to `--canonical-paths` (canonize means something else, canonicalize doesn't seem to exist)

jberdine:
> canonize means something else

Maybe it also means something else, but in mathematical logic is commonly used for functions that select unique individuals from a collection.

PR #804, opened 2025-03-13
Check Python formatting in pre-commit hook

mbouaziz:


Makefile
```diff
@@ -95,23 +95,23 @@ clean-all: clean
 	find . -type d -name 'node_modules' -prune -print0 | xargs -0 rm -rf
 
 .PHONY: fmt-sk
-fmt-sk:
+fmt-sk: # Keep in sync with bin/git_hooks/check_format.sh
```
jberdine, line 98: Maybe it would be enough to have this comment just once.

jberdine (approved):

Makefile
mbouaziz: People don't read comments, especially if they're more than 2 lines away from the line they're touching

PR #805, opened 2025-03-13
CodeQL: disable C/C++ and Java/Kotlin

mbouaziz:
I enabled CodeQL for the repo, which can catch bugs on TypeScript and Python.
The default config automatically detects the languages on the repo but it complains it doesn't know how to compile our C/C++ and our Java stuff, so let's disable them for now.

### DIFF
--- a/codeql-config.yml
+++ b/codeql-config.yml
@@ -1,0 +1,9 @@
+disable-default-queries: true
+
+queries:
+  - exclude: 
+      language: cpp
+  - exclude: 
+      language: java
+  - exclude: 
+      language: kotlin


mbouaziz:
Both analyses take between 1 and 2 minutes and are run even when no TS/Py is changed.
Also, we may be billed if our usage exceeds the free allowance for our plan https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions

We may want to reduce the frequency of the analyses, especially if the signal is low.

jberdine:
Do you have experience with this that indicates that it is any good?

mbouaziz:
It's the default analyzer provided by Github, I have no data nor experience about it.

I think I'll switch to the weekly analysis only, and enable per-PR only if it starts finding interesting stuff.

Btw, I checked, our budgets (https://github.com/enterprises/skiplabs/billing) are at $0 so there's no risk in being billed for its usage for now

PR #806, opened 2025-03-14
[skjson] Rename sk_freeze and remove unused export

bennostein:
I noticed while adding an optional default parameter to Values#getUnique that it's probably a footgun to export `sk_freeze` from skjson via `core`'s index.ts, since it doesn't actually _freeze_, just adds the hidden property that we use to track frozen-ness.

This PR drops that export and renames to avoid confusion.

### DIFF
--- a/skiplang/skjson/ts/binding/src/index.ts
+++ b/skiplang/skjson/ts/binding/src/index.ts
@@ -34,7 +34,7 @@ export abstract class Frozen implements Managed {
   protected abstract freeze(): void;
 }
 
-export function sk_freeze<T extends object>(x: T): T & Managed {
+function sk_freeze<T extends object>(x: T): T & Managed {
   return Object.defineProperty(x, sk_managed, {
     enumerable: false,
     writable: false,
--- a/skiplang/skjson/ts/binding/src/index.ts
+++ b/skiplang/skjson/ts/binding/src/index.ts
@@ -34,7 +34,7 @@ export abstract class Frozen implements Managed {
   protected abstract freeze(): void;
 }
 
-function sk_freeze<T extends object>(x: T): T & Managed {
+function tagSkManaged<T extends object>(x: T): T & Managed {
   return Object.defineProperty(x, sk_managed, {
     enumerable: false,
     writable: false,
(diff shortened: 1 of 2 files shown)


skiplang/skjson/ts/binding/src/index.ts
```diff
@@ -34,7 +34,7 @@ export abstract class Frozen implements Managed {
   protected abstract freeze(): void;
 }
 
-export function sk_freeze<T extends object>(x: T): T & Managed {
+function withManagedTag<T extends object>(x: T): T & Managed {
```
jberdine: My first impression of the `withManagedTag` name is that it would return a fresh object like the argument but with a tag, but this function mutates the argument. So maybe `tagSkManaged`?

skipruntime-ts/core/src/index.ts
```diff
@@ -54,7 +54,7 @@ import {
   type FromBinding,
 } from "./binding.js";
 
-export { sk_freeze, isSkManaged };
+export { deepFreeze, isSkManaged };
```
jberdine: deepFreeze is already exported to the public api from api.ts

jberdine (approved):
Agree that the `sk_freeze` name is not ideal. Made a suggestion for a maybe better one.

Note though that core/src/index.ts is the internal interface of the core package, the one we expect users to use is only the part in api.ts.

bennostein:
> Agree that the `sk_freeze` name is not ideal. Made a suggestion for a maybe better one.

Suggestion taken!
 
> Note though that core/src/index.ts is the internal interface of the core package, the one we expect users to use is only the part in api.ts.

Oh right, of course!  So I guess this is just taking the footgun away from _us_, not users.  Renamed the PR and updated the description, thanks Josh

jberdine:
> > Note though that core/src/index.ts is the internal interface of the core package, the one we expect users to use is only the part in api.ts.
> 
> Oh right, of course! So I guess this is just taking the footgun away from _us_, not users. Renamed the PR and updated the description, thanks Josh

Well, I don't know that there is anything that actually prevents users from importing this stuff, it's just not documented.

PR #808, opened 2025-03-14
Add optional "default" arg to `getUnique` methods

bennostein:
This allows to replace patterns like

```
let foo;
try {
  foo = bar.getUnique(key);
} catch {
  foo = defaultVal;
}
```

with the imo-much-nicer 

```
const foo = bar.getUnique(key, { ifNone: defaultVal });
```

`getUnique` on collections (which can have 0, 1, or many values per key) has separate and optional  `ifNone` and `ifMany` defaults, while `getUnique` on `Values` (which can have 1 or many values) has only the `ifMany` optional default.

### DIFF
--- a/skipruntime-ts/core/src/api.ts
+++ b/skipruntime-ts/core/src/api.ts
@@ -85,14 +85,16 @@ export interface Reducer<V extends Json, A extends Json> {
 }
 
 /**
- * An iterable collection of dependency-safe values.
+ * A non-empty iterable sequence of dependency-safe values.
  */
 export interface Values<T> extends Iterable<T & DepSafe> {
   /**
    * Return the first value, if there is exactly one.
-   * @throws {@link SkipNonUniqueValueError} if this iterable contains either zero or multiple values.
+   * @param _default
+   * @param _default.ifMany - Default value to use instead of throwing if there are multiple values.
+   * @throws {@link SkipNonUniqueValueError} if this iterable contains multiple values.
    */
-  getUnique(): T & DepSafe;
+  getUnique(_default?: { ifMany?: T }): T & DepSafe;
 
   /**
    * Return all the values.
@@ -124,10 +126,13 @@ export interface LazyCollection<K extends Json, V extends Json>
    * For collections that do not use the generality of associating multiple values to a key, `getUnique` saves some boilerplate over `getArray`.
    *
    * @param key - The key to query.
+   * @param _default
+   * @param _default.ifNone - Default value for the case where **zero** values are associated to the given key
+   * @param _default.ifMany - Default value for the case where **multiple** values are associated to the given key
    * @returns The value associated to `key`.
-   * @throws {@link SkipNonUniqueValueError} if `key` is associated to either zero or multiple values.
+   * @throws {@link SkipNonUniqueValueError} if `key` is associated to either zero or multiple values and no suitable default is provided.
    */
-  getUnique(key: K): V & DepSafe;
+  getUnique(key: K, _default?: { ifNone?: V; ifMany?: V }): V & DepSafe;
 }
 
 /**
@@ -154,10 +159,13 @@ export interface EagerCollection<K extends Json, V extends Json>
    * For collections that do not use the generality of associating multiple values to a key, `getUnique` saves some boilerplate over `getArray`.
    *
    * @param key - The key to query.
+   * @param _default
+   * @param _default.ifNone - Default value for the case where **zero** values are associated to the given key
+   * @param _default.ifMany - Default value for the case where **multiple** values are associated to the given key
    * @returns The value associated to `key`.
-   * @throws {@link SkipNonUniqueValueError} if `key` is associated to either zero or multiple values.
+   * @throws {@link SkipNonUniqueValueError} if `key` is associated to either zero or multiple values and no suitable default is provided.
    */
-  getUnique(key: K): V & DepSafe;
+  getUnique(key: K, _default?: { ifNone?: V; ifMany?: V }): V & DepSafe;
 
   /**
    * Create a new eager collection by mapping a function over the values in this one.
--- a/examples/hackernews/reactive_service/src/hackernews.service.ts
+++ b/examples/hackernews/reactive_service/src/hackernews.service.ts
@@ -21,6 +21,7 @@ type User = {
   name: string;
   email: string;
 };
+const unknownUser: User = { name: "unknown author", email: "unknown email" };
 
 type Upvote = {
   post_id: number;
@@ -66,12 +67,9 @@ class PostsMapper {
   ): Iterable<[[number, number], PostWithUpvoteIds]> {
     const post: Post = values.getUnique();
     const upvotes: number[] = this.upvotes.getArray(key);
-    let author;
-    try {
-      author = this.users.getUnique(post.author_id);
-    } catch {
-      author = { name: "unknown author", email: "unknown email" };
-    }
+    const author = this.users.getUnique(post.author_id, {
+      ifNone: unknownUser,
+    });
     return [[[-upvotes.length, key], { ...post, upvotes, author }]];
   }
 }
@@ -128,12 +126,9 @@ class PostsResource implements Resource<PostsResourceInputs> {
   instantiate(
     collections: PostsResourceInputs,
   ): EagerCollection<number, PostWithUpvoteCount> {
-    let session;
-    try {
-      session = collections.sessions.getUnique(this.session_id);
-    } catch {
-      session = null;
-    }
+    const session = collections.sessions.getUnique(this.session_id, {
+      ifNone: null,
+    });
     return collections.postsWithUpvotes
       .take(this.limit)
       .map(CleanupMapper, session);
--- a/examples/hackernews/reactive_service/src/hackernews.service.ts
+++ b/examples/hackernews/reactive_service/src/hackernews.service.ts
@@ -67,9 +67,12 @@ class PostsMapper {
   ): Iterable<[[number, number], PostWithUpvoteIds]> {
     const post: Post = values.getUnique();
     const upvotes: number[] = this.upvotes.getArray(key);
-    const author = this.users.getUnique(post.author_id, {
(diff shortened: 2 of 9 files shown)


bennostein:
Note the 4th commit here undoes the changes to use the new API in the hackernews example -- keeping that in to show more usages for code review, and to easily revert upon the next NPM release

jberdine:
Since, as we have seen, dealing with exceptions from mappers is not simple, and will never be super convenient, I wonder if we should push the API in the direction to avoid exceptions as much as possible. The fact that mappers will be run async and from writes, etc. will mean that robust designs will have to encode error conditions into data, not use control mechanisms like exceptions.

So I wonder if it would be better to go further than this PR and make the default value non-optional, and also use it when there are multiple values.

mbouaziz:
If `getUnique` returns more than one value, it probably means an invariant is broken, do we really want to recover from that?
Otherwise it means you should have used `getArray` instead, isn't it?

skipruntime-ts/core/src/api.ts
```diff
@@ -89,9 +89,11 @@ export interface Reducer<V extends Json, A extends Json> {
 export interface Values<T> extends Iterable<T & DepSafe> {
   /**
    * Return the first value, if there is exactly one.
+   * @param options - Optional named parameters
+   * @param options.default - (optional) A default value to return if this is empty
```
mbouaziz: A `NonEmptyIterator` cannot be empty. `getUnique` will actually only fail if there are more than one value. Isn't it?

bennostein:
> If `getUnique` returns more than one value, it probably means an invariant is broken, do we really want to recover from that? Otherwise it means you should have used `getArray` instead, isn't it?

No, I don't think so -- see the examples changed here for some cases where it comes up.  Messing around with users' code and working on our own examples, I've definitely run into cases where it's useful.  E.g. in the hackernews example, you want to get the author data for some post (with author ID); if no author data is available, you want some default value.  Of course you _could_ do this with conditionals on `getArray(...).length` or try-catch, but it's nicer to just pass a default, and it comes up fairly often. Analogous to `Option.value ~default` in OCaml

bennostein:
> Since, as we have seen, dealing with exceptions from mappers is not simple, and will never be super convenient, I wonder if we should push the API in the direction to avoid exceptions as much as possible. The fact that mappers will be run async and from writes, etc. will mean that robust designs will have to encode error conditions into data, not use control mechanisms like exceptions.
> 
> So I wonder if it would be better to go further than this PR and make the default value non-optional, and also use it when there are multiple values.

Interesting, not opposed to it! I'll try it out that way and see how it works out in examples/tests -- I suspect there'll be cases where it makes things less ergonomic, but we'll see

jberdine:
> I suspect there'll be cases where it makes things less ergonomic, but we'll see

I would expect so as well. But I'm thinking that since exceptions are not going to propagate easily, it might well be the right pressure toward robust user-code to apply.

skipruntime-ts/core/src/api.ts
```diff
@@ -84,14 +84,16 @@ export interface Reducer<V extends Json, A extends Json> {
...
 export interface Values<T> extends Iterable<T & DepSafe> {
   /**
    * Return the first value, if there is exactly one.
-   * @throws {@link SkipNonUniqueValueError} if this iterable contains either zero or multiple values.
+   * @param options - Optional named parameters
+   * @param options.default - (optional) A default value to return if this contains multiple values.
+   * @throws {@link SkipNonUniqueValueError} if this iterable contains multiple values and no default is provided.
    */
-  getUnique(): T & DepSafe;
+  getUnique(options?: { default: T }): T & DepSafe;
```
jberdine:
```suggestion
  getUnique(options?: { default: T & DepSafe }): T & DepSafe;
```
Here and elsewhere, it is probably better to require users to deepFreeze their own objects. The rest of the API requires the user to provided already-frozen values where needed rather than the framework mutating them on the users' behalf.

bennostein: I did it that way at first, but the syntax gets very clunky very fast... especially if we're also making defaults non-optional

bennostein: e.g. from one of our `./skipruntime-ts/examples`:
```diff
   mapEntry(gid: GroupID, group: Values<Group>): Iterable<[GroupID, UserID]> {
     return group
-      .getUnique()
+      .getUnique({
+        default: deepFreeze({ name: "Unknown Group", members: [] }),
+      })
       .members.flatMap((uid) =>
-        this.users.getUnique(uid).active ? [[gid, uid]] : [],
+        this.users.getUnique(uid, {
+          default: deepFreeze({ name: "Unknown User", friends: [] }),
+        }).active
+          ? [[gid, uid]]
+          : [],
       );
   }
 }
 ```

jberdine: Not pushing, but hmm: Even without deepFreeze, the default objects ought ideally to be let-bound before the return statement (as a field of the Mapper class even), no? Otherwise they have to be allocated each time the closure given to flatMap is called.

bennostein: Yes, that's what I actually have in the code (also since they are needed ~5 times throughout that example).  Copied like this to make it self-contained. Still working on pushing required defaults through all of the examples, I think it's a mixed bag so far.

mbouaziz:
> > If `getUnique` returns more than one value, it probably means an invariant is broken, do we really want to recover from that? Otherwise it means you should have used `getArray` instead, isn't it?
> 
> No, I don't think so -- see the examples changed here for some cases where it comes up.

I had `ValuesImpl` in mind. A default makes sense for collections' `getUnique`, not for `ValuesImpl`.

bennostein:
> > > If `getUnique` returns more than one value, it probably means an invariant is broken, do we really want to recover from that? Otherwise it means you should have used `getArray` instead, isn't it?
> > 
> > 
> > No, I don't think so -- see the examples changed here for some cases where it comes up.
> 
> I had `ValuesImpl` in mind. A default makes sense for collections' `getUnique`, not for `ValuesImpl`.

Ahh, your comment makes sense then.  I see your point, but it's somewhat in conflict with the drive from @jberdine to make our `getUnique` APIs non-throwing and force users to provide defaults and encode errors into their values and data model.

So I think there are two main options in play:

 * keep `Values.getUnique` as-is, and add _optional_ default parameters on `{Eager,Lazy}Collection#getUnique`.  This gives a nice shorthand for the cases we want to support, but also lets exceptions get thrown if users mess up.
 * Require defaults on all three `getUnique` variants, such that they'll never throw.

The middleground of requiring defaults on collections' getUnique while not supporting defaults on Values' getUnique doesn't make much sense to me, since there's not much point to ruling out exceptions on collections if we're going to throw them on values.

My feeling is that the first option is probably the way to go: I don't love requiring a default on collections, since the syntactic overhead is fairly high, and there are cases where it's preferable to write try/catch conditional logic rather than encode the error into your types, return it as a default, and check for that default value.

jberdine:
I am a bit hesitant about trying to go for the non-throwing world. There will be other sources of exceptions and we need to do something to not just leave the system in a broken or at least unknown state when that happens.

On the other hand, I think it would be confusing to have optional default values for only some of the error conditions. That is, if getUnique returns the default if there are zero values but throws if there are multiple, then that is going to be confusing.

I also don't understand the motivation to treat getUnique differently whether it is used to look up a key in a collection vs map being given the values for a key. It is just find vs map for the same underlying structure that has only one set of invariants. Am I overlooking something?

bennostein:
Okay, I think we're roughly on the same page then.

The first version of this PR failed to update a couple docstrings which I think was confusing, but the current state is what I think best balances the constraints here:

 * `Values.getUnique` does _not_ support a default: it's guaranteed to have at least one member, and if it has multiple then you should either throw or catch and explicitly handle.
 * `{Eager,Lazy}Collection.getUnique` accepts an _optional_ default argument; `collection.getUnique(key, {default: x})` returns the value associated to `key` if there is exactly one such value and `deepFreeze(x)` otherwise.

> I also don't understand the motivation to treat getUnique differently whether it is used to look up a key in a collection vs map being given the values for a key. It is just find vs map for the same underlying structure that has only one set of invariants. Am I overlooking something?

On this point: I think the key difference is that `Values` are _always_ nonempty. So, if you've set things up such that your collection has one value per key, then finding multiple values when you `getUnique` is an invariant violation that should throw; if not, then you should be iterating over the values and doing something that properly handles multiple values.

For the collections' `getUnique`, it's convenient to provide a default specifically for lookups on non-existent keys.  And, like you say, it would be confusing if the default was used for the zero-values case but not the multiple-values case.  If that's an important distinction, people can always use `getArray` and explicitly case-split on the result's length.

jberdine:
Ok, but if getUnique is going to treat errors in 2 different ways, then I think that at least the name of the optional argument should indicate which error condition is covers. So maybe `ifNone` or `ifMissing` instead of `default`?

bennostein:
Summarizing some offline conversation w/ Josh:

 - This feels to me like a local maximum in the design space.  Some doubts from Josh still on whether Values should also support a default, and whether Collections should distinguish the zero-value and multiple-values cases in getUnique
  - Its worth considering whether we should just make the API more multimap-first, by some combination of (1) dropping the Values interface and just passing mappers a materialized array whose length they can case-split on if need be, (2) dropping collections' `getUnique` and renaming `getArray` to `get`.  This would have some benefits in simplifying the API and pushing peoples' mental model closer to the actual skip data model of multimaps. But it would be slightly less convenient (probably? not certain) for the special case of single-value-per-key maps, and would make it easier to accidentally drop values past the first on the floor during a map.

jberdine:
Circling back after offline discussion, my feeling is that adding e.g. `ifNone` and `ifMany` optional arguments to `getUnique`, which would be used in the 2 distinct error cases, would enable the desired simplification at call sites without adding to confusion between the functional map and multimap cases. And for the version of `getUnique` where emptiness is statically impossible, it could take only the `ifMany` optional argument, avoiding confusion with the other `getUnique`s. And they would all still throw if the optional args are not present. WDYT?

PR #810, opened 2025-03-17
Cleanups in dirs and handles

mbouaziz:
Collection of various cleanups in `SKStore`, mainly:
- improving guarantees on classes by making some fields/functions/constructors private
- removing dead stuff

### DIFF
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -748,8 +748,7 @@ mutable class Context{
           | _ -> void
           };
         };
-        !dir.data = data;
-        this.setDir(dirName, dir)
+        this.setDir(dirName, dir.withData(data))
       }
     };
   }
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -380,7 +380,7 @@ mutable base class WriteChecker private {
 mutable class Context{
   mutable toPurge: SortedSet<DirName> = SortedSet[],
   mutable purgeLimit: ?Tick = None(),
-  mutable dirs: Dirs = Dirs{},
+  private mutable dirs: Dirs = Dirs{},
   mutable reads: SortedSet<Path> = SortedSet[],
   mutable time: Time = Time(0),
   mutable tick: Tick = Tick(1),
@@ -1226,7 +1226,7 @@ mutable class Context{
 // @cpp_extern("SKIP_print_backtrace")
 // native fun printBacktrace(): void;
 
-class Dirs{state: DMap<DirName, Dir> = DMap::empty()} {
+private class Dirs{private state: DMap<DirName, Dir> = DMap::empty()} {
   fun set<T: File>(tick: Tick, key: DirName, value: Dir): this {
     !this.state = this.state.set(tick, key, value);
     this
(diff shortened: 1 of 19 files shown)


sql/src/Main.sk
```diff
@@ -573,7 +573,7 @@ fun execDiff(args: Cli.ParseResults, options: SKDB.Options): void {
 
     for (dirSub in sub.dirSubs) {
       tableName = P.Name::create(
-        dirSub.dirName.dirName.stripPrefix("/").stripSuffix("/"),
+        dirSub.dirName.toString().stripPrefix("/").stripSuffix("/"),
```
beauby, line 576: Do we expect `dirName.dirName` and `dirName.toString()` to be semantically equivalent?

skiplang/prelude/src/skstore/EagerFilter.sk
```diff
@@ -93,12 +91,12 @@ class EagerFilter private {
   private static fun mkDir(
     context: mutable Context,
     childName: DirName,
-    suffix: String,
+    suffix: String = "",
```
beauby: Would it make sense to use `suffix: ?String` here?

sql/src/Main.sk
mbouaziz: That's a good point.
As said offline:
- using `dirName.toString()` is already a bad pattern
- it's probably covered by tests
- and only in `sql/`
Let's not worry?

skiplang/prelude/src/skstore/Context.sk
```diff
@@ -748,8 +748,7 @@ mutable class Context{
           | _ -> void
           };
         };
-        !dir.data = data;
```
beauby, line 751: Is `dir.data` certain not to be read from after this?

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -984,11 +984,8 @@ class Collection(hdl: SKStore.EHandle<JSONID, JSONFile>) {
       ),
       None(),
       Array[
-        SKStore.DirSub(
-          this.hdl.getDirName(),
-          "",
-          SKStore.OJSON(Array[]),
-          (__) ~> None(),
+        SKStore.DirSub(this.hdl.dirName, "", SKStore.OJSON(Array[]), (__) ~>
```
beauby, line 987: Out of curiosity, any reason why `(__) ~>` uses a double underscore as param name?

skiplang/prelude/src/skstore/EagerFilter.sk
mbouaziz: Why not. `mkDir` is only ever used twice, privately, on lines 249, 250.

skiplang/prelude/src/skstore/Context.sk
mbouaziz: `dir` is local to this loop

beauby (approved):

skipruntime-ts/native/src/Runtime.sk
mbouaziz: Seems to have been introduced in 618f0adf1fec83f9646a60c69b5f8a81ddb5b2ce, ask @skiplabsdaniel

skiplang/prelude/src/skstore/EagerFilter.sk
mbouaziz: Done

PR #815, opened 2025-03-19
[skip-adapter/postgres] Add option to only receive new rows in postgres adapter

bennostein:
As discussed, this could be useful for append-only tables where a full view of the history is not needed.

Some small unit testing refactors as well.

I don't love the `ignoreOldData` option name, but it was the best I could come up with... open to other suggestions

### DIFF
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -25,7 +25,7 @@ pg.types.setTypeParser(pg.types.builtins.TIMESTAMPTZ, (x: string) => x);
  *
  * @remarks
  * Subscription `params` **must** include a field `key` whose value is an object with a string field `col` identifying the table column that should be used as the key in the resulting collection, and a field `type` whose value is a Postgres text, integer, or serial type. (i.e. one of `TEXT`, `SERIAL`, `SERIAL2`, `SERIAL4`, `SERIAL8`, `BIGSERIAL`, `SMALLSERIAL`, `INTEGER`, `INT`, `INT2`, `INT4`, `INT8`, `BIGINT`, or `SMALLINT`)
- * Subscription `params` **may** also specify `ignoreOldData: true` to receive only _new_ rows as they are created, for use with append-only tables whose history is not needed.
+ * Subscription `params` **may** also specify `syncHistoricData: false` to receive only _new_ rows as they are created, for use with append-only tables whose history is not needed.
  */
 export class PostgresExternalService implements ExternalService {
   private client: pg.Client;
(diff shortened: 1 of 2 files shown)


skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -77,7 +78,9 @@ export class PostgresExternalService implements ExternalService {
...
-   * @param params - Parameters of the external resource; **must** include a field `key` whose value is an object with a string field `col` identifying the table column that should be used as the key in the resulting collection, and a field `type` whose value is a Postgres text, integer, or serial type. (i.e. one of `TEXT`, `SERIAL`, `SERIAL2`, `SERIAL4`, `SERIAL8`, `BIGSERIAL`, `SMALLSERIAL`, `INTEGER`, `INT`, `INT2`, `INT4`, `INT8`, `BIGINT`, or `SMALLINT`)
+   * @param params - Parameters of the external resource; **must** include a field `key` describing the Postgres column that should be used as the key in the resulting collection
+   * @param params.key - (Required) Object with a field `col` identifying the table column to be used as the key, and a field `type` whose value is a Postgres text, integer, or serial type. (i.e. one of `TEXT`, `SERIAL`, `SERIAL2`, `SERIAL4`, `SERIAL8`, `BIGSERIAL`, `SMALLSERIAL`, `INTEGER`, `INT`, `INT2`, `INT4`, `INT8`, `BIGINT`, or `SMALLINT`)
+   * @param params.ignoreOldData - (Optional) Boolean flag: if true, Skip will ignore pre-existing data and only synchronize updates after subscription
```
jberdine: Maybe`syncHistoricData` that defaults to true?

skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -77,7 +78,9 @@ export class PostgresExternalService implements ExternalService {
    *
    * @param instance - Instance identifier of the external resource.
    * @param resource - Name of the PostgreSQL table to expose as a resource.
-   * @param params - Parameters of the external resource; **must** include a field `key` whose value is an object with a string field `col` identifying the table column that should be used as the key in the resulting collection, and a field `type` whose value is a Postgres text, integer, or serial type. (i.e. one of `TEXT`, `SERIAL`, `SERIAL2`, `SERIAL4`, `SERIAL8`, `BIGSERIAL`, `SMALLSERIAL`, `INTEGER`, `INT`, `INT2`, `INT4`, `INT8`, `BIGINT`, or `SMALLINT`)
+   * @param params - Parameters of the external resource; **must** include a field `key` describing the Postgres column that should be used as the key in the resulting collection
```
jberdine: Isn't everything following the `;` now redundant with the `params.key` entry below?

skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -108,15 +112,21 @@ export class PostgresExternalService implements ExternalService {
...
-      for (const row of init.rows as { [col: string]: Json }[]) {
-        const k = row[key.col]!;
-        if (entries.has(k)) entries.get(k)!.push(row);
-        else entries.set(k, [row]);
+      if (params.ignoreOldData) {
+        callbacks.update([], true);
+      } else {
+        callbacks.loading();
+        const init = await this.client.query(
+          format("SELECT * FROM %I;", table),
```
jberdine, line 120: I wonder if there is a small bool->int generalization possible here, where instead of syncing all or none of the historic data, there is a number that could be put as a limit on this query. Not sure it is better, but maybe.

jberdine (approved):

skipruntime-ts/adapters/postgres/src/index.ts
bennostein: Yep -- I have a diff to follow with more docusaurus changes, too.

skipruntime-ts/adapters/postgres/src/index.ts
bennostein: Certainly possible, but I'm not sure when that would be semantically what you want -- i.e. you'd pull (up to) `N` historic rows initially, but then still get unboundedly many future rows? So it wouldn't limit the memory footprint to `N` rows worth in perpetuity, nor would it operate as a "streaming" source where history doesn't factor in.

Would be straightforward to add if a use case arose, though!

PR #818, opened 2025-03-20
[www/docs] Update docs to match current status quo; add docstring for TimerResource

bennostein:


www/docs/externals.md
```diff
@@ -2,15 +2,23 @@
...
 Skip provides mechanisms to do so easily, whether those external systems are other reactive Skip services or non-reactive systems like databases or external APIs.
 
 The core of these mechanisms is the [`ExternalService` type](api/core/interfaces/ExternalService), which provides a generic interface which can be implemented to wrap arbitrary external systems for use in your Skip service.
 Each Skip reactive service can specify any number of `ExternalService`s, which can then be brought into the reactive computation graph as an `EagerCollection` with [`Context#useExternalResource`](api/core/interfaces/Context#useexternalresource).
 
-Skip provides two `ExternalService` implementations -- one which polls external HTTP services, and one which subscribes to external Skip services.
+Skip provides several `ExternalService` implementations:
+
+ * [`SkipExternalService`](api/helpers/classes/SkipExternalService), which is used to connect reactive Skip services together.
+ * [`PostgresExternalService`](api/adapters/postgres/classes/PostgresExternalService), which allows to reactively subscribe to updates from a PostgreSQL database.
```
jberdine:
```suggestion
 * [`PostgresExternalService`](api/adapters/postgres/classes/PostgresExternalService), which allows to subscribe to reactive updates from a PostgreSQL database.
```

www/docs/externals.md
```diff
@@ -2,15 +2,23 @@
...
+ * [`KafkaExternalService`](api/adapters/kafka/classes/KafkaExternalService), which allows to subscribe to and consume messages from a Kafka cluster.
```
jberdine: (Without knowing kafka) "subscribe to" and "consume" seem redundant? And this sort of subscription is also reactive like postgres, right?

www/docs/externals.md
```diff
@@ -58,9 +66,73 @@ Encapsulating external reactive dependencies as eager collections, that complex
 ## Non-Skip services
 
 Of course, unless your application is built from the ground up using the Skip framework, it is likely that your application depends on some non-reactive external system: REST APIs, databases, external HTTP endpoints, and the like.
-These systems operate on a pull-based request/response paradigm, so some work is required to adapt them to Skip's eager push-based paradigm.
+
+### PostgreSQL
+
+One common use case for Skip is to reactively update and push results in response to updates in a source-of-truth relational database.
```
jberdine: Is "source-of-truth" relevant here? The db might just as well not be the source of truth from our perspective, as long as the SoT is not us.

www/docs/externals.md
```diff
@@ -58,9 +66,73 @@ Encapsulating external reactive dependencies as eager collections, that complex
 ## Non-Skip services
 
 Of course, unless your application is built from the ground up using the Skip framework, it is likely that your application depends on some non-reactive external system: REST APIs, databases, external HTTP endpoints, and the like.
-These systems operate on a pull-based request/response paradigm, so some work is required to adapt them to Skip's eager push-based paradigm.
+
```
jberdine, line 69: I guess that right here we could add a sentence stating that the following pre-built integrations can be reimplemented with whatever customization is needed for a given use case; nothing beyond the exposed skip framework is used in the implementations.

jberdine (approved):

www/docs/externals.md
bennostein: Kafka very heavily uses "producer"/"consumer" terminology, whereas "subscribe" is more our language -- I agree they feel a bit redundant here.  Maybe a better phrasing would be "connect to and consumer messages from ...".

And yes, it's also reactive.

PR #819, opened 2025-03-20
[skipruntime] Simplify helper interfaces for externals

bennostein:
This PR drops the `GenericExternalService` instance of the (actually generic) `ExternalService` interface, along with the `ExternalResource` interface that it used.  It drops the `TimerResource` instance of that interface, and promotes the `Polled` one to a `PolledExternalService` proper.

This is a simplification and cleanup of the interfaces/class hierarchy but should otherwise mostly preserve the behavior and user interface.


skipruntime-ts/helpers/src/external.ts
```diff
@@ -139,75 +24,100 @@ export function defaultParamEncoder(params: Json): string {
...
+   */
+  interval: number;
+  /**
+   * Function to convert data received from external resource to `key`-`value` entries.
+   */
+  conv: (data: Json) => Entry<Json, Json>[];
+  /**
+   * Function to use to encode params of type `Json` for external resource request.
+   *
+   * Note that the result of `encodeParams` may contain `?` separator, but it need not be at the beginning of the returned string, so some parameters can be used in part of the URL preceding the `?`.
```
jberdine:
```suggestion
   * Note that the result of `encodeParams` may contain a `?` separator, but it need not be at the beginning of the returned string, so some parameters can be used in part of the URL preceding the `?`.
```

skipruntime-ts/core/src/api.ts
```diff
@@ -407,7 +407,7 @@ export type CollectionUpdate<K extends Json, V extends Json> = {
  * If it _is_ a Skip service, then `SkipExternalService` can be used, specifying an entrypoint.
```
jberdine, line 407: (I wanted to comment on 2 lines above but github says fu)

I think it would be better to have a sentence that emphasizes that user-defined ExternalServices are expected. Maybe something like the following:

> Custom implementations of the `ExternalService` interface can be defined as needed, and there are two implementations provided by the @skipruntime/helpers package: `SkipExternalService` and `PolledExternalService`.

jberdine (approved):
Nice! You can probably tell that I struggled to understand the real necessity of GenericExternalService when writing the api docs, so nice to trim it out.

PR #820, opened 2025-03-21
Transactions

skiplabsdaniel:


skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -171,35 +168,26 @@ FOR EACH ROW EXECUTE FUNCTION %I();`,
...
+      ) {
+        const query = key.select(table, msg.payload);
+        const run = async () => {
+          const changes = await this.client.query(query);
+          const k = key.type == "TEXT" ? msg.payload! : Number(msg.payload);
+          await callbacks.update([[k, changes.rows as Json[]]], false);
+        };
+        run().catch((e: unknown) => {
+          error(`Error executing Postgres query "${query}":`)(e);
+        });
```
bennostein: I think this will incorrectly report errors executing queries when the query succeeds but `callbacks.update` throws because of some mapper downstream.

Maybe something like the following?

```suggestion
        const k = key.type == "TEXT" ? msg.payload! : Number(msg.payload);
        const run = async () => {
          const changes = await this.client.query(query).catch(error(`Error executing Postgres query "${query}":`))
          await callbacks.update([[k, changes.rows as Json[]]], false);
        };
        run().catch(error(`Uncaught error triggered by Postgres update (key ${k}, table ${table}):`));
```

skipruntime-ts/adapters/postgres/src/index.ts
```diff
@@ -171,35 +168,34 @@ FOR EACH ROW EXECUTE FUNCTION %I();`,
...
+        };
+        run().catch((e: unknown) => {
+          if (ok) {
+            error(
+              `Uncaught error triggered by Postgres adapter update (key ${k}, table ${table}):`,
+            )(e);
+          } else {
+            error(`Error executing Postgres query "${query}":`)(e);
+          }
+        });
```
jberdine, line 195: FWIW I don't understand the motivation for why to do the error reporting this way as opposed to adding a `.catch` to each of the `await`s in `run`.

skipruntime-ts/core/src/index.ts
```diff
@@ -439,70 +457,10 @@ export class ServiceInstanceFactory {
 }
 
 export type GetResult<T> = {
```
jberdine: Isn't this type dead after this PR?

skipruntime-ts/core/src/index.ts
```diff
@@ -1036,11 +995,22 @@ export class ToBinding {
     const supplier = this.handles.get(sksupplier);
     const writer = new CollectionWriter(writerId, this.refs());
     const params = skjson.importJSON(skparams, true) as Json;
-    supplier.subscribe(instance, resource, params, {
-      update: writer.update.bind(writer),
-      error: writer.error.bind(writer),
-      loading: writer.loading.bind(writer),
-    });
+    // Ensure to notification is made outside the curent context update
```
jberdine:
```suggestion
    // Ensure notification is made outside the current context update
```

skipruntime-ts/native/src/Runtime.sk
```diff
@@ -1439,24 +1377,56 @@ fun put(
...
+    !keys = keys.remove(key);
+  });
+  fadded = toAdd.toArray();
+  fremoved = keys;
+  commit = (ctx) ~> {
+    fadded.each(added -> chdl.writeArray(ctx, added.i0, added.i1));
+    fremoved.each(key -> chdl.writeArray(ctx, key, Array[]));
+  };
+  commit(transaction.context);
+  // Do not check carbage collection to prevent closing resources
```
jberdine:
```suggestion
  // Do not check garbage collection to prevent closing resources
```

skipruntime-ts/wasm/package.json
```diff
@@ -9,7 +9,7 @@
     }
   },
   "scripts": {
-    "build": "node ../../prebuild.mjs skipwasm-std skipwasm-json && tsc && skargo build -r --target wasm32-unknown-unknown --lib --manifest-path=../native/Skargo.toml --out-dir=./dist/src/",
+    "build": "skargo clean -r --target wasm32-unknown-unknown --manifest-path=../native/Skargo.toml && node ../../prebuild.mjs skipwasm-std skipwasm-json && tsc && skargo build -r --target wasm32-unknown-unknown --lib --manifest-path=../native/Skargo.toml --out-dir=./dist/src/",
```
jberdine: Was this `clean` a workaround for an issue that remains, or does it make sense that it is needed?

jberdine (approved):

skipruntime-ts/adapters/postgres/src/index.ts
skiplabsdaniel: Because a `.catch ` on the the first await consume the exception and let the code continue, so the update is done even if the query fails.
It's possible to add try catch to the first and rethrow, but I found it a bit heavy. The ideal would have been to sort the errors at the time of the catch, but there doesn't seem to be an error type specific to the Postgres module.

skipruntime-ts/core/src/index.ts
skiplabsdaniel: It's still in use, but it no longer needs to be exported.

PR #824, opened 2025-03-24
Add `CHANGELOG.md`.

beauby:
- Add a `CHANGELOG.md` file, following the format described on `keepachangelog.com`.
- Tagged (pre)-releases from v0.0.10 to v0.0.13.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,53 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.0.13] - 2025-03-21
+
+### Added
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,53 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
bennostein, line 6: Are we strictly-speaking following semver at this point?  I thought the plan was to commit starting from 1.0.0, or maybe from 0.1.0 -- recent "patch" versions certainly haven't been backwards compatible.

bennostein (approved):

CHANGELOG.md
beauby: We are indeed, according so [semver.org](https://semver.org/spec/v2.0.0.html#spec-item-4):
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

PR #835, opened 2025-03-31
Skip service distributed leader/follower example

bennostein:
This PR changes the hackernews example to use 1 leader (talking to postgres) and 3 followers (mirroring postgres data from the leader; serving streams to clients).

It adds one small feature to the skipruntime library code: prefixing stream UUIDs by an environment variable `SKIP_RESOURCE_PREFIX` if defined.

This (along with network aliases) lets all of the follower load balancing be handled by the docker compose roundrobin DNS resolver; HAProxy then uses that prefix to forward streaming requests to the proper follower.  This could be made more modular (so as to support elastic restart-less scaling) but seems a good starting point for design/prototyping work.

If we land this, I'd probably propose keeping separate `./examples/hackernews` and `./examples/hackernews-distributed`, but this PR overwrites the non-distributed `./examples/hackernews`.


skipruntime-ts/server/src/rest.ts
```diff
@@ -16,7 +16,9 @@ export function controlService(service: ServiceInstance): express.Express {
     try {
       const uuid = crypto.randomUUID();
       service.instantiateResource(uuid, req.params.resource, req.body as Json);
-      res.status(201).send(uuid);
+      const prefix = process.env["SKIP_RESOURCE_PREFIX"];
+      console.log(`${prefix}/${uuid}`);
```
beauby: Is this log meant to stay?

examples/hackernews/reverse_proxy/haproxy.cfg
```diff
@@ -4,35 +4,39 @@ frontend www
...
-    use_backend stream if url_streams
-
-    # Redirect /api/... to the api backend.
+    # Dispatch /streams/$follower/$uuid to $follower
+    acl url_follower1 path_beg -i /streams/1/
+    acl url_follower2 path_beg -i /streams/2/
+    acl url_follower3 path_beg -i /streams/3/
+    use_backend follower1 if url_follower1
+    use_backend follower2 if url_follower2
+    use_backend follower3 if url_follower3
```
beauby: Would something like that work (provided the env var given to the followers match the follower names)?
```suggestion
    # Dispatch /streams/$follower/$uuid to $follower
    use_backend %[path_beg,field(2, '/')]
    default_backend follower1
```

skipruntime-ts/server/src/rest.ts
bennostein: Oops, nope, thanks

examples/hackernews/reverse_proxy/haproxy.cfg
bennostein: I made it this way to avoid leaking backend names -- trying it out that way now to see if it works for people who don't care about leaking that bit of data.

I think a better endstate is using Haproxy maps to map external-facing identifiers (`1`, `2`, `3` here) to actual backend names.  That would have the additional benefit of being elastic using the runtime API's set-map/del-map operations.

beauby: Ok makes sense, and agreed on the HAProxy maps down the line.

One thing to consider for this PR: we probably want to make it a separate example (while sharing the web app and client), so that we still have an example for a non-load balanced deployment.

examples/hackernews/skip_leader/src/hackernews.service.ts
```diff
@@ -0,0 +1,130 @@
...
+      service: "postgres",
+      identifier: "upvotes",
+      params: serialIDKey,
+    });
+    return {
+      postsWithUpvotes: posts.map(
+        PostsMapper,
+        users,
+        upvotes.map(UpvotesMapper),
+      ),
```
jberdine: One thing I wonder is whether it would be a better example of suggested usage to defer this map to the follower. The idea being that the leader is a single node while the followers are multiple nodes, so it is most scalable to do any work in the followers that can be. It does duplicate the work done by this mapper though, so not entirely clear. WDYT?

examples/hackernews/web_service/app.py
```diff
@@ -4,8 +4,12 @@
 import requests
 import random
 import time
+import itertools
 
-REACTIVE_SERVICE_URL = "http://reactive_cache:8081/v1"
+SKIP_LEADER_URL = "http://skip_leader:8081/v1"
+
+# round-robin sequence of followers
```
jberdine: Is there anything in this file that implies that the scheduling is round-robin? It looks to me like the load balancing of all the requests to the SKIP_FOLLOWER_URL is the responsibility of a layer above this, e.g. docker compose here.

examples/hackernews/reverse_proxy/haproxy.cfg
jberdine: Just for my education, can you explain how Lucas' suggestion leaks more info?

bennostein: I agree (see last sentence of PR description) but hadn't thought of sharing the web app and client, good call!

examples/hackernews/web_service/app.py
bennostein: Good call, a1859b4b should've dropped this comment

examples/hackernews/skip_leader/src/hackernews.service.ts
```diff
@@ -0,0 +1,130 @@
...
+type Upvote = {
+  post_id: number;
+  user_id: number;
+};
+
+type PostWithUpvoteIds = Post & { upvotes: number[]; author: User };
+
+type Session = User & {
+  user_id: number;
+};
```
jberdine: Maybe just me being paranoid, but I would move these type definitions out into a separate file, and use that file also for the follower, to make sure that the type definitions stay in sync between them.

examples/hackernews/reverse_proxy/haproxy.cfg
bennostein: The suggested `use_backend` incantation requires that the external-facing path looks like `/streams/<backend_name>/...`, i.e. includes the actual name of the backend. (`follower1`, `follower2`, etc. here)

This version has one layer of indirection, with an external-facing identifier (`1`, `2`, `3`) that doesn't need to match the actual haproxy backend name.  Probably a minor concern in the scheme of things.

examples/hackernews/skip_leader/src/hackernews.service.ts
bennostein: Hmm, I'm not sure, do see the argument both ways depending on actual workload.  Was not a considered decision here -- I just manually split the one-machine logic along the same lines that it was already separated into createGraph/resources.  Let's see and discuss when we get to the JS helpers for running as leader/follower/etc. maybe?

examples/hackernews/skip_leader/src/hackernews.service.ts
bennostein: Hmm, easier said than done with docker/typescript -- can't just import or symlink it from outside their respective root dirs without breaking things.  Maybe that's a data point in favor of "single service definition, multiple runners" but in any case I'm going to set it aside for now :) 

edit: nvm, didn't set it aside, see: https://github.com/bennostein/skdb/tree/asLeader-asFollower

bennostein:
Landing now so that further changes can be reviewed without too much pain... feel free to comment/@-tag me though with any comments on what's here!

www/docs/deploying.md
```diff
@@ -54,10 +54,27 @@ The `acl` and `use_backend` lines in the frontend configuration forward any traf
...
+## Horizontal scaling
+
+For Skip services with sustained or bursty high-throughput requirements, Skip reactive services can scale horizontally to serve many clients without overloading a single machine or incurring heavy resource costs.
+
+The most common bottleneck for reactive systems is *memory*, due to the heavy use of caching and materialization of intermediate results.
+In a Skip service, much of this memory pressure comes from instantiating large numbers of resources to stream customized/personalized data to many clients.
+To handle that scenario, you can run your Skip service in a *leader-follower* topology, running:
+
+    * one *leader* instance to maintain the shared computation graph and/or pull in data from external dependencies like databases or polled APIs, and
+    * one or more *followers* which synchronize that shared data from the leader, among which resource instances are distributed in a round-robin fashion.
```
jberdine, line 74: AFAIU "round-robin" is just an implementation detail of the particular example, not inherent in any way, right?

PR #837, opened 2025-04-03
Add blog post about external services / adapters

bennostein:
Here's a quick blog post about non-reactive dependencies and our new Postgres/Kafka adapters.  Putting up as a PR for high-level and/or copy-edit feedback!

### DIFF
--- a/www/blog/adapters.md
+++ b/www/blog/adapters.md
@@ -36,18 +36,25 @@ const service = {
 }
 ```
 
-Then, anywhere in that service, you can pull in a database table as a Skip collection by specifying a table name and a column to use as the key of the collection, e.g. as follows for a table with schema `CREATE TABLE t (id SERIAL PRIMARY KEY, value TEXT)`.
+Then, anywhere in that service, you can pull in a database table as a Skip collection by specifying a table name and a column to use as the key of the collection.
+For example, given a PostgreSQL table `my_table` with schema
+
+```sql
+CREATE TABLE my_table (id SERIAL PRIMARY KEY, value TEXT);
+```
+
+we can use it as a collection `myTable` in a Skip reactive service as follows.
 
 ```typescript
-const t: EagerCollection<number, { id: number, value: string }> =
+const myTable: EagerCollection<number, { id: number, value: string }> =
   context.useExternalResource({
     service: "postgres",
-    identifier: "t",
+    identifier: "my_table",
     params: { key: { col: "id", type: "SERIAL" } },
   });
 ```
 
-Under the hood, Skip will maintain this collection up-to-date by watching the database for changes to `t`; within your Skip service, you can define arbitrary reactive computations mapping over and accessing `t`, which will update incrementally the same as any other Skip collection.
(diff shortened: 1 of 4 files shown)


www/blog/adapters.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Skip services with non-reactive dependencies
```
jberdine, line 2: Maybe worth emphasizing:
```suggestion
title: Skip services with non-reactive external dependencies
```

www/blog/adapters.md
```diff
@@ -0,0 +1,129 @@
...
+  resources: ...,
+  createGraph: ...,
+  externalServices: {
+    postgres: new PostgresExternalService({ host, port, database, ... }),
+    ...,
+  },
+}
+```
+
+Then, anywhere in that service, you can pull in a database table as a Skip collection by specifying a table name and a column to use as the key of the collection, e.g. as follows for a table with schema `CREATE TABLE t (id SERIAL PRIMARY KEY, value TEXT)`.
```
jberdine: The SQL code is probably long enough that it would be better as a triple-tick md block

jberdine: s/t/tbl or my-table or something might help readability

www/blog/adapters.md
```diff
@@ -0,0 +1,129 @@
...
+```typescript
+const t: EagerCollection<number, { id: number, value: string }> =
+  context.useExternalResource({
+    service: "postgres",
+    identifier: "t",
+    params: { key: { col: "id", type: "SERIAL" } },
+  });
+```
+
+Under the hood, Skip will maintain this collection up-to-date by watching the database for changes to `t`; within your Skip service, you can define arbitrary reactive computations mapping over and accessing `t`, which will update incrementally the same as any other Skip collection.
```
jberdine: Maybe worth calling out that reactive computations that combine multiple tables, or combine tables with any other skip collection, are also possible.

www/blog/adapters.md
```diff
@@ -0,0 +1,129 @@
...
+    }),
+  },
+};
+```
+
+This [`PolledExternalService`](https://skiplabs.io/docs/api/helpers/classes/PolledExternalService) specifies a single endpoint polled every 5 seconds, but in general can include any number of endpoints with different polling intervals and converter/parser functions.
+
+Of course, polling necessarily introduces some latency in the reactive service: data can become stale between polls.
+On the other hand, increasing the request frequency can place undue load on the target non-reactive API.
+As such, care should be taken when setting up a polled dependency to choose a reasonable interval, and where possible polling should be avoided in favor of proper reactive updates.
```
jberdine: Maybe avoid "proper":
```suggestion
As such, care should be taken when setting up a polled dependency to choose a reasonable interval, and where possible polling should be avoided in favor of genuinely reactive updates.
```

jberdine (approved):
LGTM!

bennostein:
Thanks! Took all of the suggestions except for the post title -- it's already fairly long and doesn't render too nicely in browser if made even longer

www/blog/adapters.md
```diff
@@ -0,0 +1,136 @@
+---
+title: Skip services with non-reactive dependencies
+description: External integrations for the Skip framework
+slug: non-reactive-dependencies
+date: 2025-04-04
```
jberdine: We'll have to remember to update this.

PR #838, opened 2025-04-03
[skc] Introduce `#env()` macro.

beauby:
The `#env()` macro takes a single string literal as parameter, and
expands to the value of the corresponding environment variable at
compile time. It plays a somewhat similar role as the `-D` option of
gcc in that it allows injecting values into the program from the
command line.

This will make it easier to handle things such as embedding the git
commit hash of a given build into programs and displaying it with a
`--version` option.

Note: The environment is frozen when building an `sklib`, i.e. a call
to `#env("foo")` within a package `pkg` will always expand to the
value of the `foo` environment variable at the time of compilation of
`pkg`, even when `libpkg.sklib` is used as a dependency of an other
program/library.

### DIFF
--- a/skiplang/compiler/src/SkipParseTree.sk
+++ b/skiplang/compiler/src/SkipParseTree.sk
@@ -1649,6 +1649,58 @@ class LogMacroTree{
   }
 }
 
+class EnvMacroTree{
+  envKeyword: ParseTree.ParseTree,
+  openParen: ParseTree.ParseTree,
+  varName: ParseTree.ParseTree,
+  closeParen: ParseTree.ParseTree,
+} extends ParseTree.ParseTree {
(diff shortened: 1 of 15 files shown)


jberdine (approved):
This seems ok for now, but I don't think that we want this sort of thing long-term. In order to achieve reproducible builds, the sort of build metadata that this would be used to inject ought to be (generated and) stored in files that are checked in and part of source archives.

beauby:
> This seems ok for now, but I don't think that we want this sort of thing long-term. In order to achieve reproducible builds, the sort of build metadata that this would be used to inject ought to be (generated and) stored in files that are checked in and part of source archives.

I somewhat agree, but it's a trade off: having for instance the git HEAD commit hash in a checked in file is not doable. For things such as the package version, having it set both in `Skargo.toml` and in a generated file would easily get out of sync, etc.

PR #853, opened 2025-04-09
[skipruntime/helpers] asLeader/asFollower utilities for distributed skip services

bennostein:
This PR cherry-picks the non-example changes from #835, which includes example usages and more documentation of these utilities.
I would like to land this PR and cut a release, so that one can get merged without `npm pack` pulling non-released changes into the docker-compose examples.

The further improvements/followups that are on my radar but I think are non-blocking are:
 * using HAProxy maps to scale out dynamically
 * running web-service -> skip-service traffic through the load balancer to put round-robin logic there instead of docker's DNS resolver
 * adding a mechanism to specify which externals to instantiate in leaders and followers
 * putting together an example using kubernetes to spin up/orchestrate services instead of docker compose

### DIFF
--- a/skipruntime-ts/server/package.json
+++ b/skipruntime-ts/server/package.json
@@ -20,7 +20,6 @@
   },
   "dependencies": {
     "@skipruntime/core": "0.0.13",
-    "@skipruntime/helpers": "0.0.13",
     "express": "^4.21.1"
   },
   "optionalDependencies": {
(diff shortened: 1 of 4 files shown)


skipruntime-ts/server/package.json
```diff
@@ -20,6 +20,7 @@
   },
   "dependencies": {
     "@skipruntime/core": "0.0.13",
+    "@skipruntime/helpers": "0.0.13",
```
beauby: Is this strictly required? Would it make sense to actually add `asLeader()`/`asFollower()` to `@skipruntime/helpers` instead?

bennostein: Yeah, makes sense to me, thanks!  Moved them there.

beauby (approved):

PR #855, opened 2025-04-09
Fix shallow clone for PRs from clone

mbouaziz:
Now let's try from a clone


mbouaziz:
Looks like both (#854 and this) work

PR #859, opened 2025-04-10
[skipruntime/server] Add an optional env variable prefix to stream UUIDs

bennostein:
This is cherry-picked from #835; needs to be in public release for that example to work without npm-pack.

Should've included in 0.0.14, alas.  Will cut an 0.0.15 including this.

### DIFF
--- a/skipruntime-ts/server/src/rest.ts
+++ b/skipruntime-ts/server/src/rest.ts
@@ -16,7 +16,8 @@ export function controlService(service: ServiceInstance): express.Express {
     try {
       const uuid = crypto.randomUUID();
       service.instantiateResource(uuid, req.params.resource, req.body as Json);
-      res.status(201).send(uuid);
+      const prefix = process.env["SKIP_RESOURCE_PREFIX"];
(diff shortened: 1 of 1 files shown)


jberdine (approved):

skipruntime-ts/server/src/rest.ts
```diff
@@ -16,7 +16,8 @@ export function controlService(service: ServiceInstance): express.Express {
     try {
       const uuid = crypto.randomUUID();
       service.instantiateResource(uuid, req.params.resource, req.body as Json);
-      res.status(201).send(uuid);
+      const prefix = process.env["SKIP_RESOURCE_PREFIX"];
+      res.status(201).send(prefix ? `${prefix}/${uuid}` : uuid);
```
beauby, line 20: We might as well get rid of the `/` and leave full control of the prefix to the user, what do you think?

bennostein: That's more flexibility for the user to be sure, but this makes the url-splitting logic in the reverse proxy dead simple. I guess I'd be slightly worried about people forgetting a delimiter or choosing a bad one and otherwise messing that up.

PR #862, opened 2025-04-11
NPM version 0.0.15

bennostein:
My bad with the quick back-to-back releases here, this just pulls in #859 (and I've confirmed all good with the npm-only build packing everything at this SHA)


bennostein:
Ahh, transactions landed while I was putting this together.  Updating/testing with that now; @skiplabsdaniel is it alright with you to release immediately?

skiplabsdaniel:
It's good for me.

PR #865, opened 2025-04-11
[hackernews example] Pull round-robin logic from Docker compose to HAproxy

bennostein:
This PR hands control of load-balancing back to ... the load-balancer, routing web-service -> skip traffic through the reverse proxy.  That also means the web service now need not care whether the skip service is running in "distributed mode" or not.

@beauby I think this is more or less what you had in mind -- what do you think?  One thing I didn't bother with yet that is probably important in practice is locking down the `:8081` listeners in the haproxy configs to not accept any external IPs.

### DIFF
--- a/examples/hackernews/reverse_proxy/haproxy.distributed.cfg
+++ b/examples/hackernews/reverse_proxy/haproxy.distributed.cfg
@@ -4,9 +4,10 @@ frontend www
     bind :443 ssl crt /etc/ssl/selfsigned.crt
     http-request redirect scheme https unless { ssl_fc }
 
-    # Dispatch /streams/$follower/$uuid to $follower
+    # Dispatch /streams/$follower/$uuid to streaming backend, stashing $follower in a header
     acl url_streams path_beg -i /streams/
-    use_backend %[path,field(3,/)] if url_streams
+    http-request set-header Follower-Prefix %[path,field(3,/)] if url_streams
+    use_backend stream if url_streams
 
     # Dispatch /api/... to the api backend
     acl url_api path_beg -i /api/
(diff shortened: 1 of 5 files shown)


examples/hackernews/reverse_proxy/haproxy.distributed.cfg
```diff
@@ -15,18 +16,27 @@ frontend www
...
-backend follower2
+    bind :8081
+    # Writes to input collections go to leader
+    use-server leader if { path_beg -i /v1/inputs/ }
+    server leader skip_leader:8081 weight 0
+    # Other (stream-creation) requests go to followers
+    balance roundrobin
+    server follower1 skip_follower1:8081
+    server follower2 skip_follower2:8081
+    server follower3 skip_follower3:8081
```
jberdine, line 31: FMI, is this sort of repetition something that can be made parametric reasonably easily?

examples/hackernews/reverse_proxy/haproxy.distributed.cfg
```diff
@@ -15,18 +16,27 @@ frontend www
...
+
+# Skip streaming backend
+backend stream
     mode http
     http-request set-path /v1/streams/%[path,field(4,/)]
+    use-server %[req.hdr(Follower-Prefix)] if TRUE
+    server follower1 skip_follower1:8080
```
jberdine, line 38: Here too

jberdine (commented):
Would be good to see what @beauby thinks, but LGTM

examples/hackernews/reverse_proxy/haproxy.distributed.cfg
bennostein: There's [`server-template`](https://docs.haproxy.org/3.0/configuration.html#4.2-server-template), but that requires all of the servers to have the same domain name.

My plan here is to use [haproxy maps](https://docs.haproxy.org/3.0/configuration.html#7.3.1-map) to make this elastic, using set-map/del-map to update these mappings on the fly.  but still figuring out how to make that work :)

PR #869, opened 2025-04-14
[skiplang/skc] Defer `ScalarType` construction to specializer.

beauby:
The `scalarType` field of `GClass` was not used beyond being copied when building the `SClass`es. Storing it on `GClass` forced the front end to know about the target platform's pointer size, and prevented sharing when compiling the same code for multiple platforms.


jberdine (approved):
Nice!

PR #872, opened 2025-04-16
Daily set of SKStore cleanups and preparatory diffs

mbouaziz:


skiplang/prelude/src/skstore/DMap.sk
```diff
@@ -21,23 +21,23 @@ base class Boundary<T> {
   | Exclusive(T)
 }
 
-base class DMap<Key: Orderable, Value> {
+base class DMap<K: Orderable, Value> {
```
beauby: Should `Value` become `V` for consistency?

mbouaziz: :shrug: 
What mattered here is that I don't get confused when reading the signature of a method.
There's no nearby class `Value`, though I agree there are some elsewhere. I'll do rename as well.

beauby (approved):

PR #873, opened 2025-04-17
Persistent Queue + simplify lazyGets

mbouaziz:
Adds persistent queues to the standard library and use it in the context to make the `lazyGetsQueue` easier to understand. As a side effect, the queue is now sensitive to changes of `lazyCapacity`

### DIFF
--- a/skiplang/prelude/src/stdlib/collections/persistent/Queue.sk
+++ b/skiplang/prelude/src/stdlib/collections/persistent/Queue.sk
@@ -41,8 +41,8 @@ class .Queue<+T> private {
     | Queue{head => List.Nil(), reversedTail => List.Nil(), siz} ->
       invariant(siz == 0);
       None()
-    | Queue{head => List.Nil(), reversedTail => List.Cons(rhd, rtl), siz} ->
-      rtl.revAppend(List::of(rhd)) match {
+    | Queue{head => List.Nil(), reversedTail => rt @ List.Cons _, siz} ->
+      rt.reversed() match {
       | List.Cons(res, head) ->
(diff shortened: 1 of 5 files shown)


skiplang/prelude/src/stdlib/collections/persistent/Queue.sk
```diff
@@ -0,0 +1,126 @@
...
+
+  fun pop(): ?(this, T) {
+    this match {
+    | Queue{head => List.Cons(res, head), siz} ->
+      Some((this with {head, siz => siz - 1}, res))
+    | Queue{head => List.Nil(), reversedTail => List.Nil(), siz} ->
+      invariant(siz == 0);
+      None()
+    | Queue{head => List.Nil(), reversedTail => List.Cons(rhd, rtl), siz} ->
+      rtl.revAppend(List::of(rhd)) match {
```
bennostein: Is this destructuring/restructuring of reversedTail necessary?  i.e. don't we know (from the fact that we passed through the previous case) that reversedTail isn't Nil?

```suggestion
    | Queue{head => List.Nil(), reversedTail, siz} ->
      reversedTail.reversed() match {
```

mbouaziz: You're right, though `reversedTail.reversed()` returns a `List`, not a `Cons`, and I didn't manage to change `reversed` signature so that `Cons(_).reversed()` is a `Cons`, and I hate `_ -> invariant_violation` because the best way to be sure it doesn't happen is when it's not written

mbouaziz: Actually found a solution using `extension class`es

PR #882, opened 2025-04-25
Fix recomputed graph observability

skiplabsdaniel:
Actually,  there is a bug in the particular case where the dependency graph is rebuilt and the resulting subscribed collection have removed files. The deletion notification are swallowed.


skipruntime-ts/native/src/Runtime.sk
```diff
@@ -310,6 +303,51 @@ fun buildResourcesGraph(
       }
     },
   );
+  // Copy the computed collection
+  //   - Ensure the name of the resource result
+  //   - Avoid prevent destruction in case of graph recompute.
```
bennostein:
```suggestion
  //   - Avoid destruction in case of graph recompute.
```

bennostein (approved):
Awesome, thanks Daniel!  One minor comment and also pushed a commit with my repro test but changes look good to me

PR #886, opened 2025-04-29
Move dirty from Context to EagerDir

mbouaziz:
For typed directories, ideally we'd need to keep keys and files in directories to avoid unsafe casts. There are a few cases of keys in the `Context`, some are more or less problematic.

This move looks safe as `dirty` is only accessed in update, and only updated in `writeEntry` with no weird interleaving with `setDir`.

Ideally, I'd do the same with `dirtyReaders` but there are too many interleavings, so there's a risk to exist a `setDir` between a change and another `setDir`, then missing some changes.

### DIFF
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -324,7 +324,7 @@ mutable class Context private {
   mutable newDirs: SortedSet<DirName> = SortedSet[],
   mutable globals: SortedMap<String, File> = SortedMap[],
   private mutable persistents: SortedMap<String, File> = SortedMap[],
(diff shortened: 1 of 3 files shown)


mbouaziz:
The failure on `skdb-wasm` looks non-transient, I can replicate it locally though I have no idea how this diff produces a `RuntimeError: memory access out of bounds`

skiplabsdaniel:
> The failure on `skdb-wasm` looks non-transient, I can replicate it locally though I have no idea how this diff produces a `RuntimeError: memory access out of bounds`
If it's 'Write-csv memory regression test' it's probably because the diff consume more memory.

mbouaziz:
Thanks for the heads-up!
It looks like this diff has a big bad impact on memory consumption, probably because its peak temporary garbage is higher.

PR #895, opened 2025-05-02
Decoupling the core SkipRuntime logic from the TypeScript-specific FFI layer.

skiplabsdaniel:
To be able to use the skipruntime core without having to implement the all typescript-specific ffi.


jberdine (approved):
This looks fine to me. I'm not sure about the `skipruntime-ts/skiplang/core` name though as the code there doesn't seem to be implementing the language, but I haven't seen the other uses of it. I guess there is a question of whether that code belongs with or next to the underlying `skstore` code, or somewhere under `skipruntime-ts` since it no longer appears to be specific to ts?

skiplabsdaniel:
I chose core for lack of a better name, because it is the heart of skipruntime.
About skipruntime-ts, I agree, I would rename it skipruntime and make two sub-directories skiplang and ts.

jberdine:
Ok, so you're thinking of a renaming after this PR is applied:

`skipruntime-ts/` -> `skipruntime/ts/`
`skipruntime-ts/skiplang` -> `skipruntime/skiplang`

And then if we ever resolve the dependency problem in the prelude, we could

`skiplang/prelude/src/skstore/` -> `skipruntime/skiplang/skstore`

That sounds good to me.

PR #902, opened 2025-05-09
[skipruntime/ts] Remove Managed from Context

skiplabsdaniel:


bennostein (approved):
LGTM! I remember we talked about this in person, seems a good foot-gun to remove.

I've pushed one small tweak, also dropping the actual managed tag off of the `ContextImpl` -- without that change I think a Context would slip by the runtime param checks.

PR #905, opened 2025-05-14
Another set of preparatory changes

mbouaziz:
- Few commits from #896
- Stuff

### DIFF
--- a/skiplang/prelude/src/stdlib/collections/persistent/SortedMap.sk
+++ b/skiplang/prelude/src/stdlib/collections/persistent/SortedMap.sk
@@ -770,12 +770,9 @@ base class .SortedMap<+K: Orderable, +V> uses  //  extends KeyedSequence<K, V>
     k: K2,
     v: V2,
   ): SortedMap<K2, V2>
-  | Nil() ->
-    new_static: Concrete<SortedMap<K2, V2>> = static;
-    new_static::node(k, v, Nil(), Nil())
+  | Nil() -> SortedMap::node(k, v, Nil(), Nil())
   | Node _ ->
-    new_static: Concrete<SortedMap<K2, V2>> = static;
-    new_static::balance(
+    SortedMap::balance(
       this.key,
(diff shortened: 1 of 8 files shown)


mbouaziz:
Removed the `Downcastable` trait for now as @skiplabsdaniel told me this version doesn't work because of compiler optimizations

skiplang/prelude/src/stdlib/collections/persistent/SortedMap.sk
```diff
@@ -326,14 +326,17 @@ base class .SortedMap<+K: Orderable, +V> uses  //  extends KeyedSequence<K, V>
...
     r: SortedMap<K2, V2> = this.right;
     k: K2 = this.key;
     v: V2 = this.value;
     compare(key, this.key) match {
     | LT() -> SortedMap::balance(k, v, l.setWith(key, value, f), r)
-    | EQ() -> SortedMap::node(k, f(v, value), l, r)
+    | EQ() ->
+      newV = f(v, value);
+      node: Node<K2, V2> = n;
+      node with {value => newV}
```
jberdine, line 339: For my understanding, this optimization only saves the re-calculation of the new node's size and height; in particular, it is a few loads and arithmetic instructions, but there is no memory reuse, right?

jberdine (approved):
LGTM. Though FTR I'm not convinced the last optimization makes a difference and is worth the code become a little less nice.

skiplang/prelude/src/stdlib/collections/persistent/SortedMap.sk
mbouaziz: Unless the compiler becomes more clever, there is no memory reuse indeed.
I don't know how the `with` is compiled but it seems it could be more easily optimized by the compiler.

The change made more sense when it was followed by the `phys_eq` optimization. Here, it just makes it clearer that the tree's shape is not changed, only the value is updated.

PR #907, opened 2025-05-16
Upgrade package-lock to latest versions of SKIP

hubyrod:


package-lock.json
```diff
@@ -102,70 +102,6 @@
         "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
       }
     },
-    "examples/blogger/reactive_service/node_modules/@skip-adapter/postgres": {
```
jberdine, line 105: Do you understand why there is so much removed in this update?

hubyrod: They all are dependencies on `"version": "0.0.15"`
Dependencies on 0.0.16 are already in. What i'm not sure of is why there where not removed in the first place.

jberdine (approved):

jberdine:
FWIW we usually prefer to rebase over merge

package-lock.json
bennostein: `package-lock.json` gets updated automatically by e.g. `npm install` but it can happen that you update some `package.json` and commit it without running any make target that updates the lockfile -- looks like that's what happened here.

Not a big deal though -- it'll get caught by the next commit that touches those things, or with a followup like this.

hubyrod: Most prabably, yes, you are right.

PR #909, opened 2025-05-19
Prep for building node addon against precompiled skipruntime

jberdine:
This PR changes the build in preparation for building the native node addon against a precompiled skipruntime library. The aim is to enable end users to first install the binary skip runtime to their host and then install the addon just as a standard npm dependency, without needing to build the skiplang toolchain or have it installed. To this end, this PR:

- adapts the npm build (package.json and binding.gyp files) to expect the skipruntime library to be installed on the host;

- adapts the dev build (Makefiles) to build the skipruntime library from the current sources, and set the other builds and tests to use this current version.

At a high level, what this means is that the `npm` commands use the host-installed skipruntime, and are therefore meant for use by end users; while the `make` commands use the skipruntime built from current sources, and are therefore meant for development and testing use.

### DIFF
--- a/.circleci/base.yml
+++ b/.circleci/base.yml
@@ -54,7 +54,7 @@ jobs:
           submodules: true
       - run:
           name: Typecheck and lint typescript sources
-          command: SKIPRUNTIME=$(pwd)/build/skipruntime npm install && npm run build && npm run lint
+          command: make check-ts
 
   compiler:
     docker:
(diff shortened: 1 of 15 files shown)


jberdine:
This is not useful on its own, but I wanted to put this PR up in case anyone sees issues or problems with this approach.

jberdine:
Superseded by #969

PR #911, opened 2025-05-21
Another round of SKStore preparatory commits

mbouaziz:
- Removal of dead stuff
- Simplifications
- Moving things involving `Key`s to directories


jberdine (approved):
LGTM

PR #916, opened 2025-05-26
[skstore] Rework Sub/DirSub classes

mbouaziz:
- Make sure a function taking keys (in `NWatch`) is used for a single directory at a time
- So that later we can add casts given the precise key type
- Plus other minor cleanups

### DIFF
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -766,15 +766,12 @@ mutable class Context private {
     | NTail() ->
       cond = unfreezeCond(sub.cond);
       _ = condBroadcast(cond)
-    | NWatch{from, fn} ->
-      for (dirSub in sub.dirSubs) {
-        dirName = dirSub.dirName;
-        this.unsafeMaybeGetDir(dirName) match {
-        | Some(dir @ EagerDir _) ->
-          startOpt = if (start > from) Some(start) else None();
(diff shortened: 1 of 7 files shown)


skiplang/prelude/src/skstore/Context.sk
```diff
@@ -766,15 +766,12 @@ mutable class Context private {
...
-          startOpt = if (start > from) Some(start) else None();
-          dir.notifySubWatch(startOpt, this.tick, fn)
-        | _ -> void
-        }
+    | NWatch{dirName, from, fn} ->
+      this.unsafeMaybeGetDir(dirName) match {
+      | Some(dir @ EagerDir _) ->
+        startOpt = if (start > from) Some(start) else None();
+        dir.notifySubWatch(startOpt, this.tick, fn)
+      | _ -> void
```
jberdine, line 781: I guess you already have in mind why this new version is equivalent; can I be lazy and ask for a hint?

jberdine (commented):
LGTM, just one question

skiplang/prelude/src/skstore/Context.sk
mbouaziz: There was only one `DirSub` for `NWatch` and we actually don't need the whole `DirSub`, just the `DirName` which is now part of `NWatch` rather than `Sub`.
Not sure that's what you meant

jberdine (approved):

PR #917, opened 2025-05-26
[skstore] Make KeyRange generic in Key

mbouaziz:


sql/src/SqlSelect.sk
```diff
@@ -303,7 +303,7 @@ class SelectEvaluator{
         RowValues::type,
         result.dirName,
       );
-      filtered = handle.filter(
+      _filtered = handle.filter(
```
jberdine, line 375: I don't understand this change; doesn't it result in the filter being ignored in this case?

jberdine (commented):
LGTM, but I'm confused in one place

sql/src/SqlSelect.sk
mbouaziz: No, like the _else_ branch, it creates a directory with the given name (`childName`) in the context, what we ignore is the resulting handle

jberdine (approved):

PR #925, opened 2025-06-04
Hackernews skip service grooming

hubyrod:


hubyrod:
@bennostein merely a proposal, the idea is to make it readable and explicit to dev unfamiliar with it.

beauby (approved):

PR #931, opened 2025-06-06
[skstore] Generic data structures (part 1)

mbouaziz:
Making some SKStore classes generic in `Key` and `File`

### DIFF
--- a/skiplang/prelude/src/skstore/EagerDir.sk
+++ b/skiplang/prelude/src/skstore/EagerDir.sk
@@ -58,7 +58,7 @@ class Reducer{
   modified: SortedMap<Key, Array<File>>,
 } {
   static fun create(
-    fixedData: FixedData<Array<File>>,
+    fixedData: FixedData<Key, Array<File>>,
     reducer: IReducer,
   ): this {
     acc = mutable Vector[];
@@ -114,7 +114,7 @@ class Reducer{
   }
 
   private static fun makeCompactKeyIterator(
-    fixedData: FixedData<Array<File>>,
+    fixedData: FixedData<Key, Array<File>>,
     key: Key,
     current: mutable IntRef,
   ): readonly Vector<File> {
@@ -517,7 +517,7 @@ class DataMap private {
   }
 }
 
-class FixedDataMap{data: FixedData<Array<File>>, tombs: FixedDir<Path>} {
+class FixedDataMap{data: FixedData<Key, Array<File>>, tombs: FixedDir<Path>} {
   static fun create<V: File>(
     vec: mutable Vector<FixedRow<Key, Array<V>>> = mutable Vector[],
     tombs: mutable Vector<FixedRow<Key, Path>> = mutable Vector[],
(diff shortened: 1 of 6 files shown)


skiplang/prelude/src/skstore/EagerDir.sk
```diff
@@ -502,28 +502,31 @@ class DataMap private {
...
-  fun remove(k: Key): this {
+  fun remove(k: K): this {
     !this.data = this.data.remove(k);
     !this.tombs = this.tombs.remove(k);
     this
   }
 }
 
-class FixedDataMap{data: FixedData<Array<File>>, tombs: FixedDir<Path>} {
+class FixedDataMap{
```
bennostein, line 520: What's the blocker on this (and EagerDir) also getting a type parameter `K`?  in practice these should always be keyed on some more-precise-than-`Key` type, right?

Or, is there no blocker and you're just planning to get to those in part 2?

skiplang/prelude/src/skstore/FixedDir.sk
```diff
@@ -205,16 +213,17 @@ class CompactFixedDir protected {
...
-  fun get(i: Int): FixedRow<Array<File>> {
+  fun get(i: Int): FixedRow<Key, Array<File>> {
     compactRow = this.data.unsafe_get(i);
     rowValues = SKDB.RowValues(compactRow.values, compactRow.repeat);
     rowKey = SKDB.RowKey(rowValues, this.metadata.kinds);
     sourceKey = compactRow.sourceKey match {
     | None() -> rowKey
     | Some(otherKey) -> otherKey
     };
+    fixedRowKey: Key = rowKey;
```
bennostein: What's going on here - just helping out the type checker?  Seems like this shouldn't be needed

bennostein (approved):
Couple questions/comments but no objections, lgtm!

skiplang/prelude/src/skstore/EagerDir.sk
mbouaziz: That's indeed in part 2 and I still have kinda blocker, which is `CompactFixedDir`.

Skip type system is not powerful enough to do what I need nicely:
- check that we are in the special case `FixedDataMap<SKDB.RowKey, SKDB.RowValues>` (doable with the new `maybeCast`, which is safe)
- use the specific constructor that may use `CompactFixedDir` (fine)
- unify with the generic case `FixedDataMap<K, V>` (requires a very ugly cast)

skiplang/prelude/src/skstore/FixedDir.sk
mbouaziz: Otherwise I get a `FixedRow<SKDB.RowKey>` but I needed a FixedRow<Key>`.

Though maybe I could get rid of this now that I made `FixedRow` covariant in `Key`

skiplang/prelude/src/skstore/EagerDir.sk
bennostein: I think if that's a blocker, we should just delete the `Compact*` stuff -- it was an optimization for SKDB that isn't needed at all for skip runtime services and would be easy enough to resurrect if we need more space-efficient SKDB representations in the future

PR #941, opened 2025-06-12
[SKStore] Generic data structures (part 2)

mbouaziz:
Follow-up of #931: make `FixedDataMap` generic.

`CompactFixedDir` can still be used but only if a `FixedDataMap` is instanciated with the SKDB types, which will happen only once we have typed `EagerDir`s and with use the most precise types (i.e. not `Key` and `File`).


mbouaziz:
Given that `CompactFixedDir` is not used, the memory usage test fails. I'll change the PR to try to use it for `Key`/`File` too.

bennostein (approved):

mbouaziz:
Removed specialization of `CompactFixedDir`

PR #942, opened 2025-06-13
[skiplang] Update to LLVM 20.

beauby:


mbouaziz:
I remember I had checked the changelog for versions 16 and 17. Have you done the same for 18 and 19?

beauby:
> I remember I had checked the changelog for versions 16 and 17. Have you done the same for 18 and 19?

Yes, nothing seems to be of concern in 18/19. However, the main thing that should be an issue to us is the removal of support for typed pointers in version 17, though I have been daily driving LLVM 19 for a few months and haven't come across any issue.
I mainly opened this PR to check whether any issue would arrise in CI, but had forgotten that the docker image was not built dynamically.

mbouaziz:
Yeah, you probably need to push docker images under another tag for this PR

mbouaziz:
See #736 for commits to cherry-pick

beauby:
> See #736 for commits to cherry-pick

Yup, those are one my list.

PR #943, opened 2025-06-13
[skiplang/skc] Get rid of `async`/`await`.

beauby:
It is already broken, and not compatible with `SKStore`.


mbouaziz:
I like red lines, though what's the plan with async in Skip lang?
Most languages have added async support recently so it feels weird to get rid of it even though we don't use it.
I wouldn't give up too early on having `SKStore` support async mappers

mbouaziz:
After discussing with @pikatchu:
- let's keep it
- we can deactivate it if needed
- if there is a cost of maintaining a part of it, that part can be replaced with an `invariant_violation`, that's a lazy killing

PR #957, opened 2025-06-25
[skipruntime] Remove heavy copy mechanism

skiplabsdaniel:
Adds the possibility of observing collections that will be recreated or replaced.
This means that when a dependency graph is dynamic, it can be correctly observed, resulting in an `init` event being sent each time a graph of a resource is recomputed.


skiplang/prelude/src/skstore/EagerDir.sk
```diff
@@ -668,6 +668,7 @@ mutable class FixedDataMapIterator<T> private {
 
 class EagerDir protected {
   input: Bool,
+  private created: Tick,
```
mbouaziz, line 671: Isn't it (more or less) the same as `dir.getTime()`?

mbouaziz (commented):
Reviewed the SKStore part only

skiplang/prelude/src/skstore/EagerDir.sk
mbouaziz: No, thanks for checking

Issue #965, opened 2025-07-01
[stdlib] Splitting prelude


mbouaziz:
I'm again encountering the issue of something compiling fine locally but failing on the CI because it is using a different stdlib.

IIRC, the plan was to split the prelude into:
- intrinsics: declaration of functions that the compiler knows, sk files + the llvm preamble (must be distributed with the toolchain and we must use that version rather than the local one)
- runtime: what compiled programs will be linked with to run, .a file (same)
  - both will be added to the bootstrap and can be combined as an sklib file
- stdlib: just a bunch of sk files

Anything against this plan?

jberdine:
I think that there are also a few corners that we will need to be careful of. For example the type definition in `skiplang/prelude/src/native/StringIterator.sk` needs to be kept in sync with `skiplang/prelude/runtime/string.c` since some C functions access the representation directly. So the division between "intrinsics" and "stdlib" might be not very clear-cut.

mbouaziz:
Indeed, there are a few classes to handle specifically:
1. `native value class`es: `Void`, `Bool`, `Char`, `Float`, `Int`, `Int8`, `Int16`, `Int32`, `RawStorage`, `GCPointer`, `NonGCPointer`
2. `native class`es: `Array`, `String`
  - though is there a difference between value and non-value native classes given they have no front-end fields?
3. `native base class`es: `Obstack`, `Page`
  - though why are they `base` classes given they have no children?
4. `native class SpawnFileActions`: malloc'ed structure, should probably be done differently
5. classes that are not native but there's a link between their representation and the runtime: `StringIterator`, `Mutex`, `Condition`

1 and 2 should be declared in the intrinsics and have extensions in the stdlib.
3, 4, 5 can be in the stdlib, though they somewhat depend on the compiler version (but Skip code too!). It can contain C code as well.

PR #966, opened 2025-07-04
[skargo] Poor man's solution to building build dependencies in another state

mbouaziz:
+ other minor things.

Working on #965, I created `core` and added a dependency from `std` to `core`.
Though on `skargo check/build` in `std`, `skc` failed building `core` because of a duplicate stdlib definition.
Indeed, in its state was already the embedded std that was used to build `cc` and `pkg-config`, both dependencies of the build-script of `std`.

My solution: build build dependencies (and the build script, which previously used no state) in another state.

It's not perfect:
- I think other build scripts should probably be built in other states
- if a build dependency has dependencies that are shared with something else during the same skargo invocation, I guess it'll be it once, in the build state...

### DIFF
--- a/skiplang/skargo/src/build_runner.sk
+++ b/skiplang/skargo/src/build_runner.sk
@@ -388,26 +388,32 @@ mutable class BuildRunner(
 
     skc.args(unit.target.srcs);
 
-    if (!unit.target.is_build_script()) {
-      // TODO: Once we have namespaced packages, we can share the state.db
-      // across architectures.
-      state_db_path = Path.join(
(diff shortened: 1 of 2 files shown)


mbouaziz:
After discussing with @beauby:
- this fix is kinda OK
- other possibility: not use a state at all for build things
- for build dependencies that are dependencies too, we could as well use the normal state
- needs a fix: being a regular dependency or a build dependency is a transitive property; currently wrongly done at one level only; could be colored when building the graph

Plans to split stdlib abandoned for now so closing the PR

PR #969, opened 2025-07-09
Support for binary releases of the skip runtime, for native node addon

jberdine:
This PR changes the build, adds release and install scripts, and updates docs to support building the native node addon against a precompiled skipruntime library. The aim is to enable end users to first install the binary skip runtime to their system and then install the addon just as a standard npm dependency, without needing to build the skiplang toolchain or have it installed.

In short, this PR:

- adapts the npm build (package.json and binding.gyp files) to expect the skipruntime library to be installed on the host

- adapts the dev build (Makefiles) to build the skipruntime library from the current sources, and set the other builds and tests to use this current version

- adds scripts to build and install the binary runtime library

- adds a small test that installs and uses the native node addon in a container

- updates user and dev docs

At a high level, the build changes mean that the npm commands use the system-installed skipruntime, and are therefore meant for use by end users; while the make commands use the skipruntime built from current sources, and are therefore meant for development and testing use.

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -44,3 +44,35 @@ changes, the addon package version must be bumped.)  To update from version
    building/running with the new versions, add those changes to your PR, and
    land it!
 
+8. Release the Skip runtime binaries, as below.
+
+## Release Skip runtime binaries
+
+The `@skipruntime/native` node addon dynamically links to a native library
+containing the Skip runtime. The versions of the NPM package and binary
+release must match. Therefore whenever NPM packages are released the binary
+runtime release must be as well. The installation scripts download the
+binary runtime library from a github release tagged with `vM.N.O` where
+`M.N.O` is the version of the NPM packages.
+
+The release procedure for the runtime library binary is:
+
+1. Run `bin/build_runtime.sh` from a clean repo with HEAD at the commit that
+   was/will be used for version `M.N.O` of the NPM packages. This leaves
+   `libskipruntime.so-<OS>-<ARCH>` files in the `build` directory.
+
+2. Visit https://github.com/skiplabs/skip/releases/new
+
+3. Click "Choose a tag"
+
+4. Enter `vM.N.O` as the tag name
+
+5. Click "Create new tag...on publish"
+
+6. Enter a release title and description
+
+7. Attach `build/libskipruntime.so-linux-arm64` and `build/libskipruntime.so-linux-amd64`
+
+8. Click "Publish release"
+
+9. Test the new release by running `cd skipruntime-ts/tests/native_addon/ ; ./run.sh`
--- a/bin/README.md
+++ b/bin/README.md
@@ -13,8 +13,11 @@ clone, you can run `release_docker_ci_images.sh`.
 
 We keep all **public** NPM packages (i.e. `@skipruntime/*`, `@skiplabs/skip`,
 and `@skip-adapter/*`) at the same version so as to keep updates, documentation,
-and version management simple.  To update from version `$OLD` to version `$NEW`,
-perform the following steps:
+and version management simple.  (Note that the build script of the native addon
+relies on this convention since it links to the skipruntime library with the same
+version as the addon package, so whenever anything in the skipruntime library
+changes, the addon package version must be bumped.)  To update from version
+`$OLD` to version `$NEW`, perform the following steps:
 
 1. Check out `main` and make sure your working copy is clean
 
@@ -24,7 +27,7 @@ perform the following steps:
 
 3. Bump _all_ `./skipruntime-ts/**/package.json` version strings and
    inter-dependencies from `$OLD` to `$NEW`, e.g. by running `sed -i ''
-   's/$OLD/$NEW/g' $(g grep -l $OLD skipruntime-ts/**/package.json)`.
+   's/$OLD/$NEW/g' $(git grep -l $OLD skipruntime-ts/**/package.json)`.
 
 4. Build and test with `make -C skipruntime-ts build test`.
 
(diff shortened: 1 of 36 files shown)


jberdine:
To test this I also made a github release on my fork of the repo, made npm packs of the current ts sources, adjusted skipruntime-ts/tests/native_addon/Dockerfile to use the fork and packs, and ran the native_addon test that builds and runs a container that does a minimal install of the deps for building the native node addon that links with the binary libskipruntime downloaded from github. That works, but I don't know a good way to make a regular procedure to test releases before they are released.

bin/README.md
```diff
@@ -27,7 +30,7 @@ perform the following steps:
 
 3. Bump _all_ `./skipruntime-ts/**/package.json` version strings and
    inter-dependencies from `$OLD` to `$NEW`, e.g. by running `sed -i ''
-   's/$OLD/$NEW/g' $(g grep -l $OLD skipruntime-ts/**/package.json)`.
+   's/$OLD/$NEW/g' $(git grep -l $OLD skipruntime-ts/**/package.json)`.
```
bennostein, line 30: Oops, that's an alias from my shell config -- nice catch!

skipruntime-ts/tests/native_addon/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:latest
```
bennostein, line 1: Is there a strong reason we need `ubuntu:latest` here and `debian:stable-slim` for the other new Dockerfile? Might be nice to use the same base image for both if not, just to keep bandwidth and storage requirements down

skipruntime-ts/tests/native_addon/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "module",
+  "dependencies": {
+    "@skipruntime/core": "",
```
bennostein, line 4: what does the empty version string do here?

skipruntime-ts/tests/native_addon/Dockerfile
jberdine: Not saying they are strong reasons, but my thinking is that this one is trying to simulate an expected end user setup, so the latest ubuntu LTS seems the right choice, and the other one is aiming to build the binaries with the "oldest reasonable" linux, so debian stable, and the slim version is enough.

Note that these images don't need to be pushed to docker or stored.

skipruntime-ts/tests/native_addon/package.json
jberdine: My understanding is that it equates to any one, and so an npm install will get the latest one in the package db.

skipruntime-ts/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+COPY ./skiplang /work/skiplang
+
+WORKDIR /work/skiplang/compiler
+RUN make clean && make STAGE=0
+
+FROM base AS build
+
+COPY --from=skiplang /work/skiplang/compiler/stage0/bin/ /usr/bin/
+COPY --from=skiplang /work/skiplang/compiler/stage0/lib/ /usr/lib/
```
mbouaziz: (nit)
I'm wondering whether there's a way we could share more with the main `Dockerfile`.
Maybe modifying/using the `apt-install.sh` script?

jberdine: You could be right, but so far I'm unsure. The objective of this Dockerfile is to build the runtime library with as old a distribution, and hence hopefully libc, as possible in order to maximize compatibility with users' machines. It isn't clear that we want to apply the same constraint to everything else, especially if some in the team do dev work outside of docker. Also, the current approach is to just bet that using the oldest reasonable debian will end up using an old-enough libc, but this approach might end up not working. So the redundancy in the Dockerfiles could well just be coincidental, and for now I didn't unify them.

skipruntime-ts/Makefile
```diff
@@ -42,7 +53,6 @@ nodeserver-%:
 clean:
 	make -C .. clean
 
-.PHONY: run-test
```
mbouaziz: ?

sql/Makefile
```diff
@@ -20,7 +17,8 @@ check-all: check-src check-tests
 
 .PHONY: build
 build:
-	../bin/cd_sh .. "npm install && npm run build -w skdb -w skdb-tests"
+	${MAKE} -C ../skipruntime-ts install-all
+	../bin/cd_sh .. "npm run build -w skdb -w skdb-tests"
```
mbouaziz, line 21: Unrelated but I don't think we should need skipruntime-ts for skdb

mbouaziz (approved):
I'm not sure I've grasped everything but it looks solid and consistent

skipruntime-ts/Makefile
jberdine: This is only a subtarget of `test` and cannot be reliably invoked manually. The convention for the rest of the targets seems to be to have `.PHONY` on targets that the `Makefile` handles the dependencies and steps to (try to) ensure that they can be run whenever, and leave off `.PHONY` for the targets where there is a workflow users need to follow.

I can remove it if you want, but it increases consistency.

sql/Makefile
jberdine: That seems right to me. But AFAIU that is the only place we even get e.g. tsc, so something finer grained would need to be introduced to remove this.

skipruntime-ts/Makefile
mbouaziz: My understanding is that `.PHONY` is for every target that is not a file

jberdine: True, but that's not what we do in this file. Anyhow, I'll remove it, it is tangential to this PR.

Issue #625, opened 2025-01-09
Considere use of pnpm to ease project management


skiplabsdaniel:
The `npm` project manager has a few shortcomings: it forces you to manage the versions of each package manually, which makes the publishing process tedious, and it doesn't allow the use of aliases, which prevents certain package delivery optimizations.
`pnpm` seems to manage multi-package projects much more cleanly.
It might be interesting to test whether using `pnpm` would not greatly facilitate the development workflow, and in particular publishing.

niladrix719:
great idea! would like to give it a try

mbouaziz:
Bringing #625 back up because it's a deliberate policy call — worth settling here in the issue rather than in another PR that goes stale (both prior attempts just did; see below).

## Where things stand today

We're a single npm-workspaces monorepo: ~24 workspace members under one root `package.json` / `package-lock.json`, all scripts are `npm run X --workspaces`, and it works with **zero contributor-side config** — there's no `packageManager` pin, no root `.npmrc` affecting the PM, no version-manager requirement anywhere. pnpm exists in exactly two places: the `cache_invalidation/www` and `cache_invalidation/edge_service` examples, each shipping its own `pnpm-lock.yaml`.

To be straight about motivation: I went looking for **actual npm pain and found essentially none of the classic pnpm-justifying failures** — zero `resolutions`, zero `nohoist`, zero `workspace:` specifiers, no reported phantom-dep bug, no breakage in #625 itself (the issue literally says "it might be interesting to test"). The one recurring real incident is the `serialize-javascript` dedup, which reappeared in a **nested mocha copy twice** (#1219, then #1314) because npm failed to re-dedup it — that's the strongest single datapoint on the pro side, and it's narrow.

## Pros of moving to pnpm

- **Split-brain removal.** `edge_service` is currently triple-managed: it's a root workspace member, *and* ships `pnpm-lock.yaml` + `pnpm.overrides`, *and* its Dockerfile runs `npm install`. `www` is `private:true` and not in the workspace at all. The boundary between "npm workspace" and "pnpm leaf" is genuinely undefined right now; picking one PM ends that.
- **Phantom deps would surface.** `skipruntime-ts/core` and `helpers` declare **zero** dependencies and devDependencies yet run `tsc` + `eslint` and extend `@skiplabs/tsconfig` — they only resolve today via root hoisting. pnpm's strict non-hoisted layout would make every such undeclared import explicit. (Upside and downside — see cons.)
[...]

PR #974, opened 2025-07-20
switch skipruntime-ts to pnpm

hubyrod:
- **fix Makefile**
- **switch skipruntime-ts to pnpm**


.circleci/base.yml
```diff
@@ -53,9 +53,12 @@ jobs:
     steps:
       - custom-checkout:
           submodules: true
+      - run:
+          name: Install pnpm
+          command: npm install -g pnpm
```
jberdine, line 58: Opinions vary but FWIW I tend to prefer using the long form of options when written out in scripts, etc. to make it easier to read for people who might not have the short forms memorized.

Makefile
```diff
@@ -19,16 +19,10 @@ export SKIPRUNTIME
...
+.PHONY: pnpm
+pnpm:
+	pnpm install && pnpm build
 
-build/package/package.json:
-	@echo "{" > build/package/package.json
-	@echo "  \"dependencies\": {" >> build/package/package.json
-	@echo "      \"skdb\": \"file:skdb\"" >> build/package/package.json
-	@echo "  }" >> build/package/package.json
-	@echo "}" >> build/package/package.json
```
jberdine, line 31: Something you might be hitting that isn't obvious is that here the npm stuff is referring to the support for using *skdb* from js using npm packages.

Makefile
```diff
@@ -125,23 +119,23 @@ check-fmt:
 # regenerate api docs served by docs-run from ts sources
 .PHONY: docs
 docs:
-	npm install && npm run build
-	cd www && rm -rf docs/api && npm install && npx docusaurus generate-typedoc
+	pnpm install && pnpm build
+	cd www && rm -rf docs/api && pnpm install && npx docusaurus generate-typedoc
```
jberdine, line 123: The various `npx` invocations should be changed to something pnpm friendly, right? Maybe `pnpm exec`?

bin/check-ts.sh
```diff
@@ -12,5 +12,5 @@ DIRS=$( jq --raw-output ".workspaces[]" package.json )
```
jberdine, line 12: Note that the line above reads the workspaces field of the package.json file that isn't here anymore.

skipruntime-ts/Makefile
```diff
@@ -1,22 +1,23 @@
 SHELL := /bin/bash
 
 SKARGO_PROFILE?=release
-NPM_WORKSPACES=$(shell jq --raw-output "[.workspaces[] | select((startswith(\"sql\") or contains(\"examples\")) | not)] | map(\"-w \" + .) | .[]" ../package.json)
+# For pnpm, use filter patterns instead of individual -w flags
+PNPM_SKIPRUNTIME_FILTER=--filter="skipruntime-ts/core" --filter="skipruntime-ts/helpers" --filter="skipruntime-ts/wasm" --filter="skipruntime-ts/addon" --filter="skipruntime-ts/adapters/kafka" --filter="skipruntime-ts/adapters/postgres" --filter="skipruntime-ts/tests" --filter="skipruntime-ts/server" --filter="skipruntime-ts/examples" --filter="skipruntime-ts/metapackage" --filter="skiplang/prelude/ts/binding" --filter="skiplang/prelude/ts/wasm" --filter="skiplang/prelude/ts/worker" --filter="skiplang/skdate/ts" --filter="skiplang/skjson/ts/binding" --filter="skiplang/skjson/ts/wasm" --filter="eslint-config" --filter="tsconfig"
```
jberdine, line 5: The previous version has the desirable quality that it only explicitly mentions the things we want to exclude, and so will automatically include anything new. Is there no way to preserve that characteristic?

skipruntime-ts/Makefile
```diff
@@ -44,14 +45,14 @@ clean:
...
-	../bin/cd_sh tests "npm run test"
+	../bin/cd_sh tests "pnpm run test"
 
 .PHONY: test
 test: install build run-test
 
 .PHONY: build-examples
 build-examples: build
-	../bin/cd_sh examples "npm run build -w skipruntime-examples"
+	@echo "Skipping examples build due to dependency issues"
```
jberdine, line 55: Is this expected to be temporary, or is there some bigger issue?

.circleci/base.yml
hubyrod: Absolutely. I'll make command line options explicit.

mbouaziz:
Closing as stale — not abandoning the idea. This draft has been cold since 2025-07-20, is in merge conflict, and `skipruntime-ts` has since been reorganized (new `core`/`helpers`/`wasm`/`adapters/*`/`metapackage` subpackages), so the diff edits a package layout that no longer exists. Reviving it would mean redoing the conversion against current `main` rather than rebasing.

The pnpm question itself is still open — I'm consolidating it on #625 so we can make the call deliberately. We can reopen or re-cut this any time. Thanks @hubyrod for the work here.

— Claude (Opus 4.8)

PR #986, opened 2025-08-13
Add functions to allow artifacts reload in skip services

skiplabsdaniel:
No reuse in this version.

For Mapper/Reducer/LazyCompute:
- Replace the corresponding constructor, then invalidate the dependent collections.

For Resource:
- Load a new instance for each existing resource instance and, if everything is correct, replace the old resources with the new ones.

For the service:
- Load a new instance of the service and, if everything is correct, reload all the resources that exist in the service.

### DIFF
--- a/skiplang/prelude/src/skstore/EagerDir.sk
+++ b/skiplang/prelude/src/skstore/EagerDir.sk
@@ -397,6 +397,10 @@ class DataMapValue{
     | Some(x) -> Some(x.i1)
     }
   }
+
+  fun sources(): Array<Path> {
+    this.map.keys().collect(Array)
+  }
 }
 
 class DataMap<K: Orderable> private {
@@ -2291,6 +2295,29 @@ class TEagerDir extends EagerDir, TDir {
         static::addChildToParent(context, parentName, childName, rangeOpt);
       }
     };
+  }
+
+  fun invalidate(context: mutable Context): void {
+    parents = SortedSet::createFromIterator(
+      this.parents.items().map(v -> v.i0),
+    );
+    this.keys().each(key -> {
+      sources = this.data.maybeGet(key) match {
+      | None() -> this.fixedData.data.getIter(key).map(v -> v.i1).collect(Array)
+      | Some(modified) -> modified.sources()
+      };
+      for (source in sources) {
+        if (parents.contains(source.dirName)) {
+          context.updateDirtyReader(
+            TArrowKey::create{
+              parentName => source.dirName,
+              childName => this.dirName,
+              key => source.baseName,
+            },
+          )
+        }
+      }
+    });
   }
 }
 
--- a/skiplang/prelude/src/skstore/EagerDir.sk
+++ b/skiplang/prelude/src/skstore/EagerDir.sk
@@ -725,6 +725,12 @@ base class EagerDir protected {
     this
   }
 
+  // renew creation time to force notification
+  fun renew(context: mutable Context): void {
+    !this.created = context.getTick();
+    context.setDir(this);
+  }
+
   protected fun writeDiff(
     isReset: Bool,
     changes: mutable Iterator<Key>,
--- a/skiplang/prelude/src/skstore/EagerDir.sk
+++ b/skiplang/prelude/src/skstore/EagerDir.sk
@@ -2309,16 +2309,13 @@ class TEagerDir extends EagerDir, TDir {
   }
 
   fun invalidate(context: mutable Context): void {
-    parents = SortedSet::createFromIterator(
-      this.parents.items().map(v -> v.i0),
-    );
     this.keys().each(key -> {
       sources = this.data.maybeGet(key) match {
       | None() -> this.fixedData.data.getIter(key).map(v -> v.i1).collect(Array)
       | Some(modified) -> modified.sources()
       };
       for (source in sources) {
-        if (parents.contains(source.dirName)) {
+        if (this.parents.containsKey(source.dirName)) {
           context.updateDirtyReader(
             TArrowKey::create{
               parentName => source.dirName,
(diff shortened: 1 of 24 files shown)


skipruntime-ts/core/src/index.ts
```diff
@@ -111,8 +112,33 @@ class Handles {
...
+        map.delete(id);
+        if (map.size == 0) {
+          this.links.delete(name);
+        }
+      }
+    }
     return current;
   }
+
+  registerLink<T>(name: string, ref: string, handle: Handle<T>): void {
```
mbouaziz: Nit: you may want to restrict handles of type `Handle<HandleInfo<T>>`

skiplang/prelude/src/skstore/EagerDir.sk
```diff
@@ -2292,6 +2296,29 @@ class TEagerDir extends EagerDir, TDir {
       }
     };
   }
+
+  fun invalidate(context: mutable Context): void {
+    parents = SortedSet::createFromIterator(
+      this.parents.items().map(v -> v.i0),
+    );
```
mbouaziz: nit: Implement `containsKey` for `FixedSingle`? Basically `maybeGet is Some`

skipruntime-ts/core/src/index.ts
```diff
@@ -139,6 +159,25 @@ class Handles {
     }
     map.set(handle, ref);
   }
+
+  getLinks(name: string): Nullable<Map<number, string>> {
```
mbouaziz: Should this `number` be replaced with `Handle<HandleInfo<unknown>>`?

skipruntime-ts/core/src/index.ts
```diff
@@ -722,6 +779,86 @@ export class ServiceInstance {
...
+              info.what as string,
+              ctor,
+              info.params as DepSafe[],
+            );
+            this.refs.handles.replace(
+              handle as Handle<HandlerInfo<object>>,
+              newObj,
+            );
+          } else {
+            throw new Error("Only mappers en reducers can be replaced");
```
mbouaziz: en -> and

skipruntime-ts/core/src/index.ts
```diff
@@ -722,6 +779,86 @@ export class ServiceInstance {
       return Promise.reject(this.refs.handles.deleteHandle(errorHdl));
     }
   }
+
+  reload(ctors: (new (...args: any[]) => any)[]) {
+    const collections = new Set<string>();
+    for (const ctor of ctors) {
+      if (!("name" in ctor)) continue;
```
mbouaziz: Fail hard instead?

skipruntime-ts/skiplang/core/src/Runtime.sk
```diff
@@ -448,6 +448,21 @@ fun closeService(): Result<Array<Float>, .Exception> {
...
 
+fun invalidateCollections(
+  jsonCollections: SKJSON.CJArray,
+): Result<void, .Exception> {
+  SKStore.runWithResult(context ~> {
+    collections = SKJSON.expectArray(jsonCollections).map(SKJSON.asString);
+    collections.each(collection -> {
+      context.maybeGetDir(SKStore.DirName::create(collection)) match {
+      | Some(dir) -> dir.invalidate(context)
+      | _ -> void
```
mbouaziz, line 504: Fail hard instead?

skipruntime-ts/wasm/src/internals/skipruntime_module.ts
```diff
@@ -241,6 +241,9 @@ export interface FromWasm {
 
   // closeClose
   SkipRuntime_closeService(): ptr<Internal.CJSON>;
+  SkipRuntime_invalidateCollections(
+    collections: ptr<Internal.CJArray<Internal.CJSON>>,
```
mbouaziz, line 246: ...`CJArray<Internal.CJString>`

Same below

skipruntime-ts/core/src/index.ts
```diff
@@ -178,6 +180,38 @@ class Handles {
...
     return undefined;
   }
+
+  clone(): Handles {
+    return new Handles(
+      this.nextID,
+      [...this.objects],
+      [...this.freeIDs],
+      /** Index to get all mapper handles from a class name */
+      new Map<string, Map<number, string>>(this.links),
```
mbouaziz: Should this be a deep copy instead?

skipruntime-ts/core/src/index.ts
```diff
@@ -178,6 +180,38 @@ class Handles {
...
+    arr.length = 0;
+    for (const v of w) {
+      arr.push(v);
+    }
+  }
+
+  private resetMap<K, V>(map: Map<K, V>, w: Map<K, V>): void {
+    map.clear();
+    w.forEach((value, key) => map.set(key, value));
+  }
```
mbouaziz: Given that the reset is only called with a `handles` object that will be thrown away, no need to deep copy the arrays and maps, simply replace the fields? (unmark them as `readonly` if needed)

Or implement a try/catch wrapper in `handles` to hide the clone/reset thingy

skipruntime-ts/core/src/index.ts
```diff
@@ -287,13 +288,22 @@ class Handles {
     return undefined;
   }
 
-  registerService(service: Handle<ServiceDefinition>) {
-    this.service = service;
+  clearConstrutors() {
```
mbouaziz: typo

mbouaziz (commented):
Wow, lots of changes! I didn't review it all because I don't master skipruntime code enough.

I was thinking reloading could be done automatically by skstore by adding a dependency on the code.

E.g. a mapping from dir/collection `A` using code `f` could be replaced by a mapping from `A` and `f` using a `genericMapper`.
`f` would then be an identifier of code, that can change. Ideally a hash of its semantics (e.g. over-approximated by the hash of the code and the transitive closure of its code and data dependencies), otherwise just a generation number that is incremented whenever the code changes.

Like that, there would be not manual invalidation. What do you think?

skiplabsdaniel:
Reworked with fork and SKStore reuse mechanism

PR #1005, opened 2025-10-17
Add sharding service adapter for Skip Runtime

pikatchu:


bennostein:
This depends on a lot of code not in this repo -- I think we should just keep this over in skipper/backend/reactive

PR #1018, opened 2025-11-12
[skipruntime-ts/server] Send an empty update every 30 seconds on SSE streams

bennostein:
This should help with some timeout issues. I'll also cut an NPM release to get this over into the skipper repo once it lands.

@skiplabsdaniel seem okay to you?

I looked into doing it at different points, but didn't want to mess with watermarks, so this seemed best.


skiplabsdaniel:
> @skiplabsdaniel seem okay to you?
Ok for me

PR #1040, opened 2025-11-17
Improve scheduling of native runtime build vs npm install

jberdine:


beauby (approved):
Thanks! Should we add a comment to the root `README.md` or `HACKING.md` to discourage running `npm install`/`npm run build` directly?

PR #1056, opened 2025-12-02
fix: preserve boolean type in JSON serialization across WASM and nati…

cristianoc:
…ve bindings

Boolean values stored in Skip collections were being returned as numbers (0/1) instead of actual booleans (false/true) when read back via JavaScript.

This fix addresses two issues:

1. WASM binding (skiplang/skjson/ts/wasm/src/skjson.ts): The Skip function SKIP_SKJSON_asBoolean returns Int32 because WASM has no native boolean type - it only supports i32, i64, f32, f64. The TypeScript interface incorrectly declared the return type as `boolean`, but since WASM actually returns i32, JavaScript received a number. Fixed by explicitly converting the i32 result to boolean with `!== 0`.

2. Native addon (skipruntime-ts/addon/src/cjson.cc): CreateCJBool was calling SKIP_SKJSON_createCJFloat instead of SKIP_SKJSON_createCJBool, causing booleans to be stored as floats.

Note: The Skip-side function (SKIP_SKJSON_asBoolean in Extern.sk) explicitly returns Int32 rather than Bool, while SKIP_SKJSON_createCJBool accepts Bool. This asymmetry may have been intentional for FFI stability reasons, or it may be an oversight. The current fix handles the conversion at the binding layer, which is the appropriate place regardless of the original intent, since WASM fundamentally cannot return boolean values.

Adds testBooleanRoundtrip to verify boolean keys are preserved correctly.

### DIFF
--- a/skiplang/skjson/ts/wasm/src/skjson.ts
+++ b/skiplang/skjson/ts/wasm/src/skjson.ts
@@ -29,7 +29,8 @@ export {
 interface WasmAccess {
   SKIP_SKJSON_typeOf: (json: ptr<Internal.CJSON>) => Type;
   SKIP_SKJSON_asNumber: (json: ptr<Internal.CJSON>) => number;
-  SKIP_SKJSON_asBoolean: (json: ptr<Internal.CJSON>) => boolean;
+  // WASM returns i32 (0 or 1), not boolean - conversion happens in WasmBinding
+  SKIP_SKJSON_asBoolean: (json: ptr<Internal.CJSON>) => number;
   SKIP_SKJSON_asString: (json: ptr<Internal.CJSON>) => ptr<Internal.String>;
   SKIP_SKJSON_asArray: (json: ptr<Internal.CJSON>) => ptr<Internal.CJArray>;
   SKIP_SKJSON_asObject: (json: ptr<Internal.CJSON>) => ptr<Internal.CJObject>;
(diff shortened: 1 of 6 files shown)


skiplang/skjson/ts/wasm/src/skjson.ts
```diff
@@ -29,7 +29,8 @@ export {
 interface WasmAccess {
   SKIP_SKJSON_typeOf: (json: ptr<Internal.CJSON>) => Type;
   SKIP_SKJSON_asNumber: (json: ptr<Internal.CJSON>) => number;
-  SKIP_SKJSON_asBoolean: (json: ptr<Internal.CJSON>) => boolean;
+  // WASM returns i32 (0 or 1), not boolean - conversion happens in WasmBinding
+  SKIP_SKJSON_asBoolean: (json: ptr<Internal.CJSON>) => number;
```
jberdine: Would it make more sense to just remove this and change all uses of SKIP_SKJSON_asBoolean to SKIP_SKJSON_asNumber with a nonzero test?

cristianoc: It would make more sense to me, just don't know what's intentional and what's accidental.

jberdine: @skiplabsdaniel, what do you think?

cristianoc:
Removed asBoolean

jberdine (approved):
Nice, thanks.

PR #1061, opened 2026-01-13
Add Iterator#zipSorted and zipSortedBy

jberdine:
Add `zipSorted` operations to merge two iterators that are assumed to be
sorted, producing an iterator indicating which elements appear in only the
left input iterator, only the right, and both.

The `zipSortedBy` variant uses the `Projectable` trait to surface the assumed sortedness in the type.

Runtime assertions check the sortedness assumptions.

### DIFF
--- a/skiplang/prelude/src/core/language/Iterator.sk
+++ b/skiplang/prelude/src/core/language/Iterator.sk
@@ -461,9 +461,21 @@ private mutable class ZipSortedIterator<K: Orderable, T> private (
 
   mutable fun next(): ?(?T, ?T) {
     none: ?T = None();
+    advance = (curr, tail, setHead) -> {
+      curr.each(old -> {
+        newHead = tail.next();
+        newHead.each(new ->
+          invariant(
+            this.project(old) < this.project(new),
+            "Iterator#zipSorted(): iterator not sorted",
+          )
+        );
+        setHead(newHead);
+      });
+    };
     go = (x, y) -> {
-      if (x.isSome()) this.!head1 = this.tail1.next();
-      if (y.isSome()) this.!head2 = this.tail2.next();
+      advance(x, this.tail1, h -> this.!head1 = h);
+      advance(y, this.tail2, h -> this.!head2 = h);
       Some((x, y))
     };
     (this.head1, this.head2) match {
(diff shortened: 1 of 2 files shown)


jberdine:
For review, note that the exposed interface is slightly different, but the functional code is copied from what we already have in skjs.

jberdine:
Also, regarding the exposed interface, perhaps see https://github.com/SkipLabs/skip/pull/1063 for a simplification. It was too fiddly to combine these two PRs, but might be useful when evaluating the interface.

skiplang/prelude/src/core/language/Iterator.sk
```diff
@@ -412,6 +430,65 @@ mutable class ZipWithIterator<T, U, +V>(
...
+  readonly fun sizeHint(): ?Int {
+    size1 = this.tail1.sizeHint().map((n) ->
+      if (this.head1.isSome()) n + 1 else n
+    );
+    size2 = this.tail2.sizeHint().map((n) ->
+      if (this.head2.isSome()) n + 1 else n
+    );
+    (size1, size2) match {
+    | (Some(m), Some(n)) ->
+      Some(m + n) // actual size will be between max and sum of sizes
```
mbouaziz: According to `Iterator#sizeHint`, `Some` should be returned only if the exact number of elements is known

skiplang/prelude/src/core/language/Iterator.sk
```diff
@@ -470,9 +470,21 @@ private mutable class ZipSortedIterator<K: Orderable, T> private (
 
   mutable fun next(): ?(?T, ?T) {
     none: ?T = None();
+    advance = (curr, tail, setHead, id) -> {
+      curr.each(old -> {
+        newHead = tail.next();
+        newHead.each(new ->
+          invariant(
+            this.project(old) < this.project(new),
+            id + " iterator not sorted",
```
mbouaziz: I'm not sure this concatenation is lazy

skiplang/prelude/src/core/language/Iterator.sk
jberdine: You're right, `Vector::mreverseFromIterator` is well-fucked: it throws if the hint isn't perfect. `sizeHint` needs to be renamed to `exactSize`, and this code had to give up on trying to avoid extra allocations.

skiplang/prelude/src/core/language/Iterator.sk
jberdine: To check I understand what you mean: If `advance` is not inlined, then the string concat to build the error message leads to allocating it on every call of `next`? I was expecting `advance` to be inlined, but better to adjust to not be reliant on that.

mbouaziz (approved):

PR #1062, opened 2026-01-13
Add Option#failureOr, dual to Result#maybeFailure

jberdine:


mbouaziz (approved):

jberdine:
FastOption version is also needed, or some instantiations fail to compile.

mbouaziz (approved):

PR #1066, opened 2026-01-26
Make sizeHint a non-exact capacity hint

jberdine:
Update Iterator.sizeHint() documentation to clarify it returns a hint
for capacity pre-allocation rather than an exact count. This allows
consumers to handle cases where the actual count differs.

- Update mreverseFromIterator to gracefully handle inexact hints by
  collecting overflow items when hint is too small, or extracting
  the filled portion when hint is too large
- Improve ZipIterator and ZipWithIterator to return a hint when at
  least one side has a hint
- Improve ZipSortedIterator to return sum of both sizes as upper bound

### DIFF
--- a/skiplang/prelude/src/stdlib/collections/mutable/Vector.sk
+++ b/skiplang/prelude/src/stdlib/collections/mutable/Vector.sk
@@ -99,32 +99,51 @@ mutable class .Vector<+T> private (
   static fun mreverseFromIterator<I: mutable Iterator<T>>(
     items: I,
   ): mutable this {
-    items.sizeHint() match {
-    | Some(size) ->
-      invariant(
-        size >= 0,
-        "Vector::mreverseFromIterator(): Expected items sizeHint() to be nonnegative.",
-      );
-      inner = unsafeMake(size);
-      index = size;
+    hint = items.sizeHint().default(0);
+    if (hint > 0) {
+      // Optimistically try fast path using hint
+      inner = unsafeMake(hint);
+      index = hint;
+      overflow = mutable Vector[];
       items.each(value -> {
-        !index = index - 1;
-        invariant(
-          index.uge(0),
-          "Vector::mreverseFromIterator(): Invalid iterator, expected each() to yield sizeHint() items.",
-        );
-        unsafeSet(inner, index, value);
+        if (index > 0) {
+          !index = index - 1;
+          unsafeSet(inner, index, value);
+        } else {
+          // More items than hint - collect overflow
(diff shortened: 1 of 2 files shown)


skiplang/prelude/src/stdlib/collections/mutable/Vector.sk
```diff
@@ -99,32 +99,51 @@ mutable class .Vector<+T> private (
...
-    | Some(size) ->
-      invariant(
-        size >= 0,
-        "Vector::mreverseFromIterator(): Expected items sizeHint() to be nonnegative.",
-      );
-      inner = unsafeMake(size);
-      index = size;
+    hint = items.sizeHint().default(0);
+    if (hint > 0) {
+      // Optimistically try fast path using hint
```
mbouaziz, line 104: Should we move to a 3-valued or interval-valued hint so that we only take the fast path when the hint is exact?

mbouaziz:
Why not, it affects only `mreverseFromIterator` which is not used.

Could there be a perf regression for `mcreateFromIterator` though? E.g. filtering when the result ends up being way smaller than expected

jberdine:
The only regression for `mcreateFromIterator` I'm thinking of is if the hint is way off on the high side, causing the preallocation of the vector to hit some pathalogical case in malloc where it has to ask the OS for something ridiculous. But otherwise, it will just be one allocation, maybe for too much but that is usually not slower, while before it would have an allocation and then perhaps need to resize if it was too small. In either case, there would be a copy out of the actually used prefix of the vector. Are you thinking of some other scenarios?

jberdine:
By the way, my motivation for making this change is mainly that the existing invariant combined with the name "sizeHint" is a bug trap.

mbouaziz:
> Are you thinking of some other scenarios?

No, something like what you said

mbouaziz (approved):

skiplang/prelude/src/stdlib/collections/mutable/Vector.sk
jberdine: That would be possible. I thought about using an interval, but thought that the cost of allocating pairs or whatever to represent the interval would overwhelm any possible savings from the difference between allocating a small vs large vector. Note that this is just a call to the `unsafe_make` compiler intrinsic, there is no expensive initialization of the contents. Do you think it is worth worrying about?

mbouaziz: Definitely not.

Btw pairs are value classes so no allocation

PR #1067, opened 2026-01-27
AIGEN: Move prettier installation from Dockerfile to Makefile

jberdine:
When prettier is installed globally in the Docker image, it gets whatever
version is current at image build time, which can drift out of sync with
the codebase. By installing via `npm install` in the Makefile, the version
is controlled by package.json and stays in sync with code changes.

- Remove `npm install -g prettier` from bin/apt-install.sh
- Add `npm install --silent` before `npx prettier` in Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@
 default: check check-ts
 
 PRETTIER_LOG_LEVEL?=warn
+PRETTIER_VERSION:=$(shell jq -r '.devDependencies.prettier' package.json)
 
 PLAYWRIGHT_REPORTER?="line"
 SKARGO_PROFILE?=release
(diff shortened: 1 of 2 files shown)


mbouaziz:
Don't we want prettier pre-installed in the Docker image?

jberdine:
I think no, for the reason in the summary. For example, I had to rebuild my image and it picked up a newer prettier, and it wants to change the formatting. Since we rebuild the images at times not synchronized with commits in the repo, that situation is inevitable with prettier installed in the image. Unless we hardcode the version in the Dockerfile and there is some scripting business to ensure that the package.json can only ever have the same version.

mbouaziz:
> Unless we hardcode the version in the Dockerfile and there is some scripting business to ensure that the package.json can only ever have the same version.

That's what I had in mind.

Otherwise the CI can fail because prettier got a new version

jberdine:
Hmm, I was trying to prevent just that. I don't understand npm I guess. Isn't the npm install added by this commit going to install the version in the package-lock.json file?

mbouaziz:
Ah, probably right.

Why can't we keep prettier in sync with our version when we can for e.g. llvm?

jberdine:
If we are super diligent, maybe. But the thing is (if I understand what I have observed correctly) that when prettier is installed in the image, then running `prettier` is ok. But if you happen to run `npx prettier`, then it can give you another version, and make a mess. There isn't anything like that for llvm. I was thinking that not installing prettier in the image would allow having the version specified in just one place, and that was preferable to having some scripting hackery to keep the multiple version specifications in sync.

mbouaziz:
JS world...

Go!

PR #1070, opened 2026-01-29
Json type: use readonly arrays and objects

mbouaziz:


skiplang/skjson/ts/binding/src/index.ts
```diff
@@ -139,6 +139,13 @@ export function deepFreeze<T>(value: T): T & DepSafe {
   }
 }
 
+/* Fix to get Array.isArray narrow readonly arrays */
+declare global {
+  interface ArrayConstructor {
+    isArray(arg: any): arg is any[] | readonly any[];
+  }
+}
```
jberdine, line 147: FTR I don't understand this, any hint?

mbouaziz: Typescript has, in `lib.es5.d.ts`:
```ts
interface ArrayConstructor {
    ...
    isArray(arg: any): arg is any[];
}

declare var Array: ArrayConstructor;
```

so checking `Array.isArray` narrows to **mutable** arrays only, which gives type errors later on (e.g. in `deepEquals`) and is also a non-`as` way of converting a readonly array to a mutable array.

This fix at least allows `Array.isArray` checks to keep the readonly array part of `Json`

jberdine: Gee, thanks TS....

jberdine (approved):

PR #1072, opened 2026-02-02
AIGEN: Pin exact prettier version in package.json

jberdine:
## Summary
- Remove caret from version specifier so the Makefile uses an exact version (3.4.1) rather than a compatible range (^3.4.1)
- Update example package.json files to use the same version
- Update the git hook to use the version from package.json

### DIFF
--- a/bin/git_hooks/check_format.sh
+++ b/bin/git_hooks/check_format.sh
@@ -4,6 +4,8 @@
 
 exec 1>&2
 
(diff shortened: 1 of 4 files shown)


mbouaziz (approved):

mbouaziz:
Could be worth ensuring all package.json use the same version.
And that the hook is using it as well

PR #1076, opened 2026-02-06
AIGEN: Fix #988: Forbid underscore-prefixed variable values

### LINKED ISSUE #988: Segfault/incorrect results for overlapping patterns with underscore-prefixed variables
I ran into this when using some variable names that collided with keywords. (wrote `class`, `static`, etc. as `_class`, `_static`, etc.) Easy workaround is to just use underscore suffixes instead of prefixes to dodge the keywords. But this should either be supported properly or throw a compiler error, as with some other uses of underscore-prefixed variables.

Seems like primitives and pointers get zeroed out when bound into underscore-prefixed variables in such patterns, causing incorrect results for primitives and segfaults for pointers.

Repro as follows:
```
base class T {
  children =
  | T1(x: Int, y: Int, z: Int)
  | T2(x: Int, y: Int, z: String)
  fun okay(): Int {
    this match {
    | T1(x, y, _)
    | T2(x, y, _) ->
      x + y
    }
  }

  fun broken(): Int {
    this match {
    | T1(_x, _y, _)
    | T2(_x, _y, _) ->
      _x + _y
    }
  }
}
base class V {
  children =
  | V1(x: String, y: String, z: Int)
  | V2(x: String, y: String, z: String)
  fun okay(): String {
    this match {
    | V1(x, y, _)
    | V2(x, y, _) ->
      x + y
    }
  }

  fun broken(): String {
    this match {
    | V1(_x, _y, _)
    | V2(_x, _y, _) ->
      _x + _y
    }
  }
}

fun main(): void {
  t1 = T1(1, 2, 3);
  t2 = T2(1, 2, "three");
  print_string(t2.okay().toString());
  print_string(t2.broken().toString());
  print_string(t1.okay().toString());
  print_string(t1.broken().toString());
  v1 = V1("one", "two", 3);
  v2 = V2("one", "two", "three");
  print_string(v2.okay());
  print_string(v2.broken());
  print_string(v1.okay());
  print_string(v1.broken()); // SEGFAULT if uncommented
}
```

Output:

```
$ skc repro.sk -o repro
$ ./repro
3
3
3
0
onetwo
onetwo
onetwo
[1]    26726 segmentation fault  ./repro
```

mbouaziz:
## Summary

- Fix issue #988: Segfault/incorrect results for overlapping patterns with underscore-prefixed variables
- Names starting with `_` no longer bind in patterns (they act as wildcards)
- Attempting to use a `_`-prefixed variable value produces a clear error message
- Underscore-prefixed struct fields are still allowed - use explicit binding syntax (`_field => var`) if you need the value

## Test plan

- [x] Added repro test case from issue #988 as `tests/syntax/invalid/under_ident_3`
- [x] All 1699 compiler tests pass
- [x] Builds successfully with `make STAGE=1`

Fixes #988

🤖 Generated with [Claude Code](https://claude.ai/code)


skiplang/compiler/src/convertTree.sk
```diff
@@ -1383,27 +1383,27 @@ class Converter{file: FileCache.InputSource} {
           range,
           SkipAst.Pat_type(
             SkipAst.Tid_object(SkipAst.Tclass((range, "Success"))),
-            Some(Positional(Array[(range, SkipAst.Pat_var((range, "__x")))])),
+            Some(Positional(Array[(range, SkipAst.Pat_var((range, "!x")))])),
```
bennostein, line 1386: Is this now using `"!x"` as an identifier? surprising that we allow that!  Or is this conversion pass already downstream of where that gets checked?

bennostein (approved):

PR #1089, opened 2026-02-13
AIGEN The test framework now reports when a subprocess failed

pikatchu:


mbouaziz:
Duplicate of #1088 ?

mbouaziz (commented):
Hey Julien, I'm Claude (working with Mehdi). We have a parallel PR at #1088 tackling the same issue — here are a couple of things worth noting:

1. **`wait()` calls `closefds()`** — you'll want to read `p.stderr` *before* calling `p.wait()`, otherwise the stderr fd is closed and any crash output is lost.

2. **Per-test vs per-worker errors** — this PR reports one synthetic failure per crashed worker. If a worker had 50 tests and crashed after test 30, you'd see a single generic error. In #1088 we track expected vs reported tests so each unreported test gets its own error result (with file/line info), which makes it easier to find what crashed.

3. **Test coverage** — #1088 includes an automated test (`test_crash_detection.sk`) that spawns the harness with a crashing test and asserts non-zero exit, to prevent regressions.

mbouaziz:
Subsumed by merged #1088

PR #1109, opened 2026-02-20
Allow undefined fields on JsonObject

skiplabsdaniel:
Today, defining an object with optional values involves the systematic use of the null value or defining the union of all versions of the type.

### DIFF
--- a/skiplang/skjson/ts/binding/src/index.ts
+++ b/skiplang/skjson/ts/binding/src/index.ts
@@ -167,7 +167,7 @@ export type Json =
 /**
  * Objects containing `Json` values.
  */
-export type JsonObject = { readonly [key: string]: Json | null };
+export type JsonObject = { readonly [key: string]: Json | null | undefined };
 
 export type Exportable =
   | null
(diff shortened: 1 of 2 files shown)


mbouaziz (changes requested):
## Review

Note: this review was written entirely by me (Claude Code), not by Mehdi.

The `defaultParamEncoder` fix is correct, but widening the core `JsonObject` type has ripple effects that aren't addressed here.

### Concerns with adding `undefined` to `JsonObject`

- **`deepFreeze()`** iterates `Object.values()` and will throw on `undefined` values since it doesn't handle that type.
- **`JSON.stringify()` silently drops `undefined`** — `ObjectHandle.toString()` uses `JSON.stringify(this.toJSON())`, so `undefined` properties would be silently omitted from serialization, making roundtrips lossy.
- **`exportJSON()` converts `undefined` → CJSON null** — meaning `undefined` values become `null` when exported to the Skip runtime. This is a silent lossy conversion.
- **`clone()`** — `Object.entries()` skips `undefined` values, so cloning a `JsonObject` with `undefined` properties would drop them.

### Only one consumer is updated

The PR updates `defaultParamEncoder` but other places that iterate over `JsonObject` values would need similar guards or would silently misbehave.

### Suggestion

If the goal is to support optional fields, a narrower approach (e.g. a separate `PartialJsonObject` type for specific use cases) might be safer than changing the foundational type the entire runtime relies on.

PR #1115, opened 2026-02-23
skc/skargo: separate compilation from linking to reduce peak memory

### LINKED ISSUE #1114: skc: high memory usage during clang invocation causes OOM on 4GB CI
## Problem

When `skc` compiles a binary with `-O2`, it peaks at ~4.3GB total memory because it holds ~3.4GB RSS while spawning `clang -cc1 -O2` which adds ~0.9GB on top. This causes OOM kills on CI environments with 4GB RAM (e.g. CircleCI medium).

## Measured Memory Profile

Captured building the skjs binary (`skargo build --release --bin skjs`) inside `skiplabs/skip:latest`:

| Time | Process | RSS |
|------|---------|-----|
| T+0s | skc starts | ~200MB |
| T+10s | skc (compilation) | ~1.8GB |
| T+15s | skc (peak, pre-clang) | **~3.4GB** |
| T+15s | skc spawns clang -cc1 | skc: 3.4GB + clang: ~270MB |
| T+25s | **PEAK** | skc: 3.4GB + clang: ~940MB = **~4.3GB total** |
| T+35s | done | released |

### Key observations

- `skc` holds **3.4GB RSS** the entire time clang runs
- `clang -cc1` grows to ~940MB during LLVM codegen at `-O2`
- The two processes overlap: skc doesn't free memory before clang finishes

## Suggested Fix

Since skc's job is done once it writes the `.ll` file, it could:

1. **`exec()` into clang** — replace the skc process entirely, reducing peak to max(3.4GB, 0.9GB) = 3.4GB
2. **Free compilation state** before invoking clang
3. **Fork/exec clang and exit** — let clang run independently

Option 1 would be the most effective.

## Reproduction

```bash
git clone https://github.com/SkipLabs/skjs
git clone --recurse-submodules https://github.com/SkipLabs/skip

# This reliably OOMs:
docker run -d --name skc-oom-test --memory=4g --memory-swap=4g \
  -v "$PWD/skjs:/work/skjs" -v "$PWD/skip:/work/skip" \
  -w /work/skjs skiplabs/skip:latest \
  skargo build --release

docker inspect skc-oom-test --format '{{.State.OOMKilled}}'  # true = OOM
```

## Current Workaround

skjs CI uses `skargo build --release --bin skjs` instead of `skargo build --release` to avoid building the test binary (which causes a second 4GB spike). This brings peak from ~4.3GB to ~3.9GB.

## Environment

- Docker image: `skiplabs/skip:latest` (skc 0.1.0, LLVM 20)
- CI: CircleCI medium executor (4GB RAM)

mbouaziz:
## Summary
- Refactors `link()` into `getLinkArgs()` that returns the link command without executing it
- Adds `--no-link` flag to skc: compiles as usual but prints link args between markers instead of invoking clang++
- skargo passes `--no-link` for binary/test/cdylib targets and invokes the linker itself

This separates compilation from linking so skc (~3.4GB) and clang (~0.9GB) never coexist in memory. Peak memory drops from ~4.3GB to ~3.4GB, fixing OOM on 4GB CI environments.

## Test plan
- [x] 1701 compiler tests pass
- [x] Verified `--no-link` output contains correct markers and link args
- [x] skjson builds cleanly with new skargo+skc
- [x] Measured peak memory: old=768MB combined, new=max(757MB, 149MB)=757MB on skjson

Fixes #1114

### DIFF
--- a/skiplang/compiler/src/Config.sk
+++ b/skiplang/compiler/src/Config.sk
@@ -68,6 +68,7 @@ private class Config{
   // binary sizes.
   useSpecializedNames: Bool,
   check: Bool,
+  no_link: Bool,
   stackAlign: Int,
   stackAlignStr: String,
   ptrByteSize: Int,
@@ -183,6 +184,7 @@ private class Config{
     sampleRate = results.getString("sample-rate").toInt();
     useSpecializedNames = results.getBool("use-specialized-names");
     check = results.getBool("check");
+    no_link = !results.getBool("link");
 
     exportAs = mutable UnorderedMap[];
     for (f in exportedAsFunctions) {
@@ -261,6 +263,7 @@ private class Config{
       sampleRate,
       useSpecializedNames,
       check,
+      no_link,
       stackAlign,
       stackAlignStr,
       ptrByteSize,
--- a/skiplang/compiler/src/compile.sk
+++ b/skiplang/compiler/src/compile.sk
@@ -405,9 +405,15 @@ fun compile_binary(
   });
 
   runCompilerPhase("native/link", () -> {
-    getLinkArgs(llFile, conf, exports).each(args ->
-      runShell(args, conf.verbose)
-    )
+    getLinkArgs(llFile, conf, exports).each(args -> {
+      if (conf.no_link) {
+        print_string("---LINK_ARGS_BEGIN---");
+        args.each(print_string);
+        print_string("---LINK_ARGS_END---")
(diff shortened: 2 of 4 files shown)


mbouaziz:
Better: add an option to `skc` to not "link" but only output the linker arguments.
By default `skargo` would invoke `skc` with this option and itself invoke the linker.

PR #1116, opened 2026-02-23
Add FileSystem.stat: single-call POSIX stat with FileKind ADT

mbouaziz:
## Summary
- Add `FileSystem.stat(path)` that calls POSIX `stat()` once and returns all fields (mode, size, mtime, atime, ctime) as a `FileStat` object, or `None()` if the path doesn't exist
- Add `FileKind` ADT (FKRegular, FKDirectory, FKSymlink, etc.) with `kind()` and `permissions()` methods for interpreting the mode field
- Add C runtime function `SKIP_file_stat` backing the native call

## Test plan
- [ ] Verify `FileSystem.stat` returns correct metadata for files, directories, and symlinks
- [ ] Verify `None()` is returned for non-existent paths
- [ ] Verify `kind()` correctly classifies file types
- [ ] Verify `permissions()` extracts the expected permission bits

### DIFF
--- a/skiplang/prelude/runtime/runtime64_specific.cpp
+++ b/skiplang/prelude/runtime/runtime64_specific.cpp
@@ -514,6 +514,29 @@ bool SKIP_is_directory(char* path) {
   return st.st_mode & S_IFDIR;
 }
 
+// Must match FileStat field layout: mode, size, mtime, atime, ctime
+typedef struct {
+  int64_t mode;
+  int64_t size;
+  int64_t mtime;
+  int64_t atime;
+  int64_t ctime;
+} sk_file_stat_t;
+
+int64_t SKIP_file_stat(sk_file_stat_t* result, char* path) {
+  struct stat st;
+  int rc = stat(path, &st);
+  if (rc < 0) {
+    return rc;
+  }
+  result->mode = st.st_mode;
+  result->size = st.st_size;
+  result->mtime = st.st_mtime;
+  result->atime = st.st_atime;
+  result->ctime = st.st_ctime;
+  return 0;
+}
+
 int64_t SKIP_system(char* cmd) {
   sk_string_check_c_safe(cmd);
   int64_t res = system(cmd);
--- a/skiplang/prelude/src/stdlib/filesystem/FileSystem.sk
+++ b/skiplang/prelude/src/stdlib/filesystem/FileSystem.sk
@@ -12,6 +12,48 @@
 @may_alloc
 native fun getcwd(): String;
 
+base class FileKind uses Orderable {
+  children =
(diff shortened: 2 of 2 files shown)


skiplang/prelude/runtime/runtime64_specific.cpp
```diff
@@ -514,6 +514,28 @@ bool SKIP_is_directory(char* path) {
...
+  int64_t atime;
+  int64_t ctime;
+} sk_file_stat_t;
+
+void SKIP_file_stat(sk_file_stat_t* result, char* path) {
+  struct stat st;
+  if (stat(path, &st) < 0) {
+    result->mode = 0;
+    return;
+  }
```
jberdine: fillStat needs to understand the implicit error condition protocol, so it seems to me that it would be better to have this function return the status code of the stat call instead of void, and then the skip code could report a more informative error.

skiplang/prelude/src/stdlib/filesystem/FileSystem.sk
```diff
@@ -12,6 +12,48 @@
...
+  // Extract file type from mode.
+  fun kind(): FileKind {
+    this.mode.and(0o170000) match {
+    | 0o100000 -> FKRegular()
+    | 0o040000 -> FKDirectory()
+    | 0o120000 -> FKSymlink()
+    | 0o020000 -> FKCharDevice()
+    | 0o060000 -> FKBlockDevice()
+    | 0o010000 -> FKFifo()
+    | 0o140000 -> FKSocket()
```
jberdine: It would be clearer to test these in decreasing order.

skiplang/prelude/src/stdlib/filesystem/FileSystem.sk
```diff
@@ -12,6 +12,48 @@
...
+    | _ -> FKUnknown()
+    }
+  }
+
+  // Extract permission bits (lower 12 bits of mode).
+  fun permissions(): Int {
+    this.mode.and(0o7777)
```
jberdine: Note that the low 12 bits are the "file mode bits" while only the low 9 are the "file permission bits".

mbouaziz:
Applied all review feedback:

1. **Return stat status code** — `SKIP_file_stat` now returns `int64_t` (the `stat()` return code) instead of `void`. The Skip side checks `rc < 0` instead of inspecting `mode == 0`.
2. **Reordered `kind()` match arms** — now in decreasing octal order (0o140000 → 0o010000).
3. **Fixed `permissions()`** — mask changed from `0o7777` (12-bit file mode bits) to `0o777` (9-bit file permission bits).

— Claude

jberdine (approved):

PR #1120, opened 2026-02-24
skargo: use per-unit state.db to avoid cache thrashing

### LINKED ISSUE #1119: Skargo: lib and test targets trample on each other's cache

cristianoc:
Lib and test targets sharing the same build directory and relocation model would read/write the same state.db, causing cache thrashing. Include the package name and unit hash in the state.db filename to make it unique per compilation unit.

Fix #1119

### DIFF
--- a/skiplang/skargo/src/build_runner.sk
+++ b/skiplang/skargo/src/build_runner.sk
@@ -391,11 +391,16 @@ mutable class BuildRunner(
     if (!unit.target.is_build_script()) {
       // TODO: Once we have namespaced packages, we can share the state.db
       // across architectures.
+      // Also split by unit to avoid cache thrashing between lib and test
+      // targets that share the same build directory.
+      unit_hash = this.hashes[unit].toStringHex();
       state_db_path = Path.join(
         this.layout_for(unit).build,
         // FIXME: This is a dirty hack that splits out the state.db between pic
         // and static, which is required until `skc` namespaces packages.
-        `${unit.build_opts.relocation_model}_state.db`,
+        `${unit.pkg.name()}-${unit_hash}_${
+          unit.build_opts.relocation_model
+        }_state.db`,
       );
       skc.args(
         Array[


beauby:
My understanding is that the lib and test targets (for the same build profile) should be able to share the same cache, because the cache for the test target should be a superset of the cache for the lib target. Is building `lib` after `test` causing pessimistic evictions?

cristianoc:
> My understanding is that the lib and test targets (for the same build profile) should be able to share the same cache, because the cache for the test target should be a superset of the cache for the lib target. Is building `lib` after `test` causing pessimistic evictions?

The observation is that `cargo check` is significantly slower when a single cache file is used.
Not sure if it matters, but the test target has additional source files that are not present in lib.

beauby:
@cristianoc Do you observe this behavior for all projects or for one in particular?

beauby:
Might be related: #971.

jberdine:
For me, just `make check` in the skip repo really demonstrates that skargo is not reusing cached work. See the gist https://gist.github.com/jberdine/9228384b34619cbb4756c36d7437a664 . It shows that over and over the same things are "Compiling" rather than "Fresh". Not saying it is a priority, but just reporting what I see.

mbouaziz:
> For me, just `make check` in the skip repo really demonstrates that skargo is not reusing cached work. See the gist https://gist.github.com/jberdine/9228384b34619cbb4756c36d7437a664 . It shows that over and over the same things are "Compiling" rather than "Fresh". Not saying it is a priority, but just reporting what I see.

The cache is per project where you build, not per built library. So each library is rebuilt where it's used. But there, there is a cache

skiplang/skargo/src/build_runner.sk
```diff
@@ -391,11 +391,16 @@ mutable class BuildRunner(
     if (!unit.target.is_build_script()) {
       // TODO: Once we have namespaced packages, we can share the state.db
       // across architectures.
+      // Also split by unit to avoid cache thrashing between lib and test
```
mbouaziz, line 394:
```suggestion
      // Also split by unit to avoid cache thrashing between bin and test
```

PR #1123, opened 2026-02-27
skargo: merge bin and test into a single binary in dev mode

### LINKED ISSUE #1121: Faster iteration: merge bin and test
At least in dev mode, we could always integrate the tests in the binary and branch depending on a command line argument or an environment variable. This way, `skargo build`, and `skargo test` would share the same binary.

mbouaziz:
## Summary

Closes #1121.

In dev mode, when a package has `[[bin]]` targets, all binaries and the test target are merged into a **single compiled binary** (`__merged`) with a dispatcher that routes to the correct entry point at runtime. Other binaries and the test binary become hardlinks to `__merged`.

### How it works

- A new `HardlinkTarget(source)` target kind is added — it creates a copy of another binary without compilation
- In dev mode with bins, `generate_root_units()` creates a synthetic `__merged` `BinTarget` and replaces all bin/test targets with `HardlinkTarget("__merged")`
- The **lib** is augmented with test sources and a generated dispatcher function (`SkipInternal.dispatch`), so `__merged` links against it like any normal bin
- The dispatcher checks `SKARGO_BIN` env var (or falls back to `argv[0]` basename) to route to the correct entry point
- `skargo test` / `skargo run` / `skargo check` all use the default filter in dev mode to trigger merged compilation
- Release mode and lib-only packages are unaffected

### Key design decisions

- **Augmented lib approach**: the lib is the single compilation unit containing src/ + tests/ + dispatcher in dev mode. `__merged` is a normal empty-source bin that links against it, matching how regular bins work
- **Dev deps on lib**: in dev mode, the root package's lib pulls `DevelopmentDep` dependencies (e.g., sktest) since it now includes test sources
- **Multi-bin support**: dispatcher handles any number of bins + optional test target

## Files changed

- **manifest.sk** — Added `HardlinkTarget(source: String)` to `TargetKind`
- **build_context.sk** — `generate_root_units()` creates `__merged` + hardlinks in dev mode; `compute_deps()` handles `HardlinkTarget` (no deps) and lib dev deps
- **build_runner.sk** — `generate_dispatcher()` method; `skc()` augments lib with test sources + dispatcher in dev mode; `compile()` and `outputs_for()` handle `HardlinkTarget`
- **commands/test.sk** — Uses default filter in merged mode, finds test as `HardlinkTarget`, sets `SKARGO_BIN=test`
- **commands/run.sk** — Sets `SKARGO_BIN` in dev mode for dispatcher routing
- **commands/check.sk** — Uses default filter in merged mode

## Test plan

- [x] `skargo check` on skargo itself passes at each commit
- [x] `skargo build` on a bin+tests project produces `__merged` + hardlinked `hello` and `test`
- [x] `skargo run` routes to user's `main()` via dispatcher
- [x] `skargo test` runs tests via dispatcher (shows "Fresh", ~0.1s on subsequent runs)
- [x] `skargo check` in dev mode checks everything in a single merged pass
- [x] No dependency cycles (dev deps restricted to root package's lib)
- [ ] Release mode falls back to separate binaries
- [ ] Compiler tests pass

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skiplang/skargo/src/build_runner.sk
+++ b/skiplang/skargo/src/build_runner.sk
@@ -181,8 +181,19 @@ mutable class BuildRunner(
 
       // Hardlink targets don't compile — just create the link.
       if (unit.target.kind is HardlinkTarget _) {
-        this.prepare_unit(unit);
-        this.link_targets(unit);
+        if (!check) {
+          this.prepare_unit(unit);
+          this.link_targets(unit)
+        };
+        continue
+      };
+
+      // In check mode, skip bin roots — the lib check covers all code.
+      if (
+        check &&
+        unit.target.kind is BinTarget _ &&
+        this.bctx.roots.contains(unit)
+      ) {
         continue
       };
 
--- a/skiplang/skargo/src/build_runner.sk
+++ b/skiplang/skargo/src/build_runner.sk
@@ -600,6 +600,66 @@ mutable class BuildRunner(
     }
   }
 
+  private mutable fun generate_dispatcher(unit: Unit): String {
+    dir = Path.join(this.layout_for(unit).build, "dispatch");
+    this.create_dir(dir);
+    dispatcher_path = Path.join(dir, "__skargo_dispatch.sk");
+
+    bins = unit.pkg.manifest.targets.filter(t -> t.is_bin()).collect(Array);
+    test_entry_opt = unit.pkg.manifest.test_target().map(t ->
+      t.kind match {
+      | TestTarget(e) -> e
+      | _ -> invariant_violation("expected TestTarget")
+      }
+    );
+
+    lines = mutable Vector[];
+    lines.push("module SkipInternal;");
+    lines.push("untracked fun dispatch(): void {");
+    lines.push("  name = Environ.varOpt(\"SKARGO_BIN\") match {");
+    lines.push("  | Some(n) -> n");
+    lines.push("  | None() -> Path.basename(Environ.args().next().fromSome())");
+    lines.push("  };");
+    lines.push("  name match {");
+    for (bin in bins) {
+      bin.kind match {
+      | BinTarget(entry) -> lines.push(`  | "${bin.name}" -> ${entry}()`)
+      | _ -> invariant_violation("expected BinTarget")
+      }
+    };
+    test_entry_opt.each(test_entry ->
+      lines.push(`  | "test" -> ${test_entry}()`)
+    );
+    if (bins.size() == 1 && test_entry_opt.isNone()) {
+      bins[0].kind match {
+      | BinTarget(entry) -> lines.push(`  | _ -> ${entry}()`)
+      | _ -> invariant_violation("expected BinTarget")
+      }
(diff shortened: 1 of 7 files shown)


mbouaziz:
We don't have to limit to 1 binary: we can dispatch between the binaries similarly as we dispatch between the bin and the test

skiplang/skargo/src/build_runner.sk
```diff
@@ -558,6 +584,26 @@ mutable class BuildRunner(
...
+    dir = Path.join(this.layout_for(unit).build, "dispatch");
+    this.create_dir(dir);
+    dispatcher_path = Path.join(dir, "__skargo_dispatch.sk");
+    content = Array[
+      "fun __skargo_dispatch(): void {",
+      `  Environ.varOpt("SKTEST") match {`,
+      `  | Some _ -> ${test_entry_point}()`,
+      `  | None() -> ${user_entry_point}()`,
+      "  }",
+      "}",
```
mbouaziz: Namespace this in a module `SkipInternalWhatever`

mbouaziz:
## Update: rewritten with augmented lib approach

The branch has been rewritten from scratch on top of `main`. Key changes since the original PR:

### Architecture change
Previously, `__merged` was a `BinTarget` that contained all bin + test sources directly. Now, the **lib** is the single compilation unit — it gets augmented with test sources and a generated dispatcher in dev mode. `__merged` is a normal empty-source bin that links against the lib, matching how regular bins already work.

### New commits (8 total)
1. **Add `HardlinkTarget` variant** to `TargetKind`
2. **Handle `HardlinkTarget` in build pipeline** — skip `skc()`, just link outputs
3. **Add dispatcher generation** — `generate_dispatcher()` method + `HardlinkTarget` in `compute_deps`
4. **Merged binary mode** — `generate_root_units` creates `__merged` + hardlinks; lib gets test sources + dispatcher; lib pulls dev deps in dev mode
5. **`skargo test`** — uses default filter in merged mode, sets `SKARGO_BIN=test`
6. **Simplifications** — minor cleanup
7. **`skargo run`** — sets `SKARGO_BIN` for dispatcher routing
8. **`skargo check`** — uses merged mode in dev builds too

### What's different from v1
- Dev deps restricted to root package's lib only (fixes dependency cycle with transitive deps)
- `skargo check` also benefits from merged mode
- `TestTarget` checked before `HardlinkTarget` in test command
- Each commit passes `skargo check` on skargo itself
- Tested on toy project: `build`, `run`, `test`, `check` all work correctly

beauby:
> * Dev deps restricted to root package's lib only (fixes dependency cycle with transitive deps)

Doesn't this defeat the purpose of dev dependencies?

mbouaziz:
> > * Dev deps restricted to root package's lib only (fixes dependency cycle with transitive deps)
> 
> Doesn't this defeat the purpose of dev dependencies?

Not sure what Claude meant here.

There's still a difference between non-dev and dev-dependencies in release mode

beauby (approved):

PR #1125, opened 2026-03-03
Several changes needed to fix the build

jberdine:
The recent churn broke things for me.
Not sure exactly what made us hit some of these now and not before.


mbouaziz (approved):
Thanks for the first fix

Not sure I get the libbacktrace fix, I'd like to avoid building it all the time

jberdine:
I split off the libbacktrace commit and will merge the others.

PR #1131, opened 2026-03-05
Avoid libbacktrace source directory pollution

jberdine:
The prelude Makefile configured libbacktrace in-tree, leaving
config.status and .lo files in the source directory. The compiler
Makefile invokes the prelude Makefile with a different OUT_DIR, so the
.a goes to the compiler's build dir while the source dir stays dirty.
When skargo later runs build.sk, the out-of-tree configure fails
because autoconf detects the dirty source directory.

Three fixes, each addressing a different build path:

- Prelude Makefile: configure and build in $(OUT_DIR)/libbacktrace-build/
  instead of in-tree, so the source directory stays clean regardless of
  which OUT_DIR is passed (bootstrap, compiler, or prelude); clean any
  stale in-tree artifacts before configuring
- build.sk: before building from source, check for a pre-built
  ./build/libbacktrace.a from the Makefile and reuse it; the fallback
  source build also uses out-of-tree configure
- Top-level Makefile: add a libbacktrace target and make check/test
  depend on it, so libbacktrace is built once before skargo runs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,12 @@ build/init.sql: sql/privacy/init.sql
 update-js-deps:
 	find . -name node_modules -not -prune -or -name target -not -prune -or -name package.json -exec sh -c 'cd $$(dirname "$$0"); bun update --latest' {} \;
 
+.PHONY: libbacktrace
+libbacktrace:
+	$(MAKE) -C skiplang/prelude libbacktrace
+
 .PHONY: check
-check:
+check: libbacktrace
 	find * -name Skargo.toml -exec sh -c 'bin/cd_sh $$(dirname {}) "skargo check"' \;
 
 .PHONY: check-ts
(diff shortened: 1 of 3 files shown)


jberdine:
@mbouaziz is this explanation clearer? Without this change, `make clean check` in the skip repo usually fails for me.

mbouaziz (approved):

jberdine:
This is now taking a different approach to ensure that libbacktrace is built just once before anything else, and so could use another round of review.

mbouaziz:
What commands do you want me to run on my machine to check it works fine?

jberdine:
Beyond what the CI does, `make clean; make check; make check` has for me uncovered the issues.

mbouaziz:
Works for me:
`make clean` -> failed because of permissions (previous build was probably in a container)
`sudo make clean` :heavy_check_mark: 
`make check` :heavy_check_mark: 
`make check` :heavy_check_mark: finishes in 5.5s

Works as well in a devcontainer based on `skiplabs/skip:latest`

mbouaziz (approved):

PR #1162, opened 2026-03-19
make skipruntime-ts api type-safe and align service bindings

cristianoc:
Tighten the skipruntime-ts API types and update the runtime bindings to match the new input definitions used when creating services.

- make core API inputs and service creation flows type-safe
- adapt addon, wasm, server, and helper bindings to the new input definitions
- update examples and tests to use the revised service creation API

### DIFF
--- a/bin/check-ts.sh
+++ b/bin/check-ts.sh
@@ -8,7 +8,13 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 REPO="$SCRIPT_DIR/.."
 
 cd "$REPO"
-DIRS=$( jq --raw-output ".workspaces[]" package.json )
+# Exclude the top-level examples/* apps: check-examples builds them in Docker
+# against the published @skipruntime/core, not the local workspace, so
+# type-checking them here against local source would impose two mutually
+# exclusive API versions on the same file whenever the two diverge.
+# skipruntime-ts/examples is unaffected (it doesn't start with "examples/")
+# and has no other type-checking coverage, so it stays included.
+DIRS=$( jq --raw-output '.workspaces[] | select(startswith("examples/") | not)' package.json )
 
 LINT=""
 if [ -z "${SKIP_LINT:-}" ]; then
--- a/examples/blogger/reactive_service/src/blogger.service.ts
+++ b/examples/blogger/reactive_service/src/blogger.service.ts
@@ -1,10 +1,10 @@
-import type {
-  Context,
-  EagerCollection,
-  Json,
-  Values,
-  Resource,
-  SkipService,
+import {
+  type Context,
+  type EagerCollection,
+  type Json,
+  type Values,
+  type Resource,
+  type AnySkipService,
 } from "@skipruntime/core";
 
 import { PostgresExternalService } from "@skip-adapter/postgres";
@@ -104,8 +104,8 @@ class PostsResource implements Resource<PostsResourceInputs> {
 
 type PostsServiceInputs = Record<string, never>;
 
-export const service: SkipService<PostsServiceInputs, PostsResourceInputs> = {
-  initialData: {},
+export const service: AnySkipService = {
+  inputs: {},
   resources: { posts: PostsResource },
   externalServices: { postgres },
   createGraph(
--- a/examples/blogger/reactive_service/src/blogger.service.ts
+++ b/examples/blogger/reactive_service/src/blogger.service.ts
@@ -1,10 +1,10 @@
-import {
-  type Context,
-  type EagerCollection,
-  type Json,
-  type Values,
-  type Resource,
-  type AnySkipService,
+import type {
+  Context,
+  EagerCollection,
+  Json,
+  Values,
+  Resource,
+  SkipService,
 } from "@skipruntime/core";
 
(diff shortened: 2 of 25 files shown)


jberdine:
This is now rebased, with a few additional changes, and a testing change to keep some of the examples from being checked twice against different api sources.

This won't be the end of the story, but is a good step that I think can be landed now.

mbouaziz:
Reviewed the full diff, with the type-level claims below verified by running tsc against this branch and the runtime claims traced through the wasm/addon binding paths. Overall: the variance design is right, the migration is thorough across bindings/examples/tests, CI is green, and I agree it's landable as a step. Findings are split into a stacked hardening PR (#1482) and two follow-up issues (#1480, #1481); details below.

### What checks out

- The sound-variance mechanics are correct and verified: `readonly` index signatures make the collection maps covariant, and `EagerCollection<K, V>` → `AbstractEagerCollection` assignability works as intended.
- The brand fields are **zero-runtime-cost**: under the repo's ES2021 target the uninitialized field declarations emit nothing — the abstract classes compile to empty bodies, so no own-property pollution on instances.
- The deprecated aliases (`NamedCollections`, `InitialData`) have zero remaining load-bearing uses in `src` anywhere — pure transition surface, safe to delete in a later pass.
- The `check-ts.sh` exclusion of the 4 top-level example apps is genuinely forced, not just convenient: they pin `@skipruntime/*@0.0.23`, but since the local workspaces are *also* 0.0.23, root `npm install` symlinks the local (new-API) source over the registry pins — so in-repo type-checking of those apps cannot pass while the APIs diverge.
- No version bump needed here per repo practice (bump-at-publish), and `release_npm.sh` refuses same-version republish, so the breaking API can't accidentally ship as 0.0.23.

### Main design note → #1480

The PR loses a compile-time guarantee the old API had at typed use sites: `initialData?: InitialData<Inputs>` used to reject a wrong key or wrong `Entry<K, V>` type against the `createGraph` parameter type. In the new API `InputDefs` and `Inputs` are unrelated type parameters, so this compiles cleanly (tsc-verified):

```ts
const s: AnySkipService = {
  inputs: { input: new InputDefinition<string, number>() },
  createGraph(inputs: { WRONGNAME: EagerCollection<string, number> }) { ... }, // lies; undefined at runtime
  ...
};
```

— and method-declaration bivariance means even arbitrarily specific wrong annotations are accepted. Since every example/test in the PR now uses `AnySkipService`, use sites are in practice *less* checked than before for a PR whose goal is type safety. The good news: an `InputCollections<D>` mapped type restores the link at zero runtime cost and composes with the sound-variance design (tsc-verified, including as a defaulted type parameter). Repro details and the proven fix candidate are in #1480, along with the related `ResourceClass<RI, never>` caveat (constructors with **non-Json** params are accepted statically and get `Json` at runtime).

### Runtime hardening → #1482 (stacked on this PR)

Two failure modes get cryptic errors today, since nothing validates at the JS boundary:

- A plain-JS consumer still passing `{ initialData: ... }` fails deep in native init with `TypeError: Cannot convert undefined or null to object` — no migration hint (`ServiceDefinition.inputs()` hits `Object.keys(undefined)`).
- The brands are structural (a hand-written `{ __sk_collectionBrand: undefined }` type-checks as a collection, ditto `__sk_inputDefBrand`), and forged objects reach `EagerCollectionImpl.getName` / `initialData`'s blind `InputDefinition` cast, producing backend-dependent crashes (`Cannot read properties of undefined (reading 'length')` / `Invalid function return type` / skjson `Must be a Array value.`) that never name the real mistake.

#1482 adds fail-fast guards with actionable messages (+ tests for both backends), removes a redundant cast, makes `InputDefinition` accept `readonly` initial data, and refreshes the stale `api.ts` doc comments (the `SkipService` overview still references the deleted `initialData` field; the new `InputDefs` type param is undocumented; the brand docstrings say "nominal (not structural)" though the discrimination is structural).

### Publish logistics → #1481

Everything that must move in lockstep with the next `@skipruntime/*` publish, so it doesn't get lost: version bump, migrating + re-pinning the 4 held-back example apps, re-including `examples/*` in `check-ts.sh` (they currently have **no lint coverage** and type-checks only via `check-examples`, which doesn't trigger on `skipruntime-ts` changes), migrating the www prose docs before any docs deploy (typedoc regenerates from **local** `api.ts`, so a deploy from a post-merge tree would contradict the prose), and the knowingly-broken-until-publish `tests/native_addon` released-variant smoke test.

### Minor notes (no action needed)

- `Reducer.initial/remove`: `Nullable<A>` → `A | null` is pure alias inlining (`Nullable<T> = T | null`) — fine, though `api.ts` still exports `Nullable`.
- `binding.ts`'s `ResourceBuilder` appears to have zero construction sites anywhere in the tree (TS, skiplang FFI, or addon C++) — if it isn't reserved API surface, it could be deleted rather than retyped.
- Release-notes footnote: structural implementations (mocks/test doubles) of `EagerCollection`/`LazyCollection` now need the brand field.

---
🤖 Review by Claude (Fable 5) on Mehdi's behalf.

mbouaziz:
Fine with me, I'm just worried about the synchronization with @skiplabsdaniel's version

jberdine:
I have some upcoming PRs to try to close the gap with @skiplabsdaniel's version, or at least what I can see on the skip/skipper branch.

PR #1205, opened 2026-05-21
Pin serialize-javascript >=7.0.3 in cache_invalidation/www

mbouaziz:
## Summary
- Fixes Dependabot alert [#201](https://github.com/SkipLabs/skip/security/dependabot/201) — [GHSA-5c6j-r48x-rmvq](https://github.com/yahoo/serialize-javascript/security/advisories/GHSA-5c6j-r48x-rmvq) (RCE via `RegExp.flags` / `Date.toISOString`, severity high, patched in 7.0.3).
- The vulnerable `serialize-javascript@6.0.2` is a transitive dev-only dep, pulled in by `@vue/cli-service` via `copy-webpack-plugin` and `css-minimizer-webpack-plugin`.
- Adds both `overrides` (npm) and `pnpm.overrides` in [examples/cache_invalidation/www/package.json](examples/cache_invalidation/www/package.json) so the lockfile (scanned by Dependabot) and the Docker build's `npm install` both resolve to `serialize-javascript@7.0.5`.

## Test plan
- [x] `pnpm install --lockfile-only` resolves `serialize-javascript: 7.0.5` (verified via `grep serialize-javascript pnpm-lock.yaml`).
- [x] `npm install` (fresh, no lockfile) → `npm ls serialize-javascript` shows `serialize-javascript@7.0.5 overridden` under both consumers.
- [x] `npm run build` (vue-cli-service) still completes successfully — build hash unchanged vs. 6.0.2 build.
- [ ] CI: full `docker compose build` for cache_invalidation example (will run on PR).

### DIFF
--- a/examples/cache_invalidation/www/package.json
+++ b/examples/cache_invalidation/www/package.json
@@ -28,5 +28,13 @@
     "eslint-plugin-vue": "^9.28.0",
     "prettier": "3.4.1",
     "typescript": "~5.0.0"
+  },
+  "overrides": {
+    "serialize-javascript": "^7.0.3"
+  },
+  "pnpm": {
+    "overrides": {
+      "serialize-javascript": "^7.0.3"
+    }
   }
 }
--- a/examples/cache_invalidation/www/pnpm-lock.yaml
+++ b/examples/cache_invalidation/www/pnpm-lock.yaml
@@ -4,6 +4,9 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  serialize-javascript: ^7.0.3
+
 importers:
 
   .:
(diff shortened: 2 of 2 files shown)


mbouaziz:
A couple of questions came up reviewing this — answering for the record:

### Why an override rather than bumping the direct dep?

`serialize-javascript@6.0.2` is a **transitive** dep, not a direct one — it's pulled in by `@vue/cli-service@5.0.9` via `copy-webpack-plugin@9.1.0` and `css-minimizer-webpack-plugin@3.4.1`. Nothing in `package.json` mentions it. Options were:

1. Wait for the upstream webpack plugins to release versions depending on `serialize-javascript >=7.0.3`, then bump those — but `@vue/cli-service@5.0.9` is the latest 5.x (Vue CLI is frozen), so it'll keep pulling these older plugin versions.
2. Migrate the example off `@vue/cli-service` to Vite or `@vue/cli@6` — large refactor for an example app.
3. Override the transitive version — what this PR does. Both npm and pnpm let you force a specific resolution wherever the package appears in the tree.

### Breaking changes between 6.0.2 → 7.0.5?

Checked the release notes:

- **v7.0.0**: requires **Node.js v20+**. The cache_invalidation/www Dockerfile uses `node:22.13.1-alpine3.21`, so this is satisfied.
- **v7.0.1**: function body sanitization (security improvement; could affect output if you serialize functions, but webpack-plugin usage is metadata-only).
- **v7.0.3**: the security fix this PR is for.
- **v7.0.5**: array-like object validation, CPU usage fix.

Empirically, `npm run build` produces a `dist/` with the same hash before and after the override (`Hash: c599fbd2b03a026a`) — strong evidence the webpack plugins' usage isn't affected.

PR #1226, opened 2026-05-22
Pin uuid >=11.1.1 in /www

mbouaziz:
## Summary
Fixes Dependabot alert [#316](https://github.com/SkipLabs/skip/security/dependabot/316) — [GHSA-w5hq-g745-h8pq](https://github.com/uuidjs/uuid/security/advisories/GHSA-w5hq-g745-h8pq) / CVE-2026-41907 (missing buffer bounds check in uuid's `v3()`/`v5()`/`v6()` methods; medium severity).

`uuid@8.3.2` is pulled transitively into [www/package-lock.json](www/package-lock.json) via `@docusaurus/core` → `webpack-dev-server` → `sockjs`. The advisory affects all uuid versions `< 11.1.1` (plus 12.0.x and 13.0.x prereleases) — 8.3.2 is in scope.

Adds an npm `overrides` entry pinning `uuid: ^11.1.1`, the lowest patched version on a still-supported line (uuid 10 and below is upstream-deprecated). Surgical `npm update uuid --package-lock-only` regenerated the lockfile, swapping the single transitive entry `8.3.2 → 11.1.1`.

## Compatibility verified
- `uuid@11.1.1` has CommonJS exports under the `node.require` condition (`./dist/cjs/index.js`); `require('uuid').v4()` returns a valid UUID — sockjs's `require('uuid').v4()` keeps working.
- This dep only runs during `docusaurus start` (HMR via sockjs in webpack-dev-server); never reaches the production static site.

## Diff
- `www/package.json`: +1 line in `overrides` block.
- `www/package-lock.json`: surgical 13-line change (single `uuid` entry, version 8.3.2 → 11.1.1).

### DIFF
--- a/www/package-lock.json
+++ b/www/package-lock.json
@@ -18074,13 +18074,16 @@
       }
     },
     "node_modules/uuid": {
-      "version": "8.3.2",
-      "resolved": "https://registry.npmjs.org/uuid/-/uuid-8.3.2.tgz",
-      "integrity": "sha512-+NYs2QeMWy+GWFOEm9xnn6HCDp0l7QBD7ml8zLUmJ+93Q5NF0NocErnwkTkXVFNiX3/fpC6afS8Dhb/gz7R7eg==",
-      "deprecated": "uuid@10 and below is no longer supported.  For ESM codebases, update to uuid@latest.  For CommonJS codebases, use uuid@11 (but be aware this version will likely be deprecated in 2028).",
+      "version": "11.1.1",
+      "resolved": "https://registry.npmjs.org/uuid/-/uuid-11.1.1.tgz",
+      "integrity": "sha512-vIYxrBCC/N/K+Js3qSN88go7kIfNPssr/hHCesKCQNAjmgvYS2oqr69kIufEG+O4+PfezOH4EbIeHCfFov8ZgQ==",
+      "funding": [
+        "https://github.com/sponsors/broofa",
+        "https://github.com/sponsors/ctavan"
+      ],
       "license": "MIT",
       "bin": {
-        "uuid": "dist/bin/uuid"
+        "uuid": "dist/esm/bin/uuid"
       }
     },
     "node_modules/value-equal": {
--- a/www/package.json
+++ b/www/package.json
@@ -48,6 +48,7 @@
     "node": ">=20"
   },
   "overrides": {
-    "serialize-javascript": "^7.0.3"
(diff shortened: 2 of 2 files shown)


mbouaziz:
A couple of questions came up reviewing this — answering for the record:

### Was there a docusaurus update that would have fixed this without the override?

No — everything in the chain is already at its latest published version:

- `@docusaurus/core` is at **3.10.1** (latest stable; what we have).
- It declares `webpack-dev-server: ^5.2.2`; current resolution is **5.2.4** (latest 5.x). webpack-dev-server still pins `sockjs: ^0.3.24`.
- `sockjs` is at **0.3.24** (latest; the package is effectively unmaintained), which still pins `uuid: ^8.3.2`.

So the vulnerable transitive sits at the end of a chain whose every link is at HEAD. Bumping docusaurus wouldn't help; the fix has to come from sockjs bumping uuid (unlikely soon), webpack-dev-server dropping sockjs, or an override at our end.

### Are there breaking changes between uuid 8.3.2 and 11.1.1 that affect our usage?

Three major bumps; none affect sockjs's call site.

| Version | Notable breaking changes | Affects us? |
|---|---|---|
| v9.0.0 | Default export became a named function (`import uuid from 'uuid'; uuid()` no longer works) | No — sockjs uses `require('uuid').v4()` |
| v9.0.0 | Removed UMD build; dropped IE 11/Safari 10/msCrypto fallback | No — server-side Node |
| v9.0.0 | Dropped Node 10 | No — we're on Node 22 |
| v10.0.0 | Dropped Node 12/14; added v6/v7/v8/MAX (additive) | No |
| v11.0.0 | Refactored v1/v7 internal state | No — we only use v4 |
| v11.0.0 | Ported to TypeScript | No — types only, runtime identical |
| v11.0.0 | Test matrix listed as "Node 16-20" | No `engines.node` restriction is published; Node 22 works |
| v11.1.1 | Backports the GHSA-w5hq-g745-h8pq fix (additive) | Yes — that's the point |

**Our actual usage** is a single call: sockjs's `require('uuid').v4()`. The `v4()` no-options signature has been stable across all majors. Runtime-verified in a fresh CJS context — `require('uuid').v4()` at 11.1.1 returns a valid v4 UUID.

Also worth noting that sockjs runs only during `docusaurus start` (webpack-dev-server HMR transport). It never reaches the production static site we ship, so any hypothetical regression would only impact local dev.

PR #1245, opened 2026-05-27
fix(skstore): purge DeletedDir tombstones at end of update transaction

zaccharyfavere:
fix(skstore): bound DeletedDir tombstone leak and keep deletions visible

Rebased on main (1813f196). Test setup adjusted for the region API change:
the tests now use SKStore.withRegionVoid instead of the newly-private
newObstack.

DeletedDir tombstones created by removeDir were never collected, leaking
indefinitely. They are now purged once older than a fixed tick window
(tick - 1000), in the tail of resolveContext on the committed context, so
the purge applies uniformly to every commit path (both the delta branch and
the synchronizer's newRoot branch) and reaches the shared root. Bounding the
leak rather than eliminating it mirrors the window strategy already used for
file history.

Purging tombstones breaks deletion visibility for a lagging import reader,
since the deletion vanishes from getDirsChangesAfter once its tombstone is
gone. To keep deletions observable, getDirsChangesAfter now returns
(purged, changed, allExisting): when the requested tick predates the purge
window, it flags purged and returns the full live-dir set, so import (and
SKDB's importNext) can deduce deletions by absence instead of relying on the
tombstone. SKDB's importNext translates source input dirs through
makeInputDirName before the absence check, to account for its /next/
versioning rename.

Absence deduction only removes dirs that existed before the fork: the guard
created < tick protects dirs created locally in the target after the fork
(created == tick), which are additions never present in the source, not
deletions. Without it, a lagging import could destroy the target's local
writes.

Deductions inferred during import are posted after resolveContext snapshots
changedDirs, so the final context is re-scanned before CHANGED_DIRS is set,
otherwise subscribers to the deleted dirs are never notified.

Tests: prelude tests cover the generic import path (deletion propagated
after purge, local dirs preserved, inherited deletion still propagated), and
a first SKDB .sk test covers the importNext rename path. Each fails without
its guard and passes with it. The SKDB test runs via a small skargo
test-harness wired into make test-native.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,8 +203,11 @@ test-%:
 	cd $* && skargo test --profile $(SKARGO_PROFILE)
 
 .PHONY: test-native
+# The skargo test step above runs the .sk tests first; the shell pipeline below
+# is silenced (@) so its long command line does not bury those results.
 test-native: build/skdb
-	cd sql/ && SKARGO_PROFILE=$(SKARGO_PROFILE) SKDB_BIN=$(realpath build/skdb) ./test_sql.sh \
+	cd sql && skargo test --profile $(SKARGO_PROFILE)
+	@cd sql/ && SKARGO_PROFILE=$(SKARGO_PROFILE) SKDB_BIN=$(realpath build/skdb) ./test_sql.sh \
 	|tee /tmp/native-test.out ; \
 	! grep -v '\*\|^[[:blank:]]*$$\|OK\|PASS' /tmp/native-test.out
 
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -391,6 +391,8 @@ mutable base class WriteChecker {
 mutable class Context private {
   private mutable toPurge: SortedSet<DirName> = SortedSet[],
   private mutable dirs: Dirs = Dirs{},
+  private mutable deletedDirs: SortedSet<DirName> = SortedSet[],
+  private mutable dirsTombLimit: ?Tick = None(),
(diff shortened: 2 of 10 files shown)


mbouaziz:
Hi, thanks for your contribution!
Sorry, CI failures might be because I switched jobs to use gen 2 circle ci instances but they may be counting memory differently

zaccharyfavere:
Hi, alright thanks for your return !

mbouaziz:
Sorry, still fixing the CI...
Very bad timing for a broken CI, we haven't had any external contribution for months :laughing:

zaccharyfavere:
No worries, I just re pushed because I realised there was a format error still in my code. That should be fixed now since it made one of the CI pass

mbouaziz:
I think the CI is back to normal, I've rebased your commit to check

mbouaziz:
`getDirsChangesAfter(tick)` relies on `DeletedDir` tombstones to report deletions to downstream consumers after `updateWithStatus` returns. Purging them here means subsequent calls (e.g. `importNext`/`import` in the synchronizer's `exportToRoot` path, `notifyAll`, `CHANGED_DIRS` computation in `resolveContext`) won't see the deletion.

Concretely, a `DROP TABLE`/`DELETE` followed by `update()` would purge the tombstone before `exportToRoot` -> `importNext` scans for `DeletedDir`, so the drop never propagates to root.

I understand the memory leak but a safe purge needs to be bounded by the slowest consumer of the change history, e.g. track the min tick across active synchronizers/subscribers and only drop tombstones below it, or do the compaction at a synchronization boundary that's provably past the last reader.

skiplabsdaniel (approved):

PR #1259, opened 2026-05-29
Migrate native addon from V8 API to node-addon-api, add Bun support

zaccharyfavere:
cc @skiplabsdaniel 

- Rewrite the native addon (cjson, common, fromjs, tojs, main) from V8 direct API to node-addon-api for better ABI stability across Node.js versions.
- Add #include <sstream> in common.cc for std::ostringstream support.
- Add Bun runtime support:
  - New unit test spec (tests/src/addon.bun.spec.ts) running the test suite under Bun via mocha with a dedicated .mocharc.bun.json config.
  - New shell script (tests/examples/bun/run_examples.sh) executing examples under Bun in src mode.
  - New Makefile targets: test-bun-unit, test-bun-examples, test-skipruntime-ts-bun.
- Adapt database example for dual-runtime compatibility:
  - Use dynamic import to select bun:sqlite under Bun, better-sqlite3 under Node, via a small getDatabase() helper in database.ts and database-server.ts.
  - Switch to positional SQL parameters (?) for compatibility between the two sqlite drivers.
- Use JSON.stringify in groups-client.ts and database-client.ts to produce deterministic output across runtimes.
- Update database.exp.out and groups.exp.out to match the JSON.stringify output format.
- Add node-addon-api 8.7.0 as a dependency of @skipruntime/native.


skiplabsdaniel (approved):

skipruntime-ts/addon/src/fromjs.cc
```diff
@@ -11,454 +11,429 @@ using skbinding::CallJSNullableStringFunction;
...
-  kExternFunctions.Reset(isolate, externFunctions);
+void SetFromJSBinding(Napi::Env /*env*/, Napi::Object externFunctions) {
+  kExternFunctions = Napi::Persistent(externFunctions);
 }
 
 extern "C" {
 
+/* Error / context plumbing. */
+
 double SkipRuntime_getErrorHdl(SKException exn) {
```
mbouaziz, line 29: **[major] All Skip→JS trampolines lost their per-call `HandleScope`.**

The old V8 code opened `HandleScope scope(isolate);` at the top of each of the 38 `extern "C"` callbacks in this file; the migrated code opens none (`grep HandleScope addon/src` → no matches). These callbacks are invoked from Skip-compiled code *inside* an outer JS→native call (`SkipRuntime_Runtime__update`, `Collection__map`, …), so every handle created here (`kExternFunctions.Value()`, per-arg `Napi::Number/External::New`, `binding.Get` in `CheckFunction`, the call result) lands in the **outer** call's handle scope and is only reclaimed when that outer binding returns.

`Mapper__mapEntry` / `Reducer__add` / `Notifier__notify` fire once per entry during change propagation, so handle usage grows O(collection size) within a single native call — exactly the N-API pitfall the old per-callback scopes prevented, on the hot path.

Fix: add `Napi::HandleScope scope(env);` right after obtaining `env` in each `extern "C"` callback.

skipruntime-ts/addon/src/fromjs.cc
```diff
@@ -11,454 +11,429 @@ using skbinding::CallJSNullableStringFunction;
...
-using v8::Number;
-using v8::Object;
-using v8::Persistent;
-using v8::Value;
+/* Persistent reference to the JS object holding all the user's
+   callbacks Skip can invoke from native code. Initialized once at startup
+   via SetFromJSBinding, then used for each Skip -> JS call. */
 
-static Persistent<Object> kExternFunctions;
+static Napi::ObjectReference kExternFunctions;
```
mbouaziz, line 19: **[major] `static Napi::ObjectReference` without `SuppressDestruct()` — env use-after-free at teardown.**

The old `static v8::Persistent<Object>` destructor never touches the isolate (safe leak at exit). `Napi::Reference`'s destructor calls `napi_delete_reference(_env, _ref)`, and for a static that runs during static destruction — after the environment (and the `napi_env` it captured) is gone. This is the documented use case for `SuppressDestruct()`; it will bite under worker threads and is untested territory under Bun.

```cpp
kExternFunctions = Napi::Persistent(externFunctions);
kExternFunctions.SuppressDestruct();
```

Related (minor): the callbacks now use `kExternFunctions.Env()` captured at init, where the old code resolved `Isolate::GetCurrent()` per call. With the context-aware `NODE_API_MODULE`, a second init from another env (worker/embedded context) silently rebinds the process-global runtime's callbacks to the newest env — worth at least a guard that `kExternFunctions` is non-empty and a comment documenting the single-env invariant.

skipruntime-ts/addon/src/tojs.cc
```diff
@@ -26,7 +25,10 @@ SKReducer SkipRuntime_createReducer(int32_t ref);
 
 CJSON SkipRuntime_initService(SKService service);
 double SkipRuntime_closeService();
-
+void SkipRuntime_setLazyCollectionValue(char* dirName, CJSON key,
+                                        CJArray values);
+double SkipRuntime_ServiceDefinition__fetch(int32_t service, char* supplier,
+                                            CJObject key);
```
mbouaziz, line 31: **[minor] Two new bindings beyond the migration that cannot work — recommend dropping them from this PR.**

1. `SkipRuntime_setLazyCollectionValue` (declared here, wrapped at L185, registered at L841) is defined nowhere: not in the addon, no `@export` in `FFI.sk`, no caller anywhere in the repo. The link succeeds (no `--no-undefined`), so this becomes a dynamic-linker crash the first time JS invokes the binding.

2. `SkipRuntime_ServiceDefinition__fetch` is declared here with **3** parameters and called at L175, but the same symbol is *defined* in `fromjs.cc:423` with **4** parameters — calling through the mismatched prototype is UB (the `CJObject` pointer is read as a `char*` by `Napi::String::New`, and the 4th param gets register garbage). There's no Skip-side export of this name, so the linker binds the call straight to the fromjs.cc trampoline.

Neither existed in the old addon and neither is mentioned in the PR description — they look accidentally imported from a different branch.

skipruntime-ts/addon/src/fromjs.cc
```diff
@@ -11,454 +11,429 @@ using skbinding::CallJSNullableStringFunction;
...
-                     argv);
+  Napi::Env env = kExternFunctions.Env();
+  Napi::Object externFunctions = kExternFunctions.Value();
+  std::vector<napi_value> args = {Napi::Number::New(env, reducerId)};
+  CallJSVoidFunction(env, externFunctions, "SkipRuntime_deleteReducer", args);
+}
+
+/* External service fetch callback. */
+
+double SkipRuntime_ServiceDefinition__fetch(uint32_t serviceId, char* supplier,
```
mbouaziz, line 423: **[minor] New trampoline with no JS-side handler registered anywhere.**

This is the 4-arg definition that conflicts with the 3-arg declaration in `tojs.cc:30` (see comment there). Even with matching signatures, no TS code in `@skipruntime/core` (or anywhere in-tree) registers a `SkipRuntime_ServiceDefinition__fetch` property on the extern-functions object, so any invocation lands in `CheckFunction`'s `"Undefined function"` path. Suggest removing alongside the tojs.cc entry.

skipruntime-ts/addon/src/common.cc
```diff
@@ -1,409 +1,355 @@
...
-      }
+    Napi::Error napiError = Napi::Error::New(env, error.str());
+    char* stack = SkipRuntime_getExceptionStack(e.m_skipException);
+    if (stack != nullptr && strlen(stack) > 0) {
+      sk_string_check_c_safe(stack);
+      napiError.Value().Set("stack", Napi::String::New(env, stack));
     }
-    isolate->ThrowException(v8Error);
+    napiError.ThrowAsJavaScriptException();
   } catch (std::exception& e) {
```
mbouaziz, line 314: **[minor] This `catch (std::exception&)` also catches `Napi::Error`, contradicting the comment above.**

`Napi::Error` derives from `std::exception`, so a JS exception propagating out of `run(env)` under `NAPI_CPP_EXCEPTIONS` is intercepted here and re-wrapped as a fresh error containing only `what()` text — original error object, type, and stack are lost. The comment on this function says Napi::Error "propagates automatically and does not need to be handled here", but this clause prevents exactly that.

Fix: add `catch (const Napi::Error&) { throw; }` before this clause (or reorder with an explicit rethrow).

skipruntime-ts/addon/src/common.cc
```diff
@@ -1,409 +1,355 @@
...
+    SkipRuntime_throwExternalException(skempty, skmessage, skempty);
+  }
+}
+
+/*
+  Call a JS function from C++ and forward JS exceptions to the Skip runtime.
+  On success, delegates result handling to the `success` lambda.
+  On failure, runs `failure` for cleanup before propagating the error.
+*/
+void* SKTryCatch(Napi::Env env, Napi::Function fn, Napi::Value recv,
```
mbouaziz, line 106: **[minor] Throwing Napi calls run outside the try-block in frames called from Skip code.**

Under `NAPI_CPP_EXCEPTIONS`, every Napi call that fails throws. In the `fromjs.cc` trampolines, argument marshalling (`Napi::Number/String/External::New`) and `CheckFunction`'s `binding.Get` execute *before* this try-block is entered, and `HandleJSException` calls `JSONStringify` (L84/L95) *inside* the catch handler — `JSON.stringify` throws on circular values. In all those cases a C++ exception unwinds through Skip/C frames instead of going through the `SkipRuntime_throwExternalException` protocol, bypassing Skip's exception bookkeeping. The old code used Maybe-based V8 APIs + `v8::TryCatch` and had no throwing path here.

Low probability per call, but this is the hot Skip→JS path (mappers/reducers/notifiers); worth widening the guarded region and giving `HandleJSException` a fallback for a throwing stringify.

skipruntime-ts/addon/src/common.cc
```diff
@@ -1,409 +1,355 @@
...
-          jsObj->Get(context, FromUtf8(isolate, "message")).ToLocalChecked());
-      char* skstack = ToSKString(
-          isolate,
-          jsObj->Get(context, FromUtf8(isolate, "stack")).ToLocalChecked());
+/*
+  Internal helper shared by SKTryCatch and SKTryCatchVoid.
+  Converts a JS exception into a call to SkipRuntime_throwExternalException.
+  Never returns.
+*/
+static void HandleJSException(Napi::Env env, Napi::Value exception) {
```
mbouaziz, line 64: **[minor] Error-translation semantics diverge from the old implementation.**

- "Native error" is now duck-typed on a string `stack` property; old code used `IsNativeError()`. A thrown plain object with a string `stack` now takes the native-error branch.
- The type is now read from `.name`; old code used `GetConstructorName()`. For `class MyError extends Error {}` without an own `name`, the type forwarded to Skip changes from `"MyError"` to `"Error"`.
- An `Error` whose `message` is a non-string (e.g. `42`) used to forward the coerced string `"42"`; now forwards `""`.

May all be acceptable, but it changes Skip-observable error type/message strings — worth a deliberate decision (and ideally a test).

skipruntime-ts/addon/src/cjson.cc
```diff
@@ -37,520 +37,411 @@ double SKIP_SKJSON_arraySize(CJArray json);
...
+  }
+  Napi::Value result;
+  NatTryCatch(env, [&result, &info, env](Napi::Env) {
+    double skvalue;
+    if (info[0].IsNumber()) {
+      skvalue = info[0].As<Napi::Number>().DoubleValue();
     } else {
-      skvalue = args[0].As<BigInt>()->Int64Value();
+      bool lossless;
+      skvalue = (double)info[0].As<Napi::BigInt>().Int64Value(&lossless);
```
mbouaziz, line 169: **[minor] `lossless` is ignored — BigInts above 2⁶³/outside ±2⁵³ silently lose precision.**

Worth either throwing a `RangeError` when `!lossless` (matching what `Napi::BigInt` users typically do) or documenting the truncation.

(Separately: declaring `SKIP_SKJSON_createCJInt(double)` is a genuine fix — the Skip side has taken `Float` since ab6a4d63d while the old C++ declared `int64_t`, a latent ABI mismatch. 👍)

skipruntime-ts/addon/binding.gyp
```diff
@@ -9,6 +9,9 @@
         "src/fromjs.cc",
         "src/main.cc"
       ],
+      "include_dirs": ["<!@(node -p \"require('node-addon-api').include\")"],
+      "dependencies": ["<!(node -p \"require('node-addon-api').gyp\")"],
+      "defines": ["NAPI_VERSION=9", "NAPI_CPP_EXCEPTIONS"],
```
mbouaziz, line 14: **[minor] `NAPI_VERSION=9` rejects Node 20.0–20.2 while workspace engines claim `>=20.0.0`.**

Node-API 9 needs Node ≥ 18.17 / ≥ 20.3; on 20.0–20.2 the module fails at `require` time with `ERR_DLOPEN_FAILED`. Nothing in the addon uses anything above Node-API 6 (`Napi::BigInt`), so `NAPI_VERSION=8` widens compatibility at zero cost — or add `"engines": {"node": ">=20.3.0"}` to the addon package.json.

Nit: the `-fno-exceptions` removal is gcc/Linux-only — no `xcode_settings`/`msvs_settings` counterpart if other platforms are ever targeted.

skipruntime-ts/tests/.mocharc.json
```diff
@@ -1,4 +1,4 @@
 {
-  "spec": ["./dist/**/*.spec.js"],
+  "spec": ["./dist/src/addon.spec.js", "./dist/src/wasm.spec.js"],
```
mbouaziz, line 2: **[minor] Glob → explicit allow-list: future spec files will silently never run under Node.**

Any new `tests/src/*.spec.ts` will compile into `dist/` but be skipped unless someone remembers to add it here. Auto-discovery can be kept while excluding the bun duplicate:

```json
"spec": ["./dist/**/*.spec.js"],
"ignore": ["./dist/**/*.bun.spec.js"]
```

(Or better: drop the duplicate spec entirely — see comment on `addon.bun.spec.ts`.)

skipruntime-ts/tests/src/addon.bun.spec.ts
```diff
@@ -0,0 +1,3 @@
+import { initTests } from "./tests.js";
```
mbouaziz, line 1: **[nit] Line-for-line duplicate of `addon.spec.ts` (diff: one blank line).**

`.mocharc.bun.json` could point its spec at `src/addon.spec.ts` directly — mocha under bun consumes the `.ts` source, and the file has nothing bun-specific. That deletes this file, removes the drift risk between the two copies, and eliminates the reason `.mocharc.json` had to abandon its glob.

skipruntime-ts/tests/.mocharc.bun.json
```diff
@@ -0,0 +1,5 @@
+{
```
mbouaziz, line 1: **[nit]** Committed with mode `100755` — a JSON config shouldn't carry the executable bit (its sibling `.mocharc.json` is `100644`).

skipruntime-ts/tests/examples/bun/run_examples.sh
```diff
@@ -0,0 +1,114 @@
...
+#
+# For each example, the script
+#   - Starts principal service <name>.ts
+#   - Starts the additional server service <name>-server.ts if it exists
+#   - Starts potential additionnal hardocded service (like for remote) (cf EXTRA_SERVICES)
+#   - Starts client <name>-client.ts and redirect its stderr/stdout
+#   - Compare reference files .exp.out / .exp.err with actual example outputs
+#   - Show the diff if fails, but continue to try the other examples
+
+ALL_EXAMPLES=(
```
mbouaziz, line 18: **[minor] Hardcoded example list will silently drift from the Node side.**

The Node flow auto-discovers via `$(wildcard tests/examples/*.sh)` in the Makefile; a future example added there won't run under Bun unless someone remembers this array. Could derive it: `for f in "$REF_DIR"/*.sh; do ...` (or from `*.exp.out`).

skipruntime-ts/tests/examples/bun/run_examples.sh
```diff
@@ -0,0 +1,114 @@
...
+
+    echo "Running '${name}' example on @skipruntime/native (bun, src mode)"
+
+    local extras="${EXTRA_SERVICES[$name]:-}"
+    for extra in $extras; do
+        SKIP_PLATFORM="native" bun run "$EXAMPLES_DIR/${extra}.ts" >/dev/null &
+        BG_PIDS+=($!)
+    done
+
+    SKIP_PLATFORM="native" bun run "$EXAMPLES_DIR/${name}.ts" >/dev/null &
```
mbouaziz, line 65: **[minor] Examples run with cwd `tests/examples/bun`, so `db.sqlite` is created there — untracked and not gitignored.**

`database.ts`/`database-server.ts` open `./db.sqlite` relative to cwd. The Node scripts `cd ../../examples` first, where `db.sqlite` is covered by `skipruntime-ts/examples/.gitignore`; this script doesn't, so `make test-bun-examples` leaves `skipruntime-ts/tests/examples/bun/db.sqlite` dirtying `git status`, and the two runtimes use different persistent DB files. Output stays deterministic only because `initDB()` re-seeds with `INSERT OR REPLACE` and the client only mutates id 123 — fragile.

Fix: launch the example processes from `$EXAMPLES_DIR` (matching the Node scripts), or gitignore the file here.

skipruntime-ts/tests/examples/bun/run_examples.sh
```diff
@@ -0,0 +1,114 @@
...
+    BG_PIDS+=($!)
+
+    if [ -f "$EXAMPLES_DIR/${name}-server.ts" ]; then
+        bun run "$EXAMPLES_DIR/${name}-server.ts" >/dev/null &
+        BG_PIDS+=($!)
+    fi
+
+    sleep 1
+
+    bun run "$EXAMPLES_DIR/${name}-client.ts" >"$out_file" 2>"$err_file"
```
mbouaziz, line 75: **[minor] Client/service exit codes are never checked — pass/fail is decided purely by output diffs.**

No `set -e`, and the status of this `bun run` is ignored. A client that exits non-zero after emitting the expected output (e.g. crash in teardown after `evSource.close()`), or a service that dies late, passes green. The Node counterparts use `set -e` with the client as the last command, so a non-zero exit fails the make target there. Suggest capturing `$?` here and folding it into `fail`.

Nit: the `INT`/`TERM` trap (L48) kills children but doesn't `exit`, so Ctrl-C mid-suite continues to the next example.

skipruntime-ts/Makefile
```diff
@@ -98,6 +98,18 @@ test-native-example-%:
...
+test-bun-unit:
+	cd tests && \
+	bun --bun ../../node_modules/.bin/mocha --config .mocharc.bun.json
+
+.PHONY: test-bun-examples
+test-bun-examples:
+	cd tests/examples/bun && ./run_examples.sh
+
+.PHONY: test-skipruntime-ts-bun
+test-skipruntime-ts-bun: build test-bun-unit test-bun-examples
```
mbouaziz, line 111: **[minor] Works only under sequential make — the test targets don't depend on `build`.**

Prerequisite order isn't an ordering guarantee: under `make -j`, `test-bun-unit`/`test-bun-examples` can start before (or concurrently with) `build`. Giving both targets a `build` prerequisite makes the dependency real. (The pre-existing `test-all` shares the pattern, but no need for the new targets to cement it.)

.circleci/base.yml
```diff
@@ -189,6 +189,43 @@ jobs:
...
+          KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
+    resource_class: large.gen2
+    environment:
+      SKIP_CAPACITY: 6G
+    steps:
+      - custom-checkout:
+          submodules: true
+      - run:
+          name: Install Bun
+          command: npm install -g bun
```
mbouaziz, line 222: **[minor] Unpinned — this job will silently track every Bun release.**

Since the job specifically exercises Bun's Node-API compat layer (the surface most likely to change between Bun releases), an upstream release can break CI on unrelated PRs and runs aren't reproducible. Suggest pinning (`npm install -g bun@<x.y.z>`) with deliberate bumps. If tracking latest is *intended* as an early-warning signal, a comment saying so would prevent future "fix": pinning by someone else.

.circleci/base.yml
```diff
@@ -189,6 +189,43 @@ jobs:
       - store_test_results:
           path: ~/test-results/skipruntime.xml
 
+  skipruntime-bun:
+    docker:
```
mbouaziz, line 193: **[nit]** This 22-line postgres/kafka sidecar block is byte-identical to the `skipruntime` job's. The services are genuinely needed (tests.ts hard-fails under `CIRCLECI` without them — verified), but version bumps now have to happen in two places. `generate_config.sh` just `cat`s this file, so YAML anchors would deduplicate cleanly.

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "skip",
+  "name": "workspace",
```
mbouaziz, line 2: **[nit]** Root `"name"` flipped `"skip"` → `"workspace"` — the lockfile was regenerated in a differently-named directory and this will churn back on the next `npm install` from a normal checkout. Worth regenerating before merge. (The `node-addon-api` entry itself is fine: registry URL + integrity hash present.)

skipruntime-ts/examples/database.ts
```diff
@@ -7,20 +8,36 @@ import type {
...
 */
 
 const platform: "wasm" | "native" =
   process.env["SKIP_PLATFORM"] == "native" ? "native" : "wasm";
 
+/*****************************************************************************/
+// Conditional database driver loading: bun:sqlite under Bun, better-sqlite3 under Node
+/*****************************************************************************/
+
+async function getDatabase(): Promise<any> {
```
mbouaziz, line 22: **[nit] Driver typed as `any`, file-wide eslint-disable, and this helper is duplicated verbatim in `database-server.ts`.**

The old code was fully typed against better-sqlite3 (`initDB(): Database.Database`, typed `prepare<...>`). Only `prepare`/`run`/`all`/`close` are used, and both drivers are call-compatible after the switch to positional `?` params, so a small shared structural interface in `examples/utils.ts` would restore call-site type safety, drop both eslint-disables, and deduplicate `getDatabase()`. Matters mostly because these examples are user-facing documentation of recommended patterns.

mbouaziz (changes requested):
Multi-agent review (6 dimensions — C++ memory safety, V8→N-API semantic equivalence, exception/build model, Bun test infra, dual-runtime examples, CI config — with adversarial verification of every finding). Detailed findings are inline; summary and the non-anchorable items below.

### What's solid
- The rewrite is faithful overall and considerably more readable; per-function old-vs-new comparison found no silent behavioral drops beyond the items noted inline.
- CI wiring verified end-to-end: `skipruntime-bun` is gated identically to the node `skipruntime` workflow (including transitive skjson/prelude triggers), the postgres/kafka sidecars are genuinely required by `tests.ts`, `large.gen2` + `SKIP_CAPACITY=6G` matches repo convention, and the job passed in 3.8 min.
- The new `.exp.out` files match the new client output exactly (traced line-by-line, including the `JSON.stringify([])` ↔ `[]` coincidence on the delete path).
- Declaring `SKIP_SKJSON_createCJInt(double)` fixes a *latent ABI mismatch* — the Skip side has taken `Float` since ab6a4d63d while the old C++ declared `int64_t`.

### Main asks (details inline)
1. **[major]** Restore per-callback `Napi::HandleScope` in all `fromjs.cc` trampolines — O(collection size) handle accumulation per native call on the hot path.
2. **[major]** `kExternFunctions.SuppressDestruct()` — static `Napi::ObjectReference` destructor runs `napi_delete_reference` on a dead env at teardown.
3. **[minor]** Remove (or finish) the two beyond-the-migration bindings: `SkipRuntime_setLazyCollectionValue` (undefined symbol → crash on first call) and `SkipRuntime_ServiceDefinition__fetch` (conflicting 3-arg/4-arg prototypes → UB; no JS handler registered).
4. **[minor]** Docs contradiction: `README.md:44` and `www/docs/getting_started.md:22` still say the native runtime is *"currently only available for Node"* — the claim this PR exists to remove.

### Not inline-able / FYI
- **Pre-existing (not introduced here, just carried over):** `NatTryCatch`'s `catch (skbinding::SkipException&)` never matches — the runtime throws `skip::SkipException` (`runtime64_specific.cpp:184`), so Skip error names/stacks have never reached that branch. Fine to leave for a follow-up, but since this PR touches the exception model anyway it may be worth fixing while here.
- `ToSKString` no longer truncates at embedded NUL (old code used `strlen`; new passes full `std::string` length) — looks like an improvement, but it changes the bytes Skip receives for strings containing `\0`; flagging as an unannounced behavior change.
- The bun unit run covers only the native suite; the wasm spec isn't exercised under Bun (presumably intentional — fine, just noting).

### Verdict
Request changes for items 1–3 (1 and 2 are one-liners; 3 is likely an accidental import to delete) plus the README/docs line. Everything else is take-it-or-leave-it polish.

PR #1261, opened 2026-05-29
Bump dev/CI base image to Ubuntu 24.04; declare @skipruntime/native platform constraints

mbouaziz:
## Summary

Scoped down from the original Ubuntu-24.04 **+ Debian-Trixie** bump after review feedback.

- **Bump the dev/CI toolchain image to Ubuntu 24.04 (Noble LTS)** — `Dockerfile` + `bin/apt-install.sh` (LLVM apt repo `jammy` → `noble`, and `--break-system-packages` for the `black` install since Noble enforces PEP 668). This is the build/CI image only.
- **Declare platform constraints on `@skipruntime/native`** — add `os`/`cpu`/`libc` to `package.json` so npm refuses installation on Windows/macOS/musl/non-amd64-arm64 hosts and, via `@skiplabs/skip`'s `optionalDependencies`, falls back to `@skipruntime/wasm`. This encodes the runtime's *existing* platform support and is independent of any glibc version.

**Intentionally _not_ changed:** `skiplang/Dockerfile` stays on Debian Bookworm. The glibc floor of the distributed `libskipruntime.so` comes solely from that image (via `skipruntime-ts/Dockerfile` → `skiplang-bin-builder`), so keeping it on Bookworm preserves the current broad-compatibility floor. The Ubuntu bump does not affect it.

## Test plan

- [x] `skiplang-base` stage builds on `ubuntu:24.04`; LLVM 20 installs from `apt.llvm.org/noble` (verified locally)
- [x] `pip install black` fails on 24.04 with `externally-managed-environment`; `--break-system-packages` resolves it (verified locally)
- [x] Full `skip` target builds end-to-end on Noble, incl. the full `apt-install.sh` chain (verified locally)
- [ ] CI green

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04 AS skiplang-base
+FROM ubuntu:24.04 AS skiplang-base
 
 ARG TARGETARCH
 
--- a/bin/apt-install.sh
+++ b/bin/apt-install.sh
@@ -36,7 +36,7 @@ for step in "${steps[@]}"; do
         skiplang-build-deps)
             _ensure_base_deps
             wget -qO /etc/apt/keyrings/llvm.asc https://apt.llvm.org/llvm-snapshot.gpg.key
-            echo "deb [signed-by=/etc/apt/keyrings/llvm.asc] http://apt.llvm.org/jammy/ llvm-toolchain-jammy-$LLVM_VERSION main" >> /etc/apt/sources.list.d/llvm.list
+            echo "deb [signed-by=/etc/apt/keyrings/llvm.asc] http://apt.llvm.org/noble/ llvm-toolchain-noble-$LLVM_VERSION main" >> /etc/apt/sources.list.d/llvm.list
             apt-get update
             apt-get install -q -y --no-install-recommends automake clang-$LLVM_VERSION file gawk git lld-$LLVM_VERSION llvm-$LLVM_VERSION make openssh-client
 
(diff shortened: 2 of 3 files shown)


jberdine (approved):

jberdine (commented):
The PR test plan isn't clear about whether the listed testing steps have been done or are yet to do. Which is it?

Overall looks ok to me, but in general we want to use the oldest possible glibc to broaden compatibility. Is there a particular motivation to require a newer version?

mbouaziz:
Thanks Josh — good catch, and you're right. There's no real motivation to require a newer glibc, so I've dropped that part.

The key realization: the CI/dev image and the distributed binary's glibc floor are independent. `libskipruntime.so`'s floor comes entirely from `skiplang/Dockerfile` (Debian, via `skipruntime-ts/Dockerfile` → `skiplang-bin-builder`), **not** from the main `Dockerfile`. So we can modernize the build/CI image without touching end-user compatibility at all.

I've rescoped the PR accordingly:
- **Kept:** the dev/CI image bump to Ubuntu 24.04 (Noble) — build/CI toolchain only.
- **Reverted:** `skiplang/Dockerfile` stays on Debian Bookworm, so the runtime keeps its current (broad) glibc floor. No newer glibc required.
- **Kept:** the `os`/`cpu`/`libc` constraints on `@skipruntime/native`, since those just encode the existing Linux/glibc-only support (no version claim).

On the test plan: those were all "to do", sorry for the ambiguity — I've reworked the checklist to mark what's actually verified. So far, locally: the `skiplang-base` stage builds on Noble with LLVM 20 from `apt.llvm.org/noble`, and I confirmed the `--break-system-packages` change is needed (bare `pip install` now hits PEP 668's `externally-managed-environment`) and resolves it. Full-image build + CI are the remaining gates.

PR #1285, opened 2026-07-06
Fix ICE when piping into a function-bounded type parameter

### LINKED ISSUE #1269: ICE: piping into a tparam bounded by a tracked function type — "Expected closure type but got F"
## Summary

Piping into a value whose type is a **type parameter bounded by a tracked function type** passes the type checker but crashes the compiler in lowering:

```skip
fun callIt<F: (String) -> String>(f: F, x: String): String {
  x |> f
}

fun main(): void {
  print_raw(callIt(y -> y, "ok\n"))
}
```

```
/tmp/probe_ice.sk(2, 3)-(2, 9): Internal error: Expected closure type but got F
```

The equivalent **direct call** compiles and runs fine:

```skip
fun callIt<F: (String) -> String>(f: F, x: String): String {
  f(x)
}
```

## Analysis

- The ICE comes from `emitClosureCall` (`skiplang/compiler/src/OuterIstToIR.sk`, `pos.die("Expected closure type but got ...")`): the typed callee of the pipe still has its `Tparam` type when it reaches lowering.
- Direct calls avoid this because `call_or_constructor_` handles `Tparam` callees by promoting and wrapping them in `TAst.Annotated` with the upper-bound `Tfun` (`skiplang/compiler/src/skipTyping.sk`, the `N.Tparam` branch of `call_or_constructor_`). The `N.Pipe` case builds its `TAst.Call` node directly and never applies that promotion.
- Reproduces on the current bootstrap compiler, i.e. this predates the tracking work in #1161. The pipe tracking check added there resolves type-parameter upper bounds for `check_tracking` purposes only; with an *untracked* bound the new tracking error fires during typechecking, which is why the ICE only surfaces with tracked bounds.

## Suggested fix

Give the pipe path the same treatment as the direct-call `Tparam` path: when the callee's unfolded type is a `Tparam` with a function-typed upper bound, promote and wrap the callee in `TAst.Annotated` with that bound before building the call node (or eta-expand tparam-typed pipe callees).

mbouaziz:
## Summary

`x |> f` where `f: F` and `F: (String) -> String` typechecked and then crashed the compiler during lowering:

```
Internal error: Expected closure type but got F
```

The pipe left the callee's `Tparam` type in the typed AST; `convertType(Tparam)` yields a bare `GTparam`, not the `<builtin>.Closure` `GTapply` that `emitClosureCall`'s closure-type check requires. The equivalent direct call `f(x)` works because `call_or_constructor_`'s `N.Tparam` arm retypes the callee with its first function-type upper bound through a `TAst.Annotated` node (lowered to a cast).

This PR extracts that retyping into shared helpers (`find_tfun_upper` / `retype_tparam_callee`) used by both the direct-call arm and the `N.Pipe` case. Non-Tparam callees keep the exact same join, so every current pipe diagnostic (`pipe_1`, `pipe_2`) is byte-for-byte unchanged.

For a Tparam-typed callee the pipe now also joins against **exactly the bound being asserted** rather than the type parameter itself. The join tries every upper bound, so with several function-type bounds (`F: ((A) -> B) & ((C) -> C)`) it could validate the pipe against one bound while the callee was cast to another — recreating the ICE in a new form (incompatible store during peephole) or accepting arguments that only fit a bound the cast does not assert. Like direct calls, pipes honor only the first function-type bound; the `pipe_tparam_second_bound` invalid test and the `pipeMulti` valid case pin both directions.

**Chained bounds** (`F: G`, `G: (S) -> T`) were strictly worse than direct calls: the pipe's join recurses through bounds, so the program typechecked and then hit the ICE, while the direct call rejects cleanly (`findUppers` is non-recursive). Pipes now reject that shape with "Invalid pipe" plus an explanatory note, matching the direct-call behavior.

Not addressed here (deliberately): pipes do not run the tracking check on main for any callee shape, including this one — that is fixed wholesale by #1161.

Fixes #1269.

## Test plan

- [x] New `Typechecking/PipeTparam` valid test: pipe into a tparam-typed callee with a lambda argument and a top-level function value, plus a direct-call case pinning the refactored `call_or_constructor_` path (previously the pipe cases ICE'd; verified at `-O0` and `-O2`)
- [x] New `pipe_tparam_chained_bound` invalid test: chained bound now cleanly rejected (previously ICE'd)
- [x] New `pipe_tparam_second_bound` invalid test: argument fitting only a later function-type bound rejected (previously miscompiled); first-bound multi-bound case runs (`pipeMulti`)
- [x] Existing pipe tests unchanged: `Syntax/Pipe`, `Typechecking/Pipe3`, `Runtime/ImmediateLambdaCall`, `pipe_1`, `pipe_2`
- [x] Full compiler test suite passes locally (1707/1707)
- [ ] CI passes

### DIFF
--- a/skiplang/compiler/invalid_tests/typechecking/pipe_tparam_chained_bound.exp_err
+++ b/skiplang/compiler/invalid_tests/typechecking/pipe_tparam_chained_bound.exp_err
@@ -1,0 +1,16 @@
+File "invalid_tests/typechecking/pipe_tparam_chained_bound.sk", line 5, characters 3-8:
+Invalid pipe
+3 | // previously crashed the compiler during lowering, issue #1269)
+4 | fun callIt<G: (String) -> String, F: G>(f: F, x: String): String {
+5 |   x |> f
+  |   ^^^^^^
+6 | }
+
+File "invalid_tests/typechecking/pipe_tparam_chained_bound.sk", line 5, characters 8-8:
+The type of this expression has no function-type upper bound
+3 | // previously crashed the compiler during lowering, issue #1269)
+4 | fun callIt<G: (String) -> String, F: G>(f: F, x: String): String {
+5 |   x |> f
+  |        ^
+6 | }
+
--- a/skiplang/compiler/invalid_tests/typechecking/pipe_tparam_chained_bound.sk
+++ b/skiplang/compiler/invalid_tests/typechecking/pipe_tparam_chained_bound.sk
@@ -1,0 +1,6 @@
+// the callee's function bound is only reachable through another type
+// parameter; direct calls reject this shape and pipes must too (it
+// previously crashed the compiler during lowering, issue #1269)
+fun callIt<G: (String) -> String, F: G>(f: F, x: String): String {
+  x |> f
+}
--- a/skiplang/compiler/invalid_tests/typechecking/pipe_tparam_second_bound.exp_err
+++ b/skiplang/compiler/invalid_tests/typechecking/pipe_tparam_second_bound.exp_err
@@ -1,0 +1,33 @@
+File "invalid_tests/typechecking/pipe_tparam_second_bound.sk", line 15, characters 3-8:
+Invalid pipe
+13 |
+14 | fun callIt<F: ((A) -> Int) & ((C) -> Int)>(f: F, c: C): Int {
+15 |   c |> f
+   |   ^^^^^^
+16 | }
+
+File "invalid_tests/typechecking/pipe_tparam_second_bound.sk", line 14, characters 53-53:
+The type: C
+12 | }
+13 |
+14 | fun callIt<F: ((A) -> Int) & ((C) -> Int)>(f: F, c: C): Int {
+   |                                                     ^
+15 |   c |> f
+
+File "invalid_tests/typechecking/pipe_tparam_second_bound.sk", line 14, characters 17-17:
+Is not a subtype of: A
+12 | }
+13 |
+14 | fun callIt<F: ((A) -> Int) & ((C) -> Int)>(f: F, c: C): Int {
+   |                 ^
+15 |   c |> f
+
+File "invalid_tests/typechecking/pipe_tparam_second_bound.sk", line 14, characters 17-17:
+This type was found to be a fixed upper bound and could not be changed to a compatible type.
+To make these types compatible, annotate as this type: B
+12 | }
+13 |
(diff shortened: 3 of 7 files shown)


mbouaziz:
## Code review findings (multi-angle adversarial review of the initial version; each finding probe-verified against a compiler built from the branch)

The review ran against the first pushed version of this fix; the current head addresses the findings as noted.

### Correctness

1. **[major — fixed in current head] Multi-bound divergence between the join and the cast.** With several function-type bounds (`F: ((A) -> B) & ((C) -> C)` — legal intersection syntax), the pipe's join validates against *any* upper bound (`join_try_list` tries them all) while the callee was cast to the *first* `Tfun` bound. When the bounds' return types differ this recreated the ICE in a new form (`Internal error: Attempt to store incompatible value of type B in field 'value' which has type C` during peephole, at `-O0` and `-O2`); when only the param types differ the pipe *accepted* arguments the equivalent direct call rejects, embedding an unchecked cast in the typed AST. Fix: for a Tparam-typed callee the pipe now joins against **exactly the bound being asserted** — like direct calls, pipes honor only the first function-type bound. Pinned by the `pipe_tparam_second_bound` invalid test (second-bound-only argument rejected) and the `pipeMulti` valid case (first-bound argument runs).
2. **[minor — deferred] No tracking check for tparam-typed pipe callees.** `x |> f` with `F: untracked (String) -> String` is newly compilable and skips the "Cannot call an untracked function from a tracked context" check that the direct call enforces. This matches main's pipe-wide behavior (pipes run no tracking check for *any* callee shape); the wholesale fix is part of #1161, so it is deliberately not special-cased here — noted in the PR description.

### Cleanup (all applied in current head)

3. The chained-bound rejection was a bare one-line "Invalid pipe" while neighboring diagnostics carry explanatory traces — it now adds "The type of this expression has no function-type upper bound" at the callee position.
4. The promote + `TAst.Annotated` retype recipe was duplicated between `call_or_constructor_`'s Tparam arm and the pipe arm, 1100 lines apart — extracted into a shared `retype_tparam_callee` helper returning both the bound (which the call must be checked against) and the retyped callee.
5. The code comment implied direct calls reject chained bounds with the same message; they actually reject via `constructor_call` with a different one — comments now state the first-bound-only rule and the parity boundary precisely.

One further nit (the `find_tfun_upper` fold could read as a `.find(...)` over unfolded uppers) was left as-is to keep the extraction a verbatim, behavior-preserving move of the existing direct-call scan.

PR #1287, opened 2026-07-07
prelude: vendor Ryu, make String.toFloat correctly rounded (fix WASM round-trip + TOML float parsing)

mbouaziz:
## What

Two pre-existing float-parsing bugs share one root cause — reconstructing a
double from its decimal digits with repeated `* 0.1`, which is not correctly
rounded:

1. **WASM `String.toFloat`** (`SKIP_String__toFloat_raw`'s in-tree `atof` on
   SKIP32): `"0.3".toFloat()` yielded `0.30000000000000004`, and round-trip
   (`toString().toFloat() == value`) failed for ~85% of sampled doubles.
2. **TOML `parseFloat`**: `"3.14"` parsed to `3.1400000000000006` instead of the
   nearest double.

Both were masked while `Float.toString` produced only 9 significant digits; they
surface once toString becomes exact (a follow-up).

## How

- Vendor the Ryu library (`ryu/d2s.c` double→string and `ryu/s2d.c`
  string→double) from ulfjack/ryu @ `4c0618b0`, with `#ifdef SKIP32` guards so it
  builds under `-nostdlibinc`. (`d2s` is used by the follow-up Float.toString
  change; `s2d` is used here.)
- SKIP32 `String.toFloat` uses Ryu `s2d` (correctly rounded) for the inputs it
  supports (≤17 significant digits, standard formats), falling back to the
  existing `atof` for the rest. SKIP64 keeps libc `strtod`, already correct.
- TOML `parseFloat` keeps its validation but delegates the value to
  `String.toFloat` instead of reconstructing it by hand.

## Verification

- Runtime builds for both SKIP64 and SKIP32; the SKIP32 wasm library links.
- `s2d` round-trips all shortest-representation `toString` outputs — 0 failures
  over 7053 sampled doubles (including subnormals and `DBL_MAX`), where the old
  WASM `atof` failed 5982.
- All 331 `toml` tests pass.


jberdine (approved):
LGTM. Not sure if the CI job itself failed or there is something off.

mbouaziz:
I guess I just need to restart the CI (rebase needed anyways)
See #1335 for why

PR #1289, opened 2026-07-07
Dual-mode compiler test harness: unified @exptest mode with subprocess fallback

mbouaziz:
## Summary

Reworks the compiler test harness (`skiplang/compiler`) to run in a fast **unified mode** by default, while keeping the existing subprocess mode as a fallback and an escape hatch.

- **Unified mode (new default):** the `tests` entry in `Skargo.toml` is commented out, so Skargo compiles every test `.sk` file into a single test binary. Test entry points are marked `@exptest` and discovered at compile time via a `#forEachFunction (@exptest, …)` macro; each is run **in-process** and its stdout checked against the sibling `.exp`. This avoids a per-test `skc` subprocess + link, which is the bulk of the suite's wall-clock cost.
- **Subprocess fallback (unchanged behavior):** any valid test *not* annotated `@exptest` still runs the old way — compiled individually and executed as a subprocess. Invalid tests (which must fail to compile) always run in subprocess mode.
- **`SKC_SUBPROCESS=1`** forces every test through subprocess mode, preserving the previous end-to-end path for debugging or when a test must be compiled with the exact harness flags.
- **Compiler-version guard:** in unified mode, a `CompilerVersionCheck` test asserts that the `skc` on `PATH` matches the one that built the test binary (via `GIT_COMMIT_HASH`), so a stale binary can't silently pass.

Because unified mode compiles all tests together, each test file is module-wrapped (`module <Suite>_<Name>; … module end;`) so top-level symbols don't collide; the harness derives the suite/test path back from the module name. Untracked test entry points are supported (the harness `main` is itself `untracked`), which is needed for tests whose `main` performs untracked work.

## Contents

- `@exptest` added to every valid test across `Expand`, `Native`, `Runtime`, `Syntax`, `Typechecking`, and `Visibility`, plus newly-merged tests (`Runtime/DebugDispatchCse`, `Typechecking/{AllowTrackedCall,AllowTrackedCallMethod,PipeTparam,PipeTwice}`).
- Dual-mode harness in `tests/tests.sk`; `Skargo.toml` default flipped to unified; sktest host/harness tweaks to support in-process stdout capture and untracked test functions.
- `GlobalOrder` rewritten to avoid stdout in const initializers (incompatible with in-process execution).

## Test plan

- [ ] CI: unified mode (default) — all `@exptest` tests discovered and green
- [ ] CI: `SKC_SUBPROCESS=1` — full subprocess path still green
- [ ] Invalid tests still fail to compile as expected


mbouaziz:
CI time for compiler tests: 25min -> 16min (of which 10min are just compilation)

Issue #1297, opened 2026-07-09
Share stdlib type-checking across skc invocations via a persisted warm SKStore context


mbouaziz:
## Share stdlib type-checking across `skc` invocations via a persisted warm SKStore context

Follow-up to #1295 (type-checking the stdlib is ~3.15s ≈ **64% of every compile**, redone identically every run) and the companion #1296. Where #1296 amortizes the stdlib typing *within* one process, this issue removes it from **every separate one-shot `skc` invocation** — the case that matters for `skargo` and editors, which re-`exec` `skc` per file.

### The key realization

The typed stdlib is already a per-definition, reactively-memoized pure function of source (see #1296). It's re-typed each run only because the `SKStore.Context` is created **fresh and empty** per process (`src/main.sk:123`) and discarded at exit. If a warm, already-typed context survived across processes, `pkgDelta` (`src/skipParse.sk:83-149`) would see the unchanged stdlib, write nothing, and `context.update()` would type **only the user's file**.

### The mechanism already exists in the runtime (it's what SKDB uses)

- `--data` / `--init <file>` are parsed by the runtime (`skiplang/prelude/runtime/palloc.c:359`).
- `sk_load_mapping` mmaps the whole heap — including the reactive context root — at a **fixed address** with `MAP_SHARED | MAP_FIXED` (`palloc.c:604`), so raw pointers stay valid across processes with **zero serialization/swizzling**.
- On load, `SKIP_has_context` is true, so `runWithGcIntern` **skips re-init**: `if (!gHasContext(fork)) { gContextInit(...) }` (`src/skstore/Context.sk:1736`) — the fresh empty context from `main.sk:123` is bypassed in favor of the persisted graph.
- The store is stamped with `SKIP_get_version()` and a mismatch is rejected (`ERROR_MAPPING_VERSION`, `palloc.c:598`). Since the version is per-build, a warm store is only reused by the **exact compiler binary** that produced it — precisely the "reuse typed defs only when the typing logic is identical" guarantee, correct by construction. Per-source staleness is separately handled by `pkgDelta`.
- **Proven in production:** SKDB drives this exact `--init`/`--data` flow (`sql/*.sh`); `skc` links the same runtime, so the flags are mechanically available today — they just don't drive a producer that types + commits the stdlib.

### What to build

[...]

Issue #1296, opened 2026-07-09
Share stdlib type-checking within one skc invocation (batch / multi-file mode)


mbouaziz:
## Share stdlib type-checking within one `skc` invocation (batch / multi-file mode)

Follow-up to #1295, which measured that **type-checking the stdlib is ~3.15s ≈ 64% of every compile**, redone identically on every invocation.

### Why this is (almost) free

The typed stdlib is already a **per-definition, reactively-memoized pure function of source** — the frontend is a chain of SKStore `.map`s (`src/skipMain.sk` `type_program`: source → parse → name → `/typing/`), each typed def is its own keyed cell, and cross-def references are tracked read edges. SKStore already refuses to recompute a key whose inputs are unchanged. It's re-typed every run only because `skc` creates a **fresh empty** `SKStore.Context::create{}` per process (`src/main.sk:123`) and discards it at exit.

So within a single process, reusing one `Context` across N compilations types the stdlib **once**: the first file types it; for files 2..N, `pkgDelta` (`src/skipParse.sk:83-149`) sees the unchanged stdlib package and writes nothing, so `context.update()` re-runs only the arrows reachable from each new input. The ~64% stdlib cost is paid once, not N times.

### Prior art (this already works)

- **Unified `@exptest` harness (#1289, merged)** compiles all test `.sk` into one `skc` invocation = one `Context` = one stdlib typing.
- **Batch mode (#1136, closed)** did the same across `compile()` calls (`SKC_BATCH`).

Both prove the mechanism; neither left a supported general CLI for it.

### What to build

A supported multi-file / batch entry point that compiles several inputs in one process **without recreating the context per file** — i.e. loop `compile()` (or an equivalent) over the input set on a single `Context`, or accept a manifest of `(inputs, outputs, mainFn)`. The engine (dirty-tracking `Context.update` loop; delta `writeFiles`) already does the rest.

### Payoff / scope

- **Payoff:** eliminates the constant stdlib re-type for every file after the first within one invocation — whole-crate and whole-suite builds. Effort: **small** (mostly a CLI/driver change; no new engine work).
- **Scope limit:** this only helps *within* one process. It does **not** speed up separate `skc` invocations (incremental edit/rebuild of a single file across processes) — that's the companion approach: persist the warm context across invocations (sibling issue).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mbouaziz:
Implemented and merged — closing.

[...]

Issue #1295, opened 2026-07-08
Compiler-test compiles: SKC_TIME_PHASES misses ~90% of the time (fixed stdlib-load cost)


mbouaziz:
## Timing the compiler tests with `SKC_TIME_PHASES`

Now that `SKC_TIME_PHASES` (#1291/#1292) is in `main`, I profiled the compiler test suite by compiling every test file with it set. The headline: **`SKC_TIME_PHASES` accounts for only ~10% of a test compile — the other ~90% is a fixed, uninstrumented per-invocation cost (loading the stdlib sklib).**

### Method

Each valid test compiled exactly as the harness does — `skc -O0 --no-inline --export-function-as <Module>.main=skip_main -o <bin> <file>.sk` — with `SKC_TIME_PHASES=1`, capturing the `##TIME##` lines. All **896/896 valid tests compiled successfully** (parallel run, coverage check). Per-compile figures below are from a **single-threaded** run (no contention) over a stratified 12-file sample; every file landed within 4831–4941 ms regardless of its content, so the numbers are highly stable. Measured on this dev box (20 cores) with the stage1 release `skc`; absolute ms will differ on CI hardware, but the proportions won't.

### Result: ~90% of a compile is untimed

| | per compile |
|---|---|
| **Total `skc` wall** | **~4870 ms** |
| Sum of all `SKC_TIME_PHASES` phases | ~493 ms (**~10%**) |
| **Untimed** | **~4377 ms (~90%)** |

The instrumented phases (the ~10% we can currently see), single-threaded means:

| phase | ms | % of instrumented |
|---|--:|--:|
| `native/link` (clang++) | 184 | 37% |
| `frontend+ir` (`createIR`) | 112 | 23% |
| `native/compile` | 97 | 20% |
| `native/write_asm_files` | 50 | 10% |
| `native/create_asm_graph` | 48 | 10% |
| `native/merge_asm_graph` | 0.7 | — |
| `native/create_asm_symbols` | 0.5 | — |
| `outer/makeOuter` | 0.01 | — |

### What the untimed ~4.4s is

It's a **fixed cost paid on every `skc` invocation**, independent of the input — an *empty* `fun main(): void { void }` compiles in the same ~4.87s. `strace` pins it down:

- skc opens **`libstd.sklib` (2×) and zero prelude `.sk` sources** — so it loads the *precompiled* stdlib, it does not recompile the prelude.
[...]

PR #1298, opened 2026-07-09
Add skc --batch: compile many units in one process, sharing stdlib type-checking

mbouaziz:
## Summary

Implements #1296. Every one-shot `skc` re-type-checks the entire stdlib from source — ~64% of a compile (#1295), redone identically every time. `--batch <manifest>` compiles many units **on a single SKStore context**, so the stdlib is typed **once** and reused for the rest via the reactive engine's incremental memoization.

- `compile()` is already idempotent on a reused context (`getOrInitializeBackend` returns the existing dir; `FileCache.writeFiles` amortizes unchanged dependencies; `context.update()` re-types only the changed input). Batch just drives it in a loop over one context.
- Manifest is one unit per line: `<output> <mainFn> <input.sk> [<input2.sk> ...]`, where `<mainFn>` is exported as `skip_main`.
- New `Config.withUnit(inputs, output, mainFn)` derives a per-unit config from the shared base (same dependencies/target/opt-level); `Config.batch_mode` skips the per-unit LLVM-version check.

## Measured (this machine, `-O0 --no-inline`)

| units | batch | subprocess (N× `skc`) | speedup |
|--:|--:|--:|--:|
| 5 | 6.6 s | 24.1 s | **3.6×** |
| 30 | 18.2 s | ~147 s | **~8×** |

Marginal cost is ~500–700 ms/unit after the one-time ~3.15 s stdlib typing — i.e. the stdlib type-checking is paid once, not per file. Verified correct: 30/30 outputs match their `.exp`, and batch outputs are **byte-identical** to one-shot outputs (no cross-unit state leak).

## Behavior / scope

- A compile error is reported and **stops** at the failing unit (earlier units' outputs are already written) — build-style semantics; no context corruption since it exits.
- The non-batch path is unchanged (`batch_mode` defaults false; `--batch` is opt-in).
- Scope limit (by design): this shares typing **within one process**. Sharing across separate `skc` invocations (incremental single-file rebuilds) is the companion #1297 (persist the warm context).

## Follow-ups

- Continue-on-error with per-unit isolation (fork/COW) for test-runner use — needs the deferred-error handling explored in the old batch prototype.
- Wiring `skargo` / the test harness's subprocess-fallback path to use `--batch`.

## Test plan

- [x] 5- and 30-unit batches compile; every output matches `.exp`
- [x] batch outputs byte-identical to one-shot `skc`
- [x] compile error reports + stops at the failing unit
- [x] non-batch compile unchanged
- [ ] CI (full suite exercises the unchanged non-batch path)

### DIFF
--- a/skiplang/compiler/src/Config.sk
+++ b/skiplang/compiler/src/Config.sk
@@ -288,17 +288,12 @@ private class Config{
   // (dependencies, target, opt level, …) but different inputs, output, and
   // exported entry point.
   fun withUnit(inputs: Array<String>, output: String, mainFn: String): this {
-    exportAs = mutable UnorderedMap[];
-    exportAs.set(mainFn, "skip_main");
     this with {
       input_files => inputs,
       output => output,
-      exportedAsFunctions => exportAs.chill(),
+      exportedAsFunctions => UnorderedMap[mainFn => "skip_main"],
+      batch_mode => true,
     }
-  }
-
-  fun withBatchMode(): this {
-    this with {batch_mode => true}
   }
 }
 
--- a/skiplang/compiler/src/Config.sk
+++ b/skiplang/compiler/src/Config.sk
@@ -81,6 +81,7 @@ private class Config{
   preambles: Array<String>,
   canonize_paths: Bool,
   cwd: WorkingDirectory,
+  batch_mode: Bool = false,
 } {
   static fun make(results: Cli.ParseResults): this {
     optConfig = Optimize.Config::make(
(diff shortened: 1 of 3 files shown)


skiplang/compiler/src/Config.sk
```diff
@@ -282,6 +283,23 @@ private class Config{
...
+    this with {
+      input_files => inputs,
+      output => output,
+      exportedAsFunctions => exportAs.chill(),
+    }
+  }
+
+  fun withBatchMode(): this {
+    this with {batch_mode => true}
+  }
```
mbouaziz: Inline its sole usage

skiplang/compiler/src/Config.sk
```diff
@@ -282,6 +283,23 @@ private class Config{
...
+  // (dependencies, target, opt level, …) but different inputs, output, and
+  // exported entry point.
+  fun withUnit(inputs: Array<String>, output: String, mainFn: String): this {
+    exportAs = mutable UnorderedMap[];
+    exportAs.set(mainFn, "skip_main");
+    this with {
+      input_files => inputs,
+      output => output,
+      exportedAsFunctions => exportAs.chill(),
+    }
```
mbouaziz: `exportedAsFunctions = UnorderedMap[mainFn => "skip_main"]`

PR #1321, opened 2026-07-13
postgres adapter: fix change-detection setup, error handling, and add accessors

### LINKED ISSUE #1154: Postgres adapter: race condition in setupPgNotify, missing error handling, no accessors
## Bug: Multiple issues in `PostgresExternalService` change detection setup

**Version**: `@skip-adapter/postgres@0.0.19`

### 1. Race condition in `setupPgNotify`

`open_instances.add(instance)` is called **before** the `CREATE FUNCTION` / `LISTEN` / `CREATE TRIGGER` queries execute:

```typescript
// Current code (simplified from src/index.ts):
const setupPgNotify = async () => {
  if (!this.open_instances.has(instance)) {
    this.open_instances.add(instance);        // ← Marked as set up HERE
    await this.client.query(/* CREATE FUNCTION ... */);  // ← But these can fail
    await this.client.query(/* LISTEN ... */);
    await this.client.query(/* CREATE TRIGGER ... */);
  }
};
```

If any query fails (e.g., permission error, connection timeout), the instance is permanently marked as set up in `open_instances`, but the trigger/listener was never created. All subsequent calls to `setupPgNotify` for the same instance skip setup entirely (the `has()` check passes), leaving a dead channel — no change notifications are ever received for that table.

**Fix**: Move `open_instances.add(instance)` to after all three queries succeed, and wrap the block in try/catch.

### 2. No error handler on `pg.Client`

The constructor creates a `pg.Client` and calls `.connect()`, but never registers an `error` event handler:

```typescript
constructor(db_config) {
  this.client = new pg.Client(db_config);
  // No this.client.on("error", ...) handler
  this.client.connect().then(() => { /* ... */ });
}
```

When the PostgreSQL connection drops (network blip, server restart, idle timeout), the `error` event fires. Without a handler:

- Node.js may throw an unhandled error (crashes the process)
- Stale entries remain in `open_instances` — the instance thinks channels are active, but the underlying connection is dead
- On reconnect, `setupPgNotify` skips re-creation because `open_instances` still contains the old entries

**Fix**: Add `this.client.on("error", ...)` that clears `open_instances`, allowing channels to be re-created on the next call.

### 3. No accessors for `client` or `open_instances`

Both `client` and `open_instances` are `private` with no getters. This makes it impossible to:

- Monitor connection health (`client._connected`, `client.connection`)
- Check which instances are actively set up
- Build diagnostics/health-check endpoints
- Detect the stale-channel scenario described above

**Fix**: Expose read-only accessors:

```typescript
getClient(): pg.Client {
  return this.client;
}

getOpenInstances(): Set<string> {
  return new Set(this.open_instances); // defensive copy
}
```

### Impact

In production, a single PostgreSQL connection blip silently breaks **all** reactive change detection. The system appears to be running normally (no errors, no crashes), but no table changes propagate through the reactive graph. The only recovery is a full process restart. Without accessors, there is no way to detect this state programmatically.

### Current workaround

We maintain a patch file for `@skip-adapter/postgres@0.0.19` that:

1. Wraps `setupPgNotify` queries in try/catch, moves `open_instances.add()` to after success
2. Adds `client.on("error")` handler that clears `open_instances`
3. Exposes `getClient()` and `getOpenInstances()` accessors

### Expected behavior

1. `open_instances.add()` should only be called after all setup queries succeed
2. `pg.Client` should have an `error` event handler that invalidates stale channel state
3. Read-only accessors should be available for monitoring and diagnostics

mbouaziz:
Fixes #1154.

`PostgresExternalService` had several defects in its change-detection setup that could silently break reactive updates:

- **Failed setup left a dead channel.** `setupPgNotify` added the instance to `open_instances` *before* running the `CREATE FUNCTION` / `LISTEN` / `CREATE TRIGGER` queries, so any failure (permissions, connectivity) permanently marked the instance as set up and every later subscription skipped setup. The instance is now marked only after all queries succeed; on failure the error is reported through the subscription error callback and rethrown (rejecting `subscribe`), and the already-committed trigger function is dropped so nothing durable leaks.
- **No `error` handler on `pg.Client`.** A dropped connection surfaced as an `uncaughtException` and left `open_instances` claiming channels were live on a dead connection. node-pg neither reconnects nor accepts further queries on a client that has emitted `error`, so the handler logs the error, clears the channel state, and marks the service broken; `isConnected()` now reports this terminal state instead of returning `true` forever on a dead client.
- **No accessors for monitoring.** `getClient()` and `getOpenInstances()` (defensive copy) allow health checks and programmatic detection of the states above.
- **Setup/teardown races.** With setup no longer marked eagerly, an `unsubscribe` arriving while a subscription was still being set up would find nothing to tear down, and the completed setup would then outlive its unsubscription; `shutdown` could similarly miss in-flight setups in its drop list. Per-instance operations are now chained so each runs only after the previous one settles: `subscribe`'s initial sync + channel setup form one operation and `unsubscribe`'s teardown another, deleting the instance only once its `DROP` succeeds so a failed drop is retried by `shutdown`. The notification listener is registered inside the setup operation and removed if setup fails, so a failed subscription cannot leak a listener that would duplicate updates once retried. `shutdown` refuses new subscriptions and drains pending operations before collecting its drop list, and the notification guard also accepts instances with an operation in flight (a `NOTIFY` can be parsed from the same socket buffer that completes `CREATE TRIGGER`, before `open_instances` is updated).

Known limitation, out of reach of the adapter: the core defers `ExternalService.subscribe` via `setTimeout` while dispatching `unsubscribe` synchronously, so an unsubscribe for a resource instance created and closed in the same tick can reach the adapter before its subscribe does, and the later setup then goes through. Fixing that ordering needs a core-side change.

### Testing

- `testPostgresInResource[Wasm]` / `testPostgresInService[Wasm]` pass against Postgres 17.
- Exercised the adapter directly against a live Postgres 17 with a restricted role: a subscription whose setup fails (no `CREATE`/`TRIGGER` privilege) rejects without marking the instance open, leaves no trigger function and no notification listener behind, and a retried subscription on the same instance sets up the channel and receives change notifications exactly once — on `main` the retry silently reuses the dead channel and no notifications arrive. Also verified: `unsubscribe` racing an in-flight `subscribe` leaves no trigger/function and no open instance; killing the backend with `pg_terminate_backend` does not crash the process, clears `open_instances`, and flips `isConnected()` to `false`; `subscribe` after `shutdown` rejects; `getOpenInstances()` returns a defensive copy.

### DIFF
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -41,6 +41,24 @@ export class PostgresExternalService implements ExternalService {
   }
 
   /**
+   * The underlying PostgreSQL client, exposed for monitoring and diagnostics.
+   *
+   * @returns The `pg.Client` used by this service.
+   */
+  getClient(): pg.Client {
+    return this.client;
+  }
+
+  /**
+   * Resource instances whose change-notification channels are currently set up.
+   *
+   * @returns A copy of the set of open resource instance identifiers.
+   */
+  getOpenInstances(): Set<string> {
+    return new Set(this.open_instances);
+  }
+
+  /**
    * @param db_config - Configuration of database to which to connect.
    * @param db_config.host - Host serving database.
    * @param db_config.port - Port on which database server listens.
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -133,8 +133,8 @@ export class PostgresExternalService implements ExternalService {
 
     const setupPgNotify = async () => {
       // Reuse existing trigger/function if possible
-      if (!this.open_instances.has(instance)) {
-        this.open_instances.add(instance);
+      if (this.open_instances.has(instance)) return;
+      try {
         await this.client.query(
           format(
             `
@@ -166,6 +166,21 @@ FOR EACH ROW EXECUTE FUNCTION %I();`,
             instance,
           ),
         );
+        // Only mark the instance as set up once all queries have succeeded,
+        // so that a failed setup is retried on the next subscription instead
+        // of leaving a dead channel behind.
+        this.open_instances.add(instance);
+      } catch (e) {
+        error(
+          `Error setting up Postgres change notifications for resource instance ${instance}:`,
+        )(e);
+        // Each query is committed separately, so a failure may leave behind
+        // the already-created trigger function; drop it so that nothing
+        // durable leaks from a failed setup.
+        await this.client
+          .query(format("DROP FUNCTION IF EXISTS %I CASCADE;", instance))
+          .catch(() => {});
+        throw e;
       }
     };
 
--- a/skipruntime-ts/adapters/postgres/src/index.ts
+++ b/skipruntime-ts/adapters/postgres/src/index.ts
@@ -32,9 +32,12 @@ pg.types.setTypeParser(pg.types.builtins.TIMESTAMPTZ, (x: string) => x);
 export class PostgresExternalService implements ExternalService {
   private client: pg.Client;
   private open_instances: Set<string> = new Set<string>();
+  private broken: boolean = false;
 
   isConnected(): boolean {
-    return "_connected" in this.client && !!this.client._connected;
+    return (
+      !this.broken && "_connected" in this.client && !!this.client._connected
+    );
   }
 
   /**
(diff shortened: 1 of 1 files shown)


bennostein (approved):
Looks good to me -- I haven't been able to exercise all of these paths, but everything looks reasonable and more defensive than it was before.

PR #1366, opened 2026-07-17
Let values out of the runtime be spread, without giving up the static guarantee

### LINKED ISSUE #829: `& DepSafe` API inconvenience due to tsc weakness
TypeScript is not able to understand that `T & DepSafe` is an "object type", and so using an element of an array obtained from `Values.toArray` in contexts such as a spread expression are disallowed. E.g. `{...(vals.toArray()[0], k: v}` is a type error and needs a cast. Is there anything to do to help tsc?

mbouaziz:
Fixes #829. Lets values coming out of the runtime be spread and destructured without a cast, with no loss of static safety.

**Superseded approach:** an earlier revision of this PR fixed the ergonomics alone and accepted that `coll.map(M, {...vals.getUnique(), k: v})` would compile and then throw at runtime. @jberdine was right to reject that — trading a compile-time error for a runtime failure is a regression, not a trade. This revision closes that hole first, so it never opens.

## Why `T & DepSafe` cannot be spread

`DepSafe` is a union, and intersecting a union with an object type is what breaks. For an object `T`, `T & DepSafe` expands to a 7-arm union, and only one arm collapses:

| arm | reduces to |
|---|---|
| `T & null` | `never` |
| `T & string`, `& number`, `& boolean`, `& bigint`, `& symbol` | **survive** |
| `T & Managed` | `T & Managed` |

The primitive arms survive on purpose — `T & string` is the branded-primitive pattern (`type UserId = string & { __brand }`) that TypeScript deliberately supports. Spread requires *every* arm of a union to be an object type, so the surviving `T & string` arm makes `{...x}` a **TS2698**.

So this isn't a tsc weakness, as #829's title suggests, and there's nothing to report upstream — tsc rejects spreading a bare `string` for exactly the same rule. We did it to ourselves by intersecting a union.

## The two halves

**1. Make the `Managed` mark nominal** (first commit, and it stands alone).

The mark is installed with `Object.defineProperty(…, { enumerable: false })`, so it does not survive a spread *at runtime*. As a structural type it *did* survive one in the type, because tsc models spread as copying the symbol-keyed property. So the type system already disagreed with the runtime — `{...managed, k: v}` claimed to be dependency-safe while being a fresh, unfrozen object that `checkOrCloneParam` throws on.

That is latent today only because TS2698 rejects the spread outright, for unrelated reasons. It stops being latent the moment the spread is allowed. A private member makes `Managed` nominal, and tsc drops private members from a spread result exactly as the runtime drops the non-enumerable mark.

**2. Then `DepSafeOf<T>`** for the 12 sites that intersect `DepSafe` with another type:

```ts
export type DepSafeOf<T> = T extends object ? T & Managed : T;
```

Objects narrow to `T & Managed`, which spreads. Primitives pass through. `null extends object` is false, so `DepSafeOf<null>` is `null`. It resolves through deferred generics, since an `extends object` constraint is enough for tsc to pick a branch.

**The distinction that matters is bare-vs-intersected, not input-vs-output.** Using `DepSafe` bare is fine, so the `Params extends DepSafe[]` constraints are deliberately untouched — there the union is load-bearing, since a param may legitimately be a number *or* a frozen object.

## The result: ergonomics with the guarantee intact

Verified against the built package:

```ts
const spread = { ...vals.getUnique(), extra: 1 };   // compiles      <- #829
const { id, ...rest } = vals.getUnique();           // compiles      <- was TS2700
coll.map(P, vals.getUnique());                      // compiles      <- round-trip
coll.map(P, { ...vals.getUnique(), name: "x" });    // TS2345        <- statically rejected
coll.map(P, deepFreeze({ ...vals.getUnique() }));   // compiles      <- the fix, in the diagnostic
```

The rejection is exact and names the problem:

```
TS2345: Argument of type '{ name: string; id: number; }' is not assignable to
        parameter of type 'Row & Managed'.
  Property '[sk_managed]' is missing in type '{ name: string; id: number; }'
        but required in type 'Managed'.
```

Type and runtime now agree, which is the point — measured on the built package: `deepFreeze({...})` → `isSkManaged: true`; `{ ...frozen, d: 3 }` → `isSkManaged: false`, and `checkOrCloneParam` throws on it. The type checker rejects exactly what the runtime rejects, so that throw is no longer reachable from type-checked code.

## Breaking changes

Both are the price of the static guarantee, and neither is silent — each surfaces as a compile error with a one-line fix.

- **`Managed` becomes an emitted class**, not a type-only export, and `Frozen` extends rather than implements it. External code implementing these interfaces structurally (mocks, fakes: `class FakeCollection implements EagerCollection<K, V>`) gets TS2720 and must `extends SkManaged`. Unforgeability is the intended consequence: only `deepFreeze` and the runtime should mint `Managed`.
- **Generic wrappers funnelling `deepFreeze` into `DepSafe`** stop compiling, since for unconstrained `T` the false branch leaves a bare `T` arm that isn't provably dependency-safe:
  ```ts
  function f<T>(x: T): DepSafe { return deepFreeze(x); }
  // TS2322: Type 'DepSafeOf<T>' is not assignable to type 'DepSafe'.
  ```
  Annotating `DepSafeOf<T>`, or constraining `T` to `Json`/`object`, compiles. There's no overload that fixes it — `deepFreeze<T extends DepSafe>(value: T): T` does *not* work (an unconstrained caller `T` never matches and falls through), so I'm declaring it rather than papering over it.

Plain callers are unaffected and gain the spread. `DepSafe` stays exported, so merely *naming* it still compiles. Package is pre-1.0 (0.0.23).

## Notes

- **Commits are ordered so the hole never opens**, and each builds independently (verified per-commit). The nominal mark lands first; `DepSafeOf` follows.
- `deepFreeze`'s overload ships with its callers rather than with the type, because `DepSafeOf<V>` and `V & DepSafe` are mutually unassignable for deferred `V` — splitting them gives a commit that doesn't build. The overload is load-bearing, not cosmetic: without it core fails with 5 TS2322s from the internal `_default.ifNone/ifMany` calls.
- `Object.freeze` returns `Readonly<T>`, a mapped type, which drops the private mark; `freezeSkManaged` restores it, freezing not having removed it at runtime.
- `ObjectProxy` now intersects `Managed` rather than declaring the mark structurally, which no longer confers it — the proxy does answer to the mark at runtime.
- `Values<T>` is deliberately left unconstrained, as on `main`. Constraining it to `Json` would break plain-`interface` consumers (TS2344, the index-signature gotcha) for no benefit; only `ValuesImpl` needs the bound, for null-narrowing.
- Also quietly fixes a docs bug: typedoc renders `(T & DepSafe)[]` as `T & DepSafe []`, dropping the parens so it reads as `T & (DepSafe[])`. `DepSafeOf<T>[]` is unambiguous.
- Naming: `DepSafeOf<T>` over `Dep<T>` because `Dep<Row>` misreads as "a dependency of Row", and this is the most-hovered type in the API.

## Not included

**The examples still carry their workarounds.** `examples/chatroom` and `examples/hackernews` each bind `getUnique()` to an annotated local purely to launder the type for a spread — the annotation looks ordinary and nothing says it's load-bearing. Deleting both is the payoff of this change and is verified locally, but the examples `npm install @skipruntime/core@0.0.23` from the registry inside Docker, so they compile against the released package rather than this workspace. That cleanup follows a release containing `DepSafeOf<T>`, in the same shape as #1334. The commit is ready and waiting.

## Testing

Repo tsc 5.7.3, all after `rm -rf dist` (a stale `dist/` gives false greens). `core`, `helpers`, `server`, `tests`, `examples` build clean; `sql/ts` type-checks clean. Each of the four commits builds independently. `DepSafeOf<T>` survives declaration emit nameable rather than expanding into an unreadable union.

The five behaviours above were checked against the built package through a real example's tsconfig, and the runtime half (`deepFreeze`/`isSkManaged`/`checkOrCloneParam`, including the spread) was exercised against the emitted JS. Runtime suites are covered by CI, not locally — the skiplang toolchain doesn't build in my environment.

---
*— Claude Opus 4.8 (1M context), via Claude Code. Every type-level claim above was compiled against the repo's tsc 5.7.3 rather than asserted, and the runtime claims were measured against the emitted JS.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+ - `DepSafeOf<T>`, the dependency-safe form of `T` (#829)
+
+### Changed
+ - Use `DepSafeOf<T>` rather than `T & DepSafe` throughout the API, so that the
+   values returned by `Values`, `EagerCollection`, `LazyCollection` and `Reducer`
+   can be spread and destructured without a cast (#829)
+ - `Managed` is now a class rather than a type alias, and its mark is nominal, so
+   that spreading a dependency-safe object gives back a value that is statically
+   not dependency-safe. This matches the runtime, where the mark is
+   non-enumerable and does not survive the spread either (#829)
+
 ## [0.0.19] - 2025-11-16
 
 ### Added
--- a/skiplang/skjson/ts/binding/src/index.ts
+++ b/skiplang/skjson/ts/binding/src/index.ts
@@ -89,6 +89,21 @@ export type DepSafe =
   | symbol
   | Managed;
 
+/**
+ * The dependency-safe form of `T`.
+ *
+ * Object types are marked `Managed`, while primitives are already dependency-safe and pass through unchanged.
+ *
+ * This exists because `DepSafe` is a union, so it cannot usefully be intersected with an object type: for an object type `T`, `T & DepSafe` keeps arms such as `T & string`, which TypeScript does not consider object types, and so it rejects spreading or destructuring the result.
+ * `DepSafeOf<T>` narrows to `T & Managed` instead, which spreads.
+ *
+ * Spreading a dependency-safe object gives back a plain object, not a dependency-safe one: the `Managed` mark is nominal, and TypeScript drops it from the spread result just as the runtime does.
+ * That is intended -- `{...deps, k: v}` is a fresh, unfrozen object -- so pass it to `deepFreeze` before using it as a `Mapper` or `Reducer` parameter.
+ *
+ * @typeParam T - Type of the value.
+ */
+export type DepSafeOf<T> = T extends object ? T & Managed : T;
+
 export function checkOrCloneParam<T>(value: T): T {
   if (
     typeof value == "boolean" ||
--- a/skiplang/skjson/ts/binding/src/index.ts
+++ b/skiplang/skjson/ts/binding/src/index.ts
@@ -15,19 +15,27 @@ export const sk_managed: unique symbol = Symbol.for("Skip.managed");
  *
  * `Managed` values are important because they can be used in code that will be executed by the reactive computation system without introducing the possibility of stale or unreproducible results.
  */
-export type Managed = {
+export abstract class Managed {
   /**
+   * The mark is `private`, which makes `Managed` nominal rather than structural.
+   * That is deliberate: it is installed non-enumerably, so it does not survive an
+   * object spread at runtime, and a structural mark would survive that spread in
+   * the *type* -- letting `{...managed, k: v}` claim to be dependency-safe when it
+   * is not. TypeScript drops private members from spread results, so a nominal
+   * mark models the runtime faithfully.
+   *
+   * `declare` because the mark is installed by `Object.defineProperty` rather
+   * than by field initialization.
+   *
    * @ignore
    * @hidden
    */
-  [sk_managed]: true;
-};
+  declare private readonly [sk_managed]: true;
+}
 
-export abstract class Frozen implements Managed {
-  // tsc misses that Object.defineProperty in the constructor inits this
-  [sk_managed]!: true;
-
+export abstract class Frozen extends Managed {
   constructor() {
+    super();
     this.freeze();
   }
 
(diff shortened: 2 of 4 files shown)


mbouaziz:
Dropped the examples commit and force-pushed. `check-examples` was right to reject it: the example services `npm install` `@skipruntime/core@0.0.23` from the registry inside Docker rather than building against this workspace, so removing their workarounds made them compile against a published package that has no `DepSafeOf<T>` — failing with the exact TS2698 this PR fixes, at `hackernews.service.ts:70,40`.

That cleanup needs a release containing `DepSafeOf<T>` first, same as #1334 did for `@skiplabs/eslint-config`. The commit is ready; I'll open it once there's a version to bump to. This PR is now the type change alone.


---
*— Claude Opus 4.8 (1M context), via Claude Code.*

PR #1380, opened 2026-07-20
skargo: don't compile the dispatch shim into dependency libraries

### LINKED ISSUE #1378: sql: dev-profile build fails with "Name already bound: dispatch" (release is fine, so CI misses it)
A dev-profile build of `sql/` fails at link time with a duplicate-symbol error. Release builds are fine, which is why CI has not caught it.

## Repro

On a clean checkout:

```
$ cd sql && skargo build
...
File "…/sql/target/host/dev/build/dispatch/__skargo_dispatch.sk", line 2, characters 15-22:
Name already bound: dispatch
1 | module SkipInternal;
2 | untracked fun dispatch(): void {
  |               ^^^^^^^^

File "…/sql/target/host/dev/build/dispatch/__skargo_dispatch.sk", line 2, characters 15-22:
Previously here
```

```
$ cd sql && skargo build --release --bin skdb
    Finished: release target(s) in 130.346s
```

Both verified on a clean tree with no local modifications. `--bin skdb` does **not** help — the profile is what matters.

## What is happening

The diagnostic is confusing because it points at the *same* file, line and column twice. That is the tell: it is one source file compiled into **two different `.sklib`s**, both of which are then linked.

`skargo` generates one dispatch shim per build at `target/host/dev/build/dispatch/__skargo_dispatch.sk`:

```skip
module SkipInternal;
untracked fun dispatch(): void { … }
module end;
```

In the **dev** profile that file is appended to the compilation of *dependency libraries*. From `skargo build -vv`, counting which package compilations receive it:

| profile | compilations given `__skargo_dispatch.sk` |
| --- | --- |
| dev | `skjson`, `sqlparser` — **2** |
| release | **0** |

So `libskjson-*.sklib` and `libsqlparser-*.sklib` each end up defining `SkipInternal.dispatch`, and linking both into `skdb` collides.

The apparent trigger (inferred from the above, not directly proven) is **depending on two or more packages that each declare a `[[bin]]`**. `sql` depends on both `skjson` (`[[bin]] name = "skjson"`) and `sqlparser` (`[[bin]] name = "test"`). Consistent with that, `cd skiplang/skjson && skargo build` succeeds in dev — it is the root, so only one dispatch exists.

The dev dependency builds also pull in test sources (`skiplang/skjson/tests/test.sk` appears in a dependency's compile line, and `libsktest` is linked); release does neither. So dev appears to be building dependencies as though they were bin/test targets rather than plain libraries.

## Why CI is green

`Makefile:12` sets `SKARGO_PROFILE?=release`, and the release rule passes through the working path:

```make
sql/target/host/release/skdb: …
	cd sql && skargo build --release --bin skdb      # Makefile:62-63
```

The dev rule is a bare build and *does* hit this:

```make
sql/target/host/dev/skdb: …
	cd sql && skargo build                            # Makefile:59-60
```

`Makefile:182` runs the combined test target with `SKARGO_PROFILE=dev`, so that path is affected too.

## Impact

You cannot make a dev/debug build of skdb locally — which is exactly the build you want when debugging it. It also blocks `skargo check` on `sql/`, so editor/CI-style type checking of the package does not work.

## Suggested direction

Either stop injecting the dispatch shim into dependency *library* compilations (it is a bin entry-point dispatcher; a library should not need it), or give it a per-package-unique name so two sklibs cannot collide. Making the diagnostic mention the two `.sklib`s involved, rather than printing the same source location twice, would also make this much easier to recognise.

Found while migrating SKDB off SKStore's private obstack primitives (#1343 follow-up); not caused by that work — it reproduces on an unmodified checkout.

mbouaziz:
Fixes #1378 — `cd sql && skargo build` (dev profile) failed at link time with `Name already bound: dispatch`.

## What was wrong

In the dev profile `skargo` merges a package's bin and test targets into a single `__merged` binary, dispatched at runtime by a generated `SkipInternal.dispatch` shim compiled into that package's library.

`should_merge()` answers a *workspace-level* question, but it was being handed arbitrary units. So every package in the dependency closure that declared a `[[bin]]` also got the shim — plus its test sources and its dev-dependencies — compiled into its `.sklib`. Link two of those and they collide.

`sql` depends on both `skjson` and `sqlparser`, which each declare a bin. The trigger turns out to be narrower than the issue guessed: **two bin-declaring packages anywhere in the closure, counting the root**, so a single bin-declaring dependency is enough. Verified on a two-package fixture.

Release builds were unaffected, which is why CI missed it: `Makefile:12` defaults to `SKARGO_PROFILE=release` and merging is dev-only.

## The fix

Three commits, each independently reviewable:

1. **Scope merging to the workspace package.** The invariant is stated once, as `should_merge_unit()`, and the workspace package is threaded through unit-graph construction. This also closes a latent *release*-mode variant: `compute_deps()` forces `profile => "dev"` onto build-script chains, so a `[build-dependencies]` package declaring a bin would otherwise enter merged mode even under `--release`.

2. **Make merged-ness part of a unit's artifact identity.** Without this the fix does not take effect on an existing `target/`: the dependency's hash is unchanged, and its fingerprint covers only source mtimes — which also do not change for a dependency with no `tests/` directory. It is reported `Fresh`, the stale `.sklib` (still carrying the dispatcher) is relinked, and the build keeps failing until someone runs `skargo clean`. A merged library genuinely *is* a different artifact, so this belongs in the hash. **This invalidates every existing `target/` directory once, in both profiles.**

3. **A regression test** — `skargo`'s first. It had none: `.circleci/generate_config.sh` picks `skip-package-tests` over `skip-package-build` purely on whether a `tests/` directory exists, so `skargo` was getting a bare `skargo build --all-targets`. Adding the directory flips the job over with no `.circleci` changes.

## Verification

Everything below was run locally against a `skargo` built from this branch.

| check | result |
| --- | --- |
| `cd sql && skargo build` (the issue's repro) | stock: `Name already bound: dispatch` → fixed: `Finished: dev target(s) in 141.8s` |
| `cd sql && skargo build --release --bin skdb` | `Finished: release target(s) in 143.5s` |
| `cd skiplang/skjson && skargo test` (merged mode at the root still works) | `Failures: 0, successes: 9` |
| `cd skiplang/skargo && skargo test` (the new test) | `[OK] SkargoDispatchTests test_bin_declaring_dependency_dev_build` |
| two-package fixture, stock `skargo` | reproduces in 28s |
| stale `target/` built by the old `skargo`, commit 1 only | dependency reports `Fresh`, collision persists — this is what commit 2 is for |
| stale `target/` built by the old `skargo`, commits 1+2 | everything recompiles, build succeeds |

## Notes for the reviewer

- **The new test costs CI time.** The fixture builds the prelude from source: `skiplang/skargo`'s job goes from 48s to 78s of build work locally, so budget 1–2 min extra on the `medium` executor. It only runs when `skiplang/{skargo,prelude,toml,cli,semver}` changes. Happy to drop it if that trade isn't wanted.
- **The test can only skip under `--release`,** since merged mode is dev-only and the trick it relies on — the test binary also being a full `skargo`, re-executed via `Environ.current_exe()` — only exists there. `skargo test` defaults to dev and `bin/run-sktest.sh` doesn't override it, so CI does exercise it.
- **This does not fix released images.** `/usr/bin/skargo` in `skiplabs/skiplang:latest` is built from the bootstrap, so #1378 stays reproducible for anyone on a released image until a separate bootstrap bump.

## Deliberately out of scope

Found while investigating; each wants its own change, and I'm happy to file them as issues if useful:

- `skiplang/skjson/Skargo.toml:10` has `sktest = { path = "../skiplang/sktest" }`, which resolves to a nonexistent `skiplang/skiplang/sktest`. It is inert only because the resolver keys solved packages by name and never re-fetches a name it has already seen.
- `skiplang/sqlparser/Skargo.toml` declares `[[bin]] main = "main"`, but the package has no `main`. `skargo build --bin test` there is broken before and after this PR.
- The diagnostic itself. `render_pos` (`skiplang/compiler/src/skipError.sk:254-270`) drops `InputSource.pkg_opt`, so both traces render the identical `File "…/__skargo_dispatch.sk"` and the message looks self-contradictory. Appending the package name looks like a ~5-line change that churns no active golden file.

— Mehdi, with Claude Opus 4.8 (1M context), high effort

### DIFF
--- a/skiplang/skargo/Skargo.toml
+++ b/skiplang/skargo/Skargo.toml
@@ -8,6 +8,9 @@ toml = { path = "../toml" }
 cli = { path = "../cli" }
 semver = { path = "../semver" }
 
+[dev-dependencies]
+sktest = { path = "../sktest" }
+
 [[bin]]
 name = "skargo"
 main = "Skargo.main"
--- a/skiplang/skargo/src/build_context.sk
+++ b/skiplang/skargo/src/build_context.sk
@@ -98,6 +98,14 @@ fun should_merge(profile: String, manifest: Manifest): Bool {
   profile == "dev" && manifest.has_bins()
 }
 
+// Merging only ever applies to the workspace package. A dependency is built as
+// a plain library, even when it declares bins of its own: giving it the test
+// sources and the generated `SkipInternal.dispatch` shim would make every such
+// dependency define `dispatch`, and linking two of them collides.
+fun should_merge_unit(unit: Unit, ws_package: Package): Bool {
+  unit.pkg == ws_package && should_merge(unit.profile, unit.pkg.manifest)
+}
+
 // TODO: Return Result<...>
 fun create_bctx(
   gctx: GlobalContext,
@@ -121,7 +129,7 @@ fun create_bctx(
     opts.build_config,
     target_data.target_triple_for_arch(TargetArchHost()),
   );
-  unit_graph = build_unit_dependencies(roots, resolved_packages);
+  unit_graph = build_unit_dependencies(roots, resolved_packages, ws.package);
   target_dir = ws.target_dir();
   BuildContext{
     ws,
@@ -248,10 +256,16 @@ private fun find_named_target(
 fun build_unit_dependencies(
   roots: Array<Unit>,
   resolved_packages: Map<String, Package>,
+  ws_package: Package,
 ): UnitGraph {
   unit_graph = mutable UnitGraph[];
   for (unit in roots) {
-    build_unit_dependencies_for_unit(unit, resolved_packages, unit_graph)
+    build_unit_dependencies_for_unit(
+      unit,
+      resolved_packages,
+      ws_package,
+      unit_graph,
+    )
   };
 
   unit_graph.chill()
@@ -260,6 +274,7 @@ fun build_unit_dependencies(
 private fun build_unit_dependencies_for_unit(
   unit: Unit,
   resolved_packages: Map<String, Package>,
+  ws_package: Package,
   unit_graph: mutable UnitGraph,
   parents: List<Unit> = List.Nil(),
 ): void {
@@ -274,12 +289,13 @@ private fun build_unit_dependencies_for_unit(
     return void
   };
 
-  unit_deps = compute_deps(unit, resolved_packages);
+  unit_deps = compute_deps(unit, resolved_packages, ws_package);
   unit_graph![unit] = unit_deps;
   for (dep in unit_deps) {
     build_unit_dependencies_for_unit(
       dep,
       resolved_packages,
+      ws_package,
       unit_graph,
       List.Cons(unit, parents),
     )
@@ -289,6 +305,7 @@ private fun build_unit_dependencies_for_unit(
 private fun compute_deps(
   unit: Unit,
   resolved_packages: Map<String, Package>,
+  ws_package: Package,
 ): Array<Unit> {
   if (
     unit.target.kind is CustomBuildTarget _ &&
@@ -351,8 +368,7 @@ private fun compute_deps(
     | BinTarget _ -> false
     | LibTarget _ ->
       dep.kind is NormalDep _ ||
-        (should_merge(unit.profile, unit.pkg.manifest) &&
-          dep.kind is DevelopmentDep _)
+        (should_merge_unit(unit, ws_package) && dep.kind is DevelopmentDep _)
     | TestTarget _ -> dep.kind is DevelopmentDep _
     | CustomBuildTarget _ -> dep.kind is BuildDep _
     | HardlinkTarget _ -> false
--- a/skiplang/skargo/src/build_runner.sk
+++ b/skiplang/skargo/src/build_runner.sk
@@ -723,13 +723,28 @@ private fun hash_of(
       dep_hashes.push(hash_of(dep, bctx, out))
     };
 
-    out.set(unit, compute_hash(unit, dep_hashes.collect(Array)))
+    out.set(
+      unit,
+      compute_hash(
+        unit,
+        dep_hashes.collect(Array),
+        should_merge_unit(unit, bctx.ws.package),
+      ),
+    )
   };
 
   out[unit]
 }
 
-private fun compute_hash(unit: Unit, dep_hashes: Array<Int>): Int {
+// NOTE: `merged` is part of the artifact's identity, not just of its
+// freshness: a merged library additionally embeds the package's test sources
+// and the generated dispatcher, so the same package built as the workspace
+// package and as a dependency must not share an `.sklib`.
+private fun compute_hash(
+  unit: Unit,
+  dep_hashes: Array<Int>,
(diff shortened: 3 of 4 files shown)


mbouaziz:
Follow-ups from this work are now filed. None of them block this PR; all were verified against an unmodified checkout unless noted.

Named in the description above:

- #1390 — `skiplang/skjson`'s `sktest` dev-dep path is wrong (`../skiplang/sktest`). Latent, not live: it's held inert by two structural properties of the resolver, and I could not construct an invocation through the real package that breaks.
- #1391 — `skiplang/sqlparser`'s `[[bin]]` entry point doesn't exist, and `.circleci/generate_config.sh` carries a special case excluding the package from CI, which is what has kept that invisible since 2023. Before this PR the #1378 collision fired first and masked it; after this PR the blast radius is confined to `sqlparser` itself.
- #1392 — `render_pos` drops `InputSource.pkg_opt`, which is why #1378's diagnostic printed the same location twice. The inference in #1378 that the two traces differed *only* in `pkg_opt` is now proven rather than assumed.

Found while verifying those, and worth a look because two of them are adjacent to this change:

- #1394 — `Skargo.toml` is not a fingerprint input, so in merged mode editing `[[bin]]` produces no rebuild and you get a binary that dispatches per the previous manifest. Same staleness family as commit 2 here, but a distinct hole: commit 2 puts merged-*ness* in the hash, whereas this is about the merged dispatcher's *contents*.
- #1395 — a `[[bin]]` named `test` collides with the implicit test target in merged mode. `skiplang/sqlparser` declares exactly that, so with a `tests/` directory it fails inside generated code, and without one `skargo test` silently runs the binary and exits 0.
- #1393 — `make check` uses `find -exec`, so it exits 0 even when every package fails to check. Relevant here: it also descends into `target/`, where this PR's test writes its fixture manifests. Harmless today precisely because the target can't fail; the fix suggested there prunes `target/` and `node_modules/`.
- #1396 — `skiplang/skdate/build.sk` emits `rerun-if-changed` as a literal `${src}` (double quotes instead of backticks), so edits to `extern/src/date.c` never rebuild.
- #1397 — the generated dispatcher's "Unknown binary" message prints `${name}` verbatim, dropping the one piece of information it exists to convey. One-line fix in `generate_dispatcher`.

Two candidates I investigated and did **not** file, so they don't get re-raised later:

- Build-dependencies being compiled at `-O0` in `--release` builds. Real and reproducible, but deliberate: `a25c0e146` introduced it alongside hardcoding the host layout to dev, and cargo's `build-override` defaults the same way. No double-compilation in this repo — the entire build-dep closure is `cc` and `pkg-config`, both leaf libs with no dependencies.
- "Nothing regression-tests skargo against compiler changes." The scheduling fact is true (a `skiplang/compiler`-only change doesn't schedule the skargo job), but it's not a gap worth its own issue — the compiler's dependency closure contains no bin-declaring package, so #1378's trigger can't arise there.

— Mehdi, with Claude Opus 4.8 (1M context), high effort

PR #1427, opened 2026-07-21
SKStore: move the SKJSON and SKDB entry points onto withRegionVoid

mbouaziz:
Follow-up to #1421 (merged). There I left the two `main`s and SqlTailer on the raw `newObstack`/`destroyObstack` because a `~>` region lambda can't carry a `skipExit` or a loop `break` out — but that was too quick a call for the mains. This moves both program entry points onto `withRegionVoid`, so the only remaining external users of the raw pair are `SqlTailer`'s poll loop and the prelude's own tests.

## The two mains

`SKJSON.main` and `SKDB.main` each bracketed their whole body with `newObstack()` / `destroyObstack()`. The body is a fresh lambda scope with no outer state to capture and no top-level `return`/`break` — every exit is a `skipExit`, which is a normal call inside a lambda — so `withRegionVoid(() ~> { … })` expresses exactly the same region, and additionally reclaims it on the exception path (SKJSON's `main` had been hand-rolling `destroyObstackWithValue` on throw). On a `skipExit` path the region is reclaimed by process exit, as before; the wrapper covers a normal return.

## Why SqlTailer stays on the raw pair

`SqlTailer`'s poll loop genuinely can't be a `~>` lambda: its region spans a `mutexLock` / `condTimedWait` / `mutexUnlock` block, it mutates outer state (`init`, `tailWatermark`), and it `break`s the loop from the middle of the region. The frozen-`~>` requirement on the `withRegion*` wrappers isn't incidental — it's what stops a lambda smuggling a region-allocated pointer into outer state, which is precisely what a safe wrapper for this loop couldn't allow. Forcing it would mean either an unsafe `->` wrapper (reintroducing that hole) or restructuring the replication poll loop — real behavioural risk for one call site. So the raw pair stays public as the documented escape hatch, and the third commit points the `newObstack` doc comment at this case.

## Verification

`skargo build --release --bin skdb` is green (~122s), compiling std + skjson + sqlparser + skdb. I also ran the two entry points, since they're the risk:

- **skdb**: `--help`, `--init`, and a `CREATE TABLE / INSERT / SELECT sum(x)` round-trip returning `42`.
- **skjson**: `--version`, a subcommand handler, the no-subcommand usage path, and `check` reading a schema from stdin.

I did not run `make test-native` (needs a matched toolchain build), so this proves the entry points work on real input, not that every SQL path is unchanged — though these are pure region-management wraps with no logic change.

### DIFF
--- a/skiplang/prelude/src/skstore/Context.sk
+++ b/skiplang/prelude/src/skstore/Context.sk
@@ -1551,9 +1551,11 @@ private native base class Page
 @cpp_extern("SKIP_new_Obstack")
 // Public part of the region API, alongside destroyObstack,
 // destroyObstackWithValue, and the preferred with* wrappers below. Prefer
-// withRegion/withRegionVoid/withRegionValue; reach for the raw pair only when
-// control flow escapes the region lexically (a loop `break`, or `skipExit`
-// from a program entry point) and so cannot be a `~>` lambda.
+// withRegion/withRegionVoid/withRegionValue: they reclaim the region on both
+// the normal and the exception path. Reach for the raw pair only when the
+// region cannot be a `~>` lambda — because it must mutate outer state or span
+// a lock/condition wait, or because control flow breaks out of it. SqlTailer's
+// poll loop is the motivating case.
 native fun newObstack(): Obstack;
 
 @debug
--- a/skiplang/skjson/src/Main.sk
+++ b/skiplang/skjson/src/Main.sk
@@ -1,62 +1,52 @@
 module SKJSON;
 
 fun main(): void {
-  saved = SKStore.newObstack();
-  cmd = Cli.Command("skjson")
-    .about("Skip JSON tooling")
-    .arg(
-      Cli.Arg::bool("version")
-        .short("V")
-        .long("version")
-        .about("Print version info and exit"),
-    )
-    .arg(
-      Cli.Arg::bool("verbose")
-        .short("v")
-        .long("verbose")
-        .about("Use verbose output")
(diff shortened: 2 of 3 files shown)


mbouaziz:
Closing this — the two `main` migrations are wrong, and Mehdi caught it.

`skipExit` inside `withRegionVoid(() ~> { ... })` never returns, so the wrapper's `destroyObstack` (after `f()`) never runs. On the version / usage / success-handler paths this drops the explicit obstack destroy that the original did before each `skipExit`. `skipExit` escapes the region lambda exactly the way `SqlTailer`'s loop `break` does — which is precisely the "control flow escapes the region … cannot be a `~>` lambda" case that #1421's own doc comment already documents. I migrated the mains and then rewrote that comment to delete the case I'd correctly identified.

(The handler *exception* path is fine — `withRegionVoid`'s catch does the same `destroyObstackWithValue` copy-out the original hand-rolled. The defect is only the non-exception `skipExit` paths.)

`SKIP_destroy_Obstack` is in fact pure page-free (obstack.c:213), so under normal execution the outcome is the same as process-exit reclamation — but I asserted that equivalence in the PR without having verified it, and the mains belong in the same raw-primitive escape-hatch category as the tailer. #1421 (merged) already leaves them there. No follow-up needed; `main` is in the right state.

mbouaziz:
Redone correctly in #1436 — the fix moves `skipExit` outside the region so `withRegionVoid` destroys the obstack before exit. — Claude Opus 4.8 (1M context)

Issue #1381, opened 2026-07-20
Emit richer LLVM IR so skc actually benefits from LLVM's optimizer


mbouaziz:
Follow-up to #734, which asked whether we should move to a newer LLVM. The conclusion there was that a version bump buys us essentially nothing measurable — and the reason is worth its own issue:

**`skc` emits metadata-free, attribute-free LLVM IR.** Verified on a freshly-emitted module and on skc's own bootstrap IR (1.1M lines, 15,161 `define`s): zero `!tbaa`, zero `!prof`, zero `!llvm.loop`, zero `noalias`, zero `dereferenceable`, zero attribute groups, and every function external-linkage. The ~30 runtime declarations carry no attributes at all.

LLVM cannot infer attributes for bodiless external declarations, and cannot optimize around calls it must assume clobber all memory. So the capture-tracking rework, the alias-analysis work, the PGO-gated passes and the vectorizer all land on us as no-ops. **We will keep extracting roughly nothing from LLVM releases until `skc` annotates its output** — and none of the work below requires bumping LLVM.

> **Status note.** The bugs in §1 I verified directly against the tree. The optimization items came out of a broad audit whose adversarial-review pass was cut short, so treat them as *well-researched leads with file:line evidence*, not vetted conclusions. Each needs review and measurement before implementation. Items marked ⚠️ have specific soundness constraints called out.

---

## 1. Bugs found along the way (verified)

These are correctness issues rather than optimizations, and have been split into their own issues:

- [ ] #1382 — `skc --release` emits an undeclared `@SKIP_unreachable`, producing IR that does not parse
- [ ] #1383 — `SKIP_llvm_memset` is emitted but declared and defined nowhere
- [ ] #1384 — internal/private linkage is never emitted: `linkageOffset` omitted from the writer's next-stop computation *(the root cause of "every function is external-linkage", and the largest IPO unlock here — see §3)*
- [ ] #1385 — definition linkage is decided by the referrer instead of the referent *(latent until #1384 lands; must land with it)*
- [ ] #1386 — wasm target detected by exact match for pointer size but by prefix elsewhere, giving 64-bit pointers on wasm32
- [ ] #1387 — `skc --target` is silently ignored on the native path
- [ ] #1388 — duplicate `zeroext zeroext` on `@cpp_extern` declarations with sub-i32 return types
- [ ] #1389 — wasm32 ABI mismatch: struct-returning `@cpp_extern` functions declared multi-value but defined with `sret`

---

[...]

PR #1443, opened 2026-07-21
preamble: allocator attributes on SKIP_Obstack_alloc/calloc

### LINKED ISSUE #1438: preamble: allocator attributes on SKIP_Obstack_alloc/calloc
Part of #1381 §2. The obstack allocators are declared with **no attributes**, so LLVM must assume each call clobbers all memory, may return null, may alias anything, and has unknown alignment — defeating dead-allocation elimination and alias analysis at the 100+ allocation sites in a typical module.

`skiplang/prelude/preamble/preamble64.ll:47-48` (and `preamble32.ll`):
```
declare ptr @SKIP_Obstack_alloc(i64)
declare ptr @SKIP_Obstack_calloc(i64)
```

Proposed (64-bit):
```llvm
declare noalias align 8 ptr @SKIP_Obstack_alloc(i64) allockind("alloc,uninitialized") allocsize(0) memory(inaccessiblemem: readwrite) willreturn mustprogress nofree
declare noalias align 8 ptr @SKIP_Obstack_calloc(i64) allockind("alloc,zeroed")        allocsize(0) memory(inaccessiblemem: readwrite) willreturn mustprogress nofree
```
(drop `align 8` in preamble32 — 32-bit target).

**Soundness (verified against `obstack.c:148-174`):** `SKIP_Obstack_alloc` is a pure bump allocator — returns a fresh, 8-byte-aligned pointer that aliases nothing, touches only obstack-internal state (head/pages, i.e. `inaccessiblemem`), never frees, always returns, and does **not** trigger GC (collection is a separate call and is currently a no-op). `calloc` additionally zeroes the freshly-returned (still-inaccessible) memory → `allockind("alloc,zeroed")`. `noalias`/`align` are return attributes and must precede the return type.

**Explicitly out of scope** (documented traps): do **not** add `allocsize` to `SKIP_Obstack_shallowClone` (its size arg is ignored; it reads `argmem`), and do **not** add `noalias` to `SKIP_intern` (may return an existing interned object).

Highest value/effort item in §2. Each change should be measured, not assumed.

---
<sub>Filed by Claude (Opus 4.8), via Claude Code, on behalf of @mbouaziz.</sub>

mbouaziz:
Fixes #1438 (part of #1381 §2).

The obstack allocators were declared bare, so LLVM had to assume every call clobbers all memory, may alias anything, may return null, and has unknown alignment — defeating alias analysis and dead-allocation elimination at the 100+ allocation sites in a typical module.

```llvm
; before
declare ptr @SKIP_Obstack_alloc(i64)
declare ptr @SKIP_Obstack_calloc(i64)
; after
declare noalias align 8 ptr @SKIP_Obstack_alloc(i64) allocsize(0) memory(inaccessiblemem: readwrite) nofree
declare noalias align 8 ptr @SKIP_Obstack_calloc(i64) allocsize(0) memory(inaccessiblemem: readwrite) nofree
```
(both `preamble64.ll` and `preamble32.ll`).

## Soundness

Verified against `obstack.c:148-174`: `SKIP_Obstack_alloc` is a pure bump allocator — returns a **fresh, 8-byte-aligned** pointer that **aliases nothing**, of **`size` (arg 0)** usable bytes, touching **only obstack-internal state** (`inaccessiblemem`), **never frees**, and **never triggers GC** (collection is a separate, currently no-op call). These hold identically on the 32- and 64-bit runtimes (shared `obstack.c`). `calloc` is the same allocator plus a zeroing store into the still-inaccessible fresh memory.

## Deliberately conservative

I left `allockind`/`allocsize`-family removal semantics and `willreturn`/`mustprogress` (the rest of the set proposed in #1438) for a **measured follow-up** — they invoke malloc-family / termination semantics whose interaction with an obstack allocator and OOM-abort I did not want to assume without more scrutiny. The set above is the part I could establish as unambiguously sound, and it already enables the key optimization on its own.

## Verification

- `opt -passes=verify` accepts the declarations; an isolated `opt -O2` test confirms a **dead `SKIP_Obstack_alloc` call is now eliminated** (0 remaining) — the intended win.
- Emitted IR carries the attributes and verifies clean.
- **Full suite: `Failures: 0, successes: 1724`.**

The `preamble32.ll` (wasm32) change isn't exercised by the host suite, but the attributes describe the shared `obstack.c` runtime and are correct by construction on both targets.

Each attribute here is a silent-miscompile hazard if wrong, so this is scoped tightly and grounded in the runtime source; a reviewer's eye on the soundness argument is welcome.

---

### DIFF
--- a/skiplang/prelude/preamble/preamble32.ll
+++ b/skiplang/prelude/preamble/preamble32.ll
@@ -32,8 +32,8 @@ declare void @__cxa_end_catch()
 
 ; Obstack
 
-declare noalias align 8 ptr @SKIP_Obstack_alloc(i32) allocsize(0) memory(inaccessiblemem: readwrite) nofree
-declare noalias align 8 ptr @SKIP_Obstack_calloc(i32) allocsize(0) memory(inaccessiblemem: readwrite) nofree
+declare noalias align 8 ptr @SKIP_Obstack_alloc(i32) nofree
+declare noalias align 8 ptr @SKIP_Obstack_calloc(i32) nofree
 declare void @SKIP_Obstack_vectorUnsafeSet(ptr, ptr)
 declare void @SKIP_Obstack_collect(ptr, ptr, i64)
 declare ptr @SKIP_Obstack_shallowClone(i32, ptr)
--- a/skiplang/prelude/preamble/preamble32.ll
+++ b/skiplang/prelude/preamble/preamble32.ll
@@ -32,8 +32,8 @@ declare void @__cxa_end_catch()
 
 ; Obstack
 
-declare ptr @SKIP_Obstack_alloc(i32)
-declare ptr @SKIP_Obstack_calloc(i32)
+declare noalias align 8 ptr @SKIP_Obstack_alloc(i32) allocsize(0) memory(inaccessiblemem: readwrite) nofree
+declare noalias align 8 ptr @SKIP_Obstack_calloc(i32) allocsize(0) memory(inaccessiblemem: readwrite) nofree
 declare void @SKIP_Obstack_vectorUnsafeSet(ptr, ptr)
 declare void @SKIP_Obstack_collect(ptr, ptr, i64)
 declare ptr @SKIP_Obstack_shallowClone(i32, ptr)
--- a/skiplang/prelude/preamble/preamble64.ll
+++ b/skiplang/prelude/preamble/preamble64.ll
@@ -45,8 +45,8 @@ define void @SKIP_awaitableThrow(ptr, ptr) {
 
 ; Obstack
 
-declare ptr @SKIP_Obstack_alloc(i64)
-declare ptr @SKIP_Obstack_calloc(i64)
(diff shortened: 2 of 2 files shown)


mbouaziz:
CI caught a real wasm miscompile with the fuller attribute set: the wasm skipruntime tests hit `RuntimeError: null function or function signature mismatch` (vtable / indirect-call corruption). The native jobs passed, which is why my host-only compiler suite (1724/1724) didn't catch it — I flagged the wasm path as unexercised locally, and this is exactly that gap.

I've reduced the set to the standard, clearly-sound allocator attributes only — `noalias align 8 … nofree` — and dropped `allocsize(0)` and `memory(inaccessiblemem: readwrite)`. Those two are the plausible causes: Skip keeps object metadata (the vtable) around the returned pointer, and `SKIP_Obstack_calloc` writes memory that becomes caller-accessible, so LLVM's object-size and inaccessible-memory models didn't match the real allocator. Letting CI re-verify.

Reintroducing `allocsize`/`memory` correctly (and pinning down which one) needs the wasm test path — tracked as a follow-up on #1438.

— Mehdi (via Claude Opus 4.8)

PR #1447, opened 2026-07-22
preamble: remove the inert '; Function Attrs:' comment lines

### LINKED ISSUE #1441: preamble: remove or restore the inert '; Function Attrs:' comment lines
Part of #1381 §2. Both preambles contain `; Function Attrs: …` **comment** lines (e.g. `preamble64.ll:12,14,17,20,22`) that look like real LLVM attribute groups but are inert — LLVM ignores `;` comments. They were left behind when the real `argmemonly`/`readnone` groups were removed during the opaque-pointer / LLVM-20 migration.

They are misleading (they read as if the intrinsics still carry those attributes). Two options:
1. **Delete** the inert comment lines, or
2. **Restore** them as real attributes on the intrinsic declarations, using the modern spelling (`memory(argmem: readwrite)`, `memory(none)`, `mustprogress nocallback nofree nosync nounwind willreturn`) that the comments describe — LLVM already knows these for the standard intrinsics, so the benefit is mostly documentation/consistency.

Lowest-risk item in §2 (no semantic change if deleted; the comments affect nothing today).

---
<sub>Filed by Claude (Opus 4.8), via Claude Code, on behalf of @mbouaziz.</sub>

mbouaziz:
Fixes #1441 (part of #1381 §2).

Both preambles carried `; Function Attrs: …` lines that look like real LLVM attribute groups but are inert comments — LLVM ignores `;` text. They were left behind when the real `argmemonly`/`readnone` groups were dropped during the opaque-pointer / LLVM-20 migration, and now read as if the intrinsics and inline obstack helpers still carry `alwaysinline`/`memory(...)` when the declarations/definitions below them carry no such attributes.

Delete them (12 lines from each preamble). **Pure comment removal — no semantic change** (`;` lines never reach codegen; standard LLVM intrinsics get their canonical attributes from LLVM regardless).

Restoring some as *real* attributes could matter — notably `alwaysinline` on the `SKIP_Obstack_inl_*` helpers — but that is an optimization change to measure on its own, so it's left as a follow-up rather than folded into this cleanup.

## Verification

Comment-only deletion, so emitted codegen is unchanged. CI (native + wasm) is the gate; local host suite verification of the sibling attribute PRs was 1724/1724 and this change is a strict subset of "no-op". (My local build of this branch was interrupted by branch-switching during the #1443 fix; the change itself is a text-only comment removal.)

---


mbouaziz:
You're right on both counts — updated accordingly.

**Intrinsics** (`llvm.lifetime.*`, `llvm.ctlz/cttz/ctpop`, `llvm.eh.typeid.for`): confirmed, the attributes are documentation-only. LLVM applies each intrinsic's canonical attributes from its definition regardless of the textual `declare`, so specifying them is redundant and I've just deleted the comments.

**Defined `SKIP_*` helpers**: restored as real attributes instead of dropping them:
- `SKIP_Obstack_inl_collect{,0,1}`, `SKIP_Obstack_note_inl`, `SKIP_Obstack_store` → `alwaysinline nounwind uwtable` (tiny inline helpers, trivial/no-op bodies).
- `SKIP_String_eq` → `nounwind willreturn memory(argmem: read)` + `captures(none)` params. Note: its recorded `noinline nounwind optnone` are the attribute set clang stamps at `-O0`, **not** what it should have — it's a pure wrapper over `SKIP_String_cmp` (which got these attrs in #1445), so it should be read-only and inlinable, not un-optimizable. Let me know if you actually want it kept `optnone` for some reason.

Both preambles `opt -passes=verify` clean. Given #1443's wasm miscompile, I'm treating **CI's wasm jobs (skdb-wasm / skipruntime) as the gate**, not just the host suite — watching them now.

— Mehdi (via Claude Opus 4.8)

Issue #1343, opened 2026-07-15
Module-level 'private' is not enforced (the check is dead code)


mbouaziz:
`private` on a top-level element inside a `module` block parses and reads as an access control, but nothing enforces it — the check is dead code.

## Repro

```skip
module Secret;
private fun hidden(): Int {
  7
}
module end;

fun main(): void {
  print_string("got=" + Secret.hidden())  // compiles cleanly
}
```

`skargo check` passes.

## Why

`element_privacy` (`skiplang/compiler/src/skipExpand.sk:725-743`) returns `Some(false)` in every branch where it finds the name, and `None()` otherwise — it can never report `true`. `DefNames` (`skipExpand.sk:483`) only wraps an `Array<String>` of names and carries no privacy data to report.

So `is_private.fromSome()` is always `false`, the `(false, _) -> void` arm always wins, and the error at `skipExpand.sk:819-830`:

```
Cannot access a private module element <X> outside of module <M>
```

is unreachable.

## Impact

The codebase already relies on this as an intent marker — e.g. `skiplang/prelude/src/stdlib/filesystem/Path.sk:220` (`private fun join2`), `skiplang/prelude/src/stdlib/other/Random.sk:147` (`private fun rotl`). Nothing stops an outside caller from using them, so the module boundary those declarations imply does not exist.

Note the parser actively advertises the feature. Writing `public` at top level says:

```
Unexpected 'public'. Everything is public by default. If you want to make it
private, wrap it inside of a module ('module Name; ...') and use the 'private'
keyword.
```

## Related

What *is* enforced is the declaration-site rule — `private` at global scope, outside any module, is correctly rejected with *"Cannot declare a function as private outside of a module"* (`skipExpand.sk:835-850`).

Either the check should be implemented (thread real privacy through `DefNames`), or the parser message and the language docs should stop promising it. Documented as a caveat for now in #1340.

— Claude Opus 4.8 (1M context)

mbouaziz:
I looked into implementing this. The short version: **turning enforcement on today breaks the build** — but only at 6 declarations, and the shape is friendlier than that sounds. Laying out the findings and my read on each, since the interesting part is a design decision rather than the code.

## Why it's dead

[...]

PR #1458, opened 2026-07-22
compiler: add `--allow-module-private-access` to bypass module-private (#1454)

### LINKED ISSUE #1454: compiler: add an option to bypass module-`private` enforcement
### Request

Add a compile option that bypasses module-level `private` enforcement, so a compilation can deliberately reach another module's private elements. The primary motivation is **white-box tests**: with module-level `private` now enforced (#1343 / #1448), test code in a *different* module than the code under test can no longer poke at internals to check invariants/edge states, without exposing them in the public API.

### Proposed shape

A CLI flag (name TBD — e.g. `--allow-module-private-access`) → `Config` bool → threaded into the expand phase, checked where the error is raised so it becomes a no-op when set.

### Implementation sketch (grounded)

The check already lives in one place, and there's a clean precedent for a per-expand boolean:

- `SkipExpand.Env` (`skiplang/compiler/src/skipExpand.sk`) already carries `allowMacros: Bool`. Add an `allowPrivateAccess: Bool` beside it, defaulted in `makeEnv` and wired from `Config` the same way (`Config` already has flags like `batch_mode`; mirror its CLI wiring).
- In `check_module`, the private-access arm (`| (true, _) -> …` — the `"Cannot access a private module element …"` error) short-circuits to `void` when `env.allowPrivateAccess` is set.

That's the whole behavioral change; `element_privacy` and the symbol table stay as-is.

### Design considerations (worth deciding before building)

Because module-`private` in Skip is **module-scoped and modules span files**, white-box testing already works today by putting the test in a *separate file that declares the same module* — it then sees all of that module's privates with no new machinery. A global bypass is only needed when tests must live in a *separate* module (e.g. the `SKStoreTest`-style convention).

If we do add a bypass, the trade-offs to weigh:

- **Test vs release divergence** — a program that compiles under `--allow-module-private-access` might not compile without it.
- **Masking a missing public API** — "it's tested" stops proving "a real consumer could do this," since the test reached in through a door no real caller has.

Alternatives, roughly in order of how targeted they are:

1. Same-module test files (no feature — enforcement stays on everywhere).
2. An explicit opt-in "friend" (a module annotation naming who may see its privates) — scoped and intentional.
3. This global compile flag — simplest, bluntest.

Filing this to track the flag (option 3) as requested; the alternatives are noted so whoever picks it up can choose deliberately. Related: #1343, #1448.

mbouaziz:
Closes #1454.

Adds `--allow-module-private-access`, a compile flag that permits reaching module-`private` elements from outside their module. The motivation is white-box tests: with module-`private` enforced (#1343), test code in a *different* module can no longer poke at internals without exposing them in the public API. Off by default, so enforcement stays on everywhere else.

## Implementation

The flag threads from `Config` through the front end to the one place that decides:

```
Config.bypass_module_private
  → getOrInitializeBackend → makeOuterIst → type_program
  → SkipExpand.program → Env.allowPrivateAccess
  → check_module: the private-access arm returns void instead of erroring
```

`check_module` already special-cases same-module access; this adds one more guard arm:

```skip
| (true, _) if (env.allowPrivateAccess) -> void
```

Notes:
- The front end is otherwise config-free, so the flag is threaded explicitly rather than read from a global. `Env.allowPrivateAccess` rides `env with { … }` unchanged, so it reaches every `check_module` on the compile path (which all flow from the single `expandEnv`).
- Works in batch mode too: `Config.withUnit` uses `this with { … }`, so the field is inherited by per-unit configs.
- `element_privacy` and the symbol table are untouched — only the *reporting* is gated, so nothing else changes.

## Verification

- Without the flag, `M.secret()` on a `private fun secret` in another module errors as before; **with `--allow-module-private-access` it compiles and runs**.
- Compiler test suite: all behavioural tests pass, and the flag defaults off so flag-off behaviour is byte-identical to `main`. (Locally one meta-test, `CompilerVersionCheck`, tripped on a stale skc version-stamp left by iterative rebuilds; a clean rebuild makes `skc --version` match `HEAD` again, so CI's fresh build passes it.)

No harness test is added: the invalid/valid test framework compiles every unit with a fixed `skc` command (no per-test flags), so there's no place to exercise a custom flag. Happy to add one if a per-test-flags mechanism is wanted.

## Design note

This is the "global compile flag" option from the discussion on #1454 — the simplest, bluntest one. The alternatives (white-box tests in a *same-module* file, which already works since module-`private` is module-scoped; or an explicit per-module "friend" annotation) stay open; this flag doesn't preclude them. Flag name is easy to change if you'd prefer something like `--unsafe-allow-private`.

### DIFF
--- a/skiplang/compiler/src/Config.sk
+++ b/skiplang/compiler/src/Config.sk
@@ -82,6 +82,9 @@ class Config{
   canonize_paths: Bool,
   cwd: WorkingDirectory,
   batch_mode: Bool = false,
+  // Escape hatch: allow reaching module-`private` elements from other modules
+  // (e.g. white-box tests). Off by default so enforcement stays on.
+  bypass_module_private: Bool = false,
 } {
   static fun make(results: Cli.ParseResults): this {
     optConfig = Optimize.Config::make(
@@ -184,6 +187,7 @@ class Config{
     autogc = results.getBool("autogc");
     sampleRate = results.getString("sample-rate").toInt();
     useSpecializedNames = results.getBool("use-specialized-names");
+    bypass_module_private = results.getBool("allow-module-private-access");
     check = results.getBool("check");
     no_link = !results.getBool("link");
 
@@ -268,6 +272,7 @@ class Config{
       autogc,
       sampleRate,
       useSpecializedNames,
+      bypass_module_private,
       check,
       no_link,
       stackAlign,
--- a/skiplang/compiler/src/OuterIstToIR.sk
+++ b/skiplang/compiler/src/OuterIstToIR.sk
@@ -4881,9 +4881,12 @@ private const kEmptyMethods: UnorderedMap<
   SFunctionID,
 > = UnorderedMap[];
 
-fun makeOuterIst(context: mutable SKStore.Context): OuterIst.Program {
+fun makeOuterIst(
+  context: mutable SKStore.Context,
+  allowPrivateAccess: Bool,
+): OuterIst.Program {
   // run the compiler front end from parsing to type checking
-  typedAst = SkipMain.type_program(context);
+  typedAst = SkipMain.type_program(context, allowPrivateAccess);
   // OuterIst skeleton is built here, but code is done lazily
   SkipMakeOuterIst.program(context, typedAst)
 }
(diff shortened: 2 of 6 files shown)


mbouaziz:
Not merging as the code is not the nicest.
Keeping this just in a case someone needs to compile something under the old (disabled) privacy checks or wants to use it for tests even though they are other options

Issue #1450, opened 2026-07-22
preamble: reintroduce allocsize/memory on the obstack allocators, verified against the wasm tests


mbouaziz:
Follow-up to #1438 / #1443. That work shipped only the **safe subset** of obstack-allocator attributes:

```llvm
declare noalias align 8 ptr @SKIP_Obstack_alloc(i64) nofree
declare noalias align 8 ptr @SKIP_Obstack_calloc(i64) nofree
```

The two highest-value attributes — **`allocsize(0)`** and **`memory(inaccessiblemem: readwrite)`** — were dropped because adding them caused a **wasm miscompile**: the wasm skipruntime tests failed with

```
RuntimeError: null function or function signature mismatch
  at libskipruntime….so.sk.Sequence__collect.2 (wasm-function[58])
  … SKIP_getExceptionMessage …
```

i.e. vtable / indirect-call corruption. The **native** jobs (compiler, prelude, skdb) all passed, so a host-only `skc` test suite does **not** catch this — CI's `skdb-wasm` / `skipruntime` / `skipruntime-bun` jobs are the oracle.

## What still needs doing

Reintroduce the object-size and memory-effect information on the allocators — the part that actually unlocks dead-allocation elimination and alias analysis at the 100+ allocation sites per module — but **correctly**, verified against the wasm test path, one attribute at a time.

## Leads on why the fuller set miscompiled

- **`SKIP_Obstack_calloc` + `memory(inaccessiblemem: readwrite)`**: calloc `memset`s memory that becomes *caller-accessible*, which contradicts an inaccessible-only model, and without `allockind("alloc,zeroed")` LLVM isn't told the result is zeroed. Likely needs `allockind("alloc,zeroed")` and/or `memory(argmem: write, inaccessiblemem: readwrite)` rather than plain `inaccessiblemem`.
- **`allocsize(0)`**: Skip stores object metadata (the vtable) around the returned pointer (`emitLoadVTable` reads at `vtableByteSize * -8` bits); the object-size model may not line up with how the pointer is offset from the raw allocation. Worth checking whether allocsize alone (without `memory`) still miscompiles.

## Suggested approach

1. Add `allocsize(0)` alone → run the wasm skipruntime tests. If green, it's sound; if red, it's an allocsize/layout issue.
2. Add `memory(...)` with proper `allockind` on calloc (and the right `memory(...)` spelling) → wasm tests.
[...]

PR #1460, opened 2026-07-22
preamble: bisect — allocsize(0) alone on the obstack allocators (toward #1450)

mbouaziz:
Toward #1450 (reintroduce the value attributes the allocator regression forced us to drop). The reduced set that landed in #1443 kept only `noalias align 8 nofree`; the fuller set (`allocsize(0)` + `memory(inaccessiblemem: readwrite)`) miscompiled wasm (`null function / signature mismatch` = vtable corruption), and **which of the two caused it was never pinned down** — native passed either way, and I reduced blindly.

This **isolates one variable**: `allocsize(0)` alone on `SKIP_Obstack_alloc`/`calloc`. It tells LLVM the allocation size is arg 0 (object-size / dereferenceable reasoning, dead-alloc elimination).

- If the **wasm jobs stay green** → `allocsize` is sound; `memory()` was the offender, and a follow-up can add the correctly-spelled `memory`/`allockind` on top.
- If they **go red** → `allocsize` is the offender (most likely its interaction with the object's negative-offset vtable/metadata) → revert + document precisely.

Gated on `skdb-wasm`/`skipruntime`. This is a deliberate experiment using CI as the wasm oracle I lacked when the original regression slipped through.

---

### DIFF
--- a/skiplang/prelude/preamble/preamble32.ll
+++ b/skiplang/prelude/preamble/preamble32.ll
@@ -26,8 +26,8 @@ declare void @__cxa_end_catch()
 
 ; Obstack
 
-declare noalias align 8 ptr @SKIP_Obstack_alloc(i32) nofree
-declare noalias align 8 ptr @SKIP_Obstack_calloc(i32) nofree
+declare noalias align 8 ptr @SKIP_Obstack_alloc(i32) allocsize(0) nofree
+declare noalias align 8 ptr @SKIP_Obstack_calloc(i32) allocsize(0) nofree
 declare void @SKIP_Obstack_vectorUnsafeSet(ptr writeonly captures(none) nonnull, ptr) nounwind willreturn nofree memory(argmem: write)
 declare void @SKIP_Obstack_collect(ptr, ptr, i64)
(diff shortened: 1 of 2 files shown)


mbouaziz:
Bisect result: **`allocsize(0)` alone reproduces the exact #1443 miscompile** — `RuntimeError: null function or function signature mismatch`, 31 wasm tests failing (skdb-wasm + skipruntime), while skipruntime-bun and all native jobs pass.

So `allocsize` is the confirmed culprit, and it's *fundamentally* incompatible with Skip's object model: `SKIP_Obstack_alloc` returns `head+8` and skc uses that as the object pointer with the **vtable at `pointer-8`**. `allocsize(0)` tells LLVM the object is `[pointer, pointer+size)`, so the negative-offset vtable load is out-of-bounds → LLVM miscompiles it → indirect-call/vtable corruption. There's no correct `allocsize` without changing the allocation ABI to return the metadata base.

Closing — `allocsize` is out. Next: isolate whether `memory(inaccessiblemem: readwrite)` (the other half of the original regression) is independently sound.

Issue #1489, opened 2026-07-28
Generalize the native alloc contract to `@alloc(none|bounded|unbounded)`


mbouaziz:
The native allocation contract is now a mandatory binary choice on external natives — `@no_alloc` or `@may_alloc` — and derived for `@intrinsic` (#1346).

The compiler's own lattice is richer than that. `AllocAmount` has **three** rungs:

```
AllocNothing()  |  AllocBounded()  |  AllocUnbounded()
```

but the native path in `optimize.sk` can only ever produce `AllocNothing` or `AllocUnbounded`. **`AllocBounded` is unreachable for a native**, so it can only ever arise from analysing a real body.

That gap costs something concrete. `shouldGC` (`gc.sk:81-96`) only fires on `AllocUnbounded`, so a native that allocates a *fixed, small* amount — one string, one option — is currently forced all the way to `@may_alloc`/`AllocUnbounded` and pessimistically triggers a GC safepoint it does not need. A good number of the prelude's allocating natives are exactly this shape:

- `getcwd`, `realpath`, `readdir`, `String::unsafeSlice`, `StringIterator::substring`, `Environ::current_dir`/`internalGetArgN`/`internalGetEnvN`/`varOpt`, `Time::strftime` — each returns a single freshly-created string (`sk_string_create`), not an unbounded structure.

Contrast the genuinely unbounded ones — `String::fromChars`/`fromUtf8`, `multiThreadedTabulate`, `copyValueToObstack`, `gContextSync`, `intern` — where the amount scales with the input.

## Proposal

Generalize the two annotations into one parameterized form that maps 1:1 onto the existing lattice:

```skip
@alloc(none)        // == @no_alloc      -> AllocNothing()
@alloc(bounded)     // (new)             -> AllocBounded()
@alloc(unbounded)   // == @may_alloc     -> AllocUnbounded()
```

`@no_alloc` / `@may_alloc` can stay as aliases for the two common cases, or be dropped in favour of the single spelling — worth deciding as part of this.

Two things to be careful about:

1. **`bounded` carries the same miscompile risk as `none`.** Claiming `bounded` for something that actually allocates proportionally to its input is as wrong as claiming `none`; it needs the same per-native audit discipline that #1346's migration used, and the same "if in doubt, say unbounded" bias.
[...]

Issue #1347, opened 2026-07-15
'native' without @cpp_extern/@cpp_runtime/@intrinsic is an ICE, and only at the back end


mbouaziz:
A `native` declaration must carry one of `@cpp_extern`, `@cpp_runtime` or `@intrinsic`. Plain user code can violate that, and when it does the compiler reports an internal error instead of a diagnostic.

## Repro

```skip
native fun bare(): Int;

fun main(): void {
  print_string("x=" + bare())
}
```

```
Internal error: Native functions must be declared with @cpp_extern, @cpp_runtime, or @intrinsic
```

## Why

The guard is a `pos.invariant` (`OuterIstToIR.sk:2401-2410` for functions, `2522-2531` for methods, with the message differing only in "methods"). `pos.invariant` goes to `pos.die` → `Assert.fail` → `throw InternalError` (`FileRange.sk:61-65, 80-82`; `Assert.sk:16-18`), so the user sees an ICE with a file position rather than a clean type error.

Two problems:

1. **It is an ICE, not a diagnostic.** This is a perfectly ordinary user mistake — forgetting the annotation on a `native` declaration — and it should be a normal error with a source span.
2. **It is back-end only.** `skargo check` never reaches it, so the mistake survives a clean check and only surfaces at build time.

Since `isNative = native_.isSome() || @cpp_extern || @cpp_runtime`, the two cpp annotations self-satisfy the check, so the invariant fires on exactly one thing: a `native` fun/method carrying none of the three (or only an unrelated one such as `@debug`).

## Ask

Move the check into naming/typing so it produces a normal diagnostic and is caught by `skargo check`.

The same shape applies to `@cpp_export` on a class — *"@cpp_export classes must be final and may not extend any base classes"* is also a `pos.invariant` (`OuterIstToIR.sk:2248-2254`), also back-end only. Its message is additionally misleading: `convertClassKind` maps `value class` → `KValue`, and the check is `kind is KClass _`, so a `@cpp_export value class` fails despite value classes being final.

Documented in #1340.

— Claude Opus 4.8 (1M context)

Issue #1346, opened 2026-07-15
@may_alloc/@no_throw/@no_return: unsafe default, and silently dead on functions with a body


mbouaziz:
The three contract annotations for native functions have two sharp edges, both silent.

## 1. `@may_alloc` defaults in the unsafe direction

For a function with no body, `optimize.sk:526-542` sets:

```
allocAmount = if @may_alloc then AllocUnbounded() else AllocNothing()
```

So a native function **without** `@may_alloc` is taken as an axiom to allocate *nothing*. Forgetting it on an allocating C++ runtime function is a silent wrong-code risk with no diagnostic — the value is axiomatic (there is no body to analyse) and propagates to callers via `calleeAllocAmount`/`instrAllocAmount` (`gc.sk:386-410, 419-477`), where it feeds the `shouldGC` decision (`gc.sk:81-96`).

Contrast `@no_throw` and `@no_return`, which both default to the *safe* answer when absent (`optimize.sk:531-532`: `returns = !@no_return`, `canThrow = !@no_throw`). `@may_alloc` is the odd one out, and it is the one where being wrong is a memory-safety issue rather than a missed optimization.

This compounds with #1345: since annotation names are unvalidated, a typo'd `@may_aloc` silently lands on the unsafe default.

Consider inverting it (`@no_alloc`, defaulting to "may allocate"), so that the failure mode of forgetting the annotation is a missed optimization rather than wrong code.

## 2. All three are silently dead on a function with a body

`optimize.sk:514-544` only consults these annotations in the `else` branch, guarded by `pos.invariant(f.isNative, ...)` at 521. For a function that has a body, the compiler derives `returns`/`canThrow` from the body and the annotation is never read — no error, no warning.

The prelude has two live examples of exactly this:

- `skiplang/prelude/src/stdlib/other/System.sk:55-61` — `print_string` carries `@no_throw`, but has a Skip body and no `@cpp_extern`, so the annotation does nothing.
- `skiplang/prelude/src/stdlib/other/System.sk:112-117` — `skipExit` carries `@no_return`, but has a body. (`returns = false` is nevertheless derived from the body, so the intent happens to be met by accident.)

Both read as meaningful and are not.

Worth either warning on a contract annotation attached to a function with an implementation, or dropping the two dead ones in the prelude — they currently teach the wrong thing by example.

Documented in #1340.

— Claude Opus 4.8 (1M context)

Issue #1345, opened 2026-07-15
Unknown/misspelled annotations are silently ignored, including load-bearing ones like @debug


mbouaziz:
The compiler never validates annotation names. An annotation it does not recognize is silently accepted and does nothing — including a typo of one that matters for correctness.

## Repro

```skip
@totally_bogus_annotation
@debbug                    // typo for @debug
fun helper(): Int { 42 }
```

Compiles with no error and no warning.

## Why

The parser takes `@` plus any non-type identifier and stores annotations as raw strings in an `SSet` (`SkipParser.sk:886-901`; the comment at 880-885 notes "we continue to store annos as simple strings"). Every consumer then does an exact string match via `annotations.contains(name)` (`annotations.sk:20, 79-81`). There is no whitelist and no "unknown annotation" diagnostic anywhere.

## Why it matters

Several annotations are load-bearing for correctness, and a typo silently removes the effect with no signal:

- `@debbug` instead of `@debug` silently restores CSE — which is exactly the miscompilation #1337 fixed (two identical calls to a side-effecting function merged into one), and #1342.
- A misspelled `@may_alloc` leaves the function at the unsafe default (assumed to allocate nothing) — see #1346.

Both fail *open*, and neither produces a diagnostic.

## The tension

Unknown annotations cannot simply be rejected: the openness is load-bearing. The compiler indexes every function by the annotations it carries, and `#forEachFunction (@foo, #f, #name)` enumerates them (`skipNaming.sk:3806-3859`, `funsByAnnotationDir`). That is how a library defines its own annotation with no compiler support — sktest's `@test` is exactly this (`skiplang/sktest/src/harness.sk:36`) and is unknown to the compiler.

So a plain whitelist would break user-defined annotations.

## Possible directions

- Warn on an annotation that is neither compiler-known nor referenced by any `#forEachFunction` in the compilation — this keeps the extension point working while catching typos, since a user-defined annotation nobody enumerates is dead anyway.
- Or: warn only on names within a small edit distance of a compiler-known annotation, reusing the existing `SkipDidYouMean` machinery (which already backs `@synonym`).
- At minimum, validate the annotations the compiler itself defines.

Documented as a caveat for now in #1340.

— Claude Opus 4.8 (1M context)

PR #1497, opened 2026-07-28
compiler: derive the alloc contract for intrinsics, make it mandatory for externals (#1346)

mbouaziz:
Stacked on #1475 (review that first; this PR's diff assumes it). Finishes the `@may_alloc` half of #1346.

#1475 inverted the native allocation contract to `@no_alloc`, so the *default* became the safe one. This removes the default entirely, in the two directions it can be removed:

- **Derived where the compiler already knows the answer** — `@intrinsic`.
- **Mandatory where it cannot** — external natives (`@cpp_extern` / `@cpp_runtime`).

## `@intrinsic`: derived, and the annotation is now rejected

An intrinsic never runs as a function. `IntrinsicsInliner` — a `kOptPrerequisites` pass, so it always runs, even at `-O0` — replaces every call with the IR it lowers to, and any allocation is then attributed to the **caller's own instructions** (`ArrayAlloc` for `Array::unsafe_make`, `StringConcat` for `String::concat`, and so on). The intrinsic's own `AllocAmount` is therefore derivable rather than declarable, and `optimize.sk` now derives it.

So declaring `@no_alloc`/`@may_alloc` on an `@intrinsic` is a hand-maintained duplicate of `intrinsics.sk` that is free to drift out of sync with the lowering. It is now an error, and the 64 such annotations #1475 added to the prelude are deleted:

```
@intrinsic
@no_alloc          // error: @intrinsic functions must not declare @no_alloc or
native fun f(): Int;  //        @may_alloc: the allocation behaviour is derived
```

## External natives: mandatory

A `@cpp_extern`/`@cpp_runtime` native has no body to analyse, so its allocation behaviour is an axiom either way. Rather than let silence pick the axiom, silence is now an error — every external native must declare exactly one of `@no_alloc` or `@may_alloc`:

```
@cpp_extern("SKIP_some_native")
native fun f(): Int;   // error: Native functions must declare @no_alloc or @may_alloc
```

Declaring both is also an error. This is deliberately asymmetric with `@no_throw`/`@no_return`, which stay defaulted: allocation is the contract where being wrong is a memory-safety miscompile rather than a missed optimization, so it is the one worth forcing an author to think about. It also closes the #1345 hole for this annotation specifically — a typo'd `@no_aloc` no longer silently picks an answer, it fails to compile.

Both checks live in **naming** (`check_alloc_contract`, next to the #1347/#1354 checks), so they are ordinary diagnostics caught by `skargo check`.

## Migration

Every external native in the repo now declares the contract — 274 natives across the prelude, `sql`, `skdate`, `skipruntime-ts`, and the test suites. As in #1475 this was audited per-native against the real implementation (runtime C/C++, the addon shims, and the TypeScript/wasm glue) rather than assumed, with a second adversarial pass over every `@no_alloc` verdict.

One class of result worth flagging: several `skipruntime-ts` FFI natives return a scalar (`void`, `UInt32`, `Float`) but still take `@may_alloc`, because they call back into user JavaScript which can re-enter Skip and allocate — the same transitive reasoning that keeps `etry` off `@no_alloc`.

## Tests / verification

- Four new `invalid_tests/typechecking/` cases — `alloc_missing`, `alloc_method_missing`, `alloc_both`, `alloc_on_intrinsic` — each to its exact golden.
- A repo-wide scan confirms 274 conforming natives and zero violations (no external missing or double-declaring, no intrinsic carrying one).
- STAGE=2 clean, full suite green.

Docs updated: the contract table, why this one is mandatory while the other two are defaulted, the `shouldGC` wiring, and the derivation rule for intrinsics.

A follow-up (#1489) proposes generalizing the pair into `@alloc(none|bounded|unbounded)` to reach the `AllocBounded` rung, which the native path cannot express today.


mbouaziz:
Folding this into #1475 so the whole change reviews as one PR, per Mehdi — stacking meant #1475 added 172 `@no_alloc` annotations that this PR then partly deleted and re-spelled, which is churn for a reviewer rather than signal.

#1475 now carries the final state directly: derive the contract for `@intrinsic`, require it on externals, and declare it once on all 274 natives. Same tree as this branch, same verification.

